PR #36, opened 2015-05-23
Modify weewx shebang

knnniggett:
Issue:
Weewx registers itself in the process list generically as "python" rather than as "weewxd".  While this does not affect weewx's functionality, this makes it more difficult for health monitoring apps, such as Zabbix, to monitor the status of weewx.

Cause:
After some research, it seems that the root cause is due to the shebang statement in weewxd. This pull request removes env from the shebang.  It seems the only disadvantage to this proposed change is flexibility in choosing which python interpreter to use, but is that really an issue in this case?

Further Reading:
Note that there are at lease two other ways to "solve" this issue, but they both involved importing additional libraries into weewx. At the time, those solutions seemed unnecessary.

Reference:
http://stackoverflow.com/questions/2429511/why-do-people-write-usr-bin-env-python-on-the-first-line-of-a-python-script

Testing this change produced the desired results. Weewx registers its name as weewxd, rather than python, and Zabbix stops complaining that the weewx service is not running:
```
[abauer@pidora bin]$ systemctl status weewx
weewx.service - weewx weather system
   Loaded: loaded (/usr/lib/systemd/system/weewx.service; enabled)
   Active: active (running) since Sat 2015-05-23 07:37:44 CDT; 7min ago
 Main PID: 9005 (weewxd)
   CGroup: /system.slice/weewx.service
           └─9005 /usr/bin/python /usr/bin/weewxd --pidfile=/var/run/weewx.pi...

May 23 07:44:30 pidora.local weewx[9005]: acurite: next read in 18 seconds
May 23 07:44:48 pidora.local weewx[9005]: acurite: Found station at bus=001 ...5
May 23 07:44:48 pidora.local weewx[9005]: acurite: R1: 01 00 bd 78 00 28 1f ...f
May 23 07:44:48 pidora.local weewx[9005]: acurite: next read in 6 seconds
May 23 07:44:54 pidora.local weewx[9005]: acurite: Found station at bus=001 ...5
May 23 07:44:54 pidora.local weewx[9005]: acurite: R2: 02 00 00 80 00 00 00 ...e
May 23 07:44:54 pidora.local weewx[9005]: acurite: next read in 12 seconds
May 23 07:45:06 pidora.local weewx[9005]: acurite: Found station at bus=001 ...5
May 23 07:45:06 pidora.local weewx[9005]: acurite: R1: 01 00 bd 71 00 1b 09 ...f
May 23 07:45:06 pidora.local weewx[9005]: acurite: next read in 18 seconds
Hint: Some lines were ellipsized, use -l to show in full.
[abauer@pidora bin]$ ps -A |grep weewx
 9005 ?        00:00:57 weewxd
```

### DIFF
--- a/bin/weewxd
+++ b/bin/weewxd
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python
 #
 #    Copyright (c) 2009-2015 Tom Keffer <tkeffer@gmail.com>
 #


tkeffer:
Hello, and thanks.

Can we be sure all installations will have `/usr/bin/python` defined? What if Python was installed in, say, `/usr/local/bin/python`?

-tk

matthewwall:
here are a few other use cases to consider:

when python is located in /opt
when python is located in /sw (for example, MacOSX using fink)
tomato/openwrt installations when python is on a separate usb/sd
when there are multiple python versions installed

matthewwall:
perhaps you could keep zabbix happy by invoking weewx like this:

`/path/to/python /path/to/weewxd /path/to/weewx.conf`

for example in the rc script or systemd .service file.  that way the .py entry points in weewx would be generic (i.e., would work on any system that has /usr/bin/env), and only the system-specific stuff (init.d or .service) would need to be customized for a specific platform.

knnniggett:
I grep'ed all the python scripts on my system in search of one that called the python interpreter via env, and I found none, nor did I find one that called another interpreter in a similar manner such as perl.   

The way I see it, imagine the problem a sys admin would face if every script on the system run by init/systemd showed up in the process list as python, perl, bash, etc. rather than the real name of the script.  This is not normal behavior for a service.

Since Python is available as a package in all major linux distros, I'm not sure one would ever encounter the python interpreter with a prefix of /usr/local, but I agree the *BSD's would pose a problem since they like to put things under /opt. Using the @PREFIX@ make macro would take care of that.  

I'm currently on vacation at the moment, and can respond in more detail next week when I return to the country.  I'd like to take a close look at some other projects and see how they address this issue.

tkeffer:
Thanks, Andrew

-tk

Fat-fingered from my Android

knnniggett:
I browsed through several Python projects on github, and made the following observation.

Take the project Ansible, for example.  Browsing through the project files, one can see that the upstream source files use the `#!/usr/bin/env python` shebang, which seems to be the python standard.

However, if I yum-installed the Ansible package onto my system, the same files had a shebang of `#!/usr/bin/python`.  

This observation was consistent for all the Python projects I looked at. That explains why none of the other python packages on my (Fedora) system were not using env in the shebang.  

Just to make sure this wasn't a Redhat thing, I performed the same steps on my Ubuntu workstation and got the same results.  The packaging process for both distros appear to be intentionally replacing the shebang.

Further investigation reveals this is indeed intentional:
https://fedoraproject.org/wiki/Features/SystemPythonExecutablesUseSystemPython
https://fedoraproject.org/wiki/Script_Interpreters_%28draft%29
http://pkgs.fedoraproject.org/cgit/python.git/tree/python.spec#n1373

Since modifying the upstream source during the packaging process appears to be the industry adopted practice, I plan to follow suit.  

It looks like there are multiple tools/methods available to accomplish this:
- distutils may automagically do this (have to verify)
- setup.py has the "-e" option
- pathfix.py script from the python-tools package

I'll have to play with the rpmbuild process to figure out which works best, but for now I will close out this PR since it is no longer relevant.

PR #49, opened 2015-07-08
Archiving and write over usb

amiles347:
Supports station's archive ability
Configurator expanded to support reading/setting of station settings

### DIFF
--- a/bin/weewx/drivers/te923.py
+++ b/bin/weewx/drivers/te923.py
@@ -1,12 +1,22 @@
 #!/usr/bin/env python
-#
-# Copyright 2013 Matthew Wall
+# $Id: te923.py 2775 2014-12-03 20:19:47Z mwall $
+# Copyright 2015 Matthew Wall/Andrew Miles
 # See the file LICENSE.txt for your full rights.
 #
 # Thanks to Sebastian John for the te923tool written in C (v0.6.1):
 #   http://te923.fukz.org/
 # Thanks to Mark Teel for the te923 implementation in wview:
 #   http://www.wviewweather.com/
+#
+# History: 0.12 - Experimental version developed by Matthew Wall
+#          0.13 - Added reading of latitude, longitude, altitude
+#               - Added reading of data logger
+#               - Automatic detection of memory size
+#               - More data passed back to weewx even if schema doesn't use it
+#               - Simplified sensor state to ok, no_link or invalid.  No evidence of out_of_range or error states present in te923tool code
+#          0.14 - Improved read (based on 0.12)
+#          0.15 - Added write to station
+#               - Multiple options added to configurator
 
 """Classes and functions for interfacing with te923 weather stations.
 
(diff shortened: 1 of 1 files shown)


matthewwall:
andrew, thanks for this. that is a lot of work - very nice.  i will do some testing with it on the small memory model te923. m

amiles347:
Don't mention it, it was a fun project!  Only wish the TE923 had a better data model.  No time and the date is updated at midday for some reason.  I still need to look at the rain sensor link.  I installed the sensor on the roof so it's not in the most convenient place to test the link failing :)

Andy

amiles347:
Why closed and not merged?

tkeffer:
My apologies. I don't know why I closed this! I think I must have confused it with something else.

tkeffer:
Matthew, were you able to test this?

tkeffer:
MW: any progress?

matthewwall:
andrew, i finally merged your changes.  the result is a complete overhaul of the te923 driver at commit f1ffb0b4819cbb32c6af2e5788e974194c5ef189

a few things to note:
- the decode_* methods should be pushed into the TE923Station class.  i'll get to it, but i hate refactoring python (tabification nightmares!)
- the result is about 4K smaller than andrew's request, and more than double (95K) the original driver (40K)
- i have verified byte-compatibility when reading memory by comparing with te923 tool
- i have disabled record_generation=hardware for now
- there are still some suspect bits, such as altitude and location (why are these in multiple places in the station memory?)

i'm running it on a te923 with small memory, but that station is almost exactly on the other side of the planet from where i am now, so at the moment i cannot check everything against the console.

i'm wondering if we should make record_generation=software the default for weewx.  the te923 defaults to 1 hour archive intervals.  the ws23xx defaults to 1 hour archive intervals.  the fo stations default to 30 minute archive intervals.  in each case record_generation=software seems to be a more appropriate default, especially for noobs.

amiles347:
Hi Matthew

Thanks for finally merging :)

Some answer to your comments:

* Size of driver: Not sure why you see 4K less but doubling the original
driver is expected.  I added a lot - mostly in configuring the device.
* Agree with switching to software.  The hardware generation has already
caused some users compatibility issues (see posts a few months ago) and as
you say the TE923 factory default is 1 hour and has to be changed before it
becomes useful.
* Altitude/Location: I wasn't too impressed with the data structure used by
Nexus.  Seems a bit random.  However, the device holds 2 locations.  One is
for displaying the time and the other is for the sunrise/sunset
calculations.  Why you wish to have these different is a bit of a mystery
to me.  Maybe you wish to wake up in New York and see when the sun rises in
London :)

Andy

tkeffer:
The first thing a new user does is compare what s/he is seeing on the console to what's showing up in weewx. So, deferring to hardware setting generally follows the principle of least astonishment. 

But, most users are not very aware of the archive record interval of their hardware, so maybe it's an exception.

PR #50, opened 2015-07-10
Make sure rain amount doesn't become negative.

maniac103:
Total rain amount may decrease (e.g. when exchanging sensor batteries or
on otherwise triggered sensor reset), so don't write (large) negative
values in that case.

I've seen this happen to me 3 times since setting up weewx 3 months ago. Each time, it yielded an entry of about -1800mm of rain.

### DIFF
--- a/bin/weewx/drivers/wmr100.py
+++ b/bin/weewx/drivers/wmr100.py
@@ -30,6 +30,7 @@ import syslog
 import usb
 
 import weewx.drivers
+import weewx.wxformulas
 import weeutil.weeutil
 
 DRIVER_NAME = 'WMR100'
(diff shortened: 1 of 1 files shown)


maniac103:
Looking further in the archive data, I see a rain value of -1588mm on the 7th of April at 16:15, followed by a value of 1588mm one interval (5 minutes) later. So it looks like this was just a faulty transmission of the rain data (0 instead of the correct value) for one interval.

I wonder whether this should additionally (or instead) be avoided by adding meaningful default constraints (e.g. 0-40mm) for the StdQC configuration.

tkeffer:
Hello and thank you for your pull request!

The right way to do this is to use the function wxformulas.calculate_rain(), which basically does what you're doing, except in a module that can be used by other drivers. Want to try that?

-tk

maniac103:
Good point, done. Do you have any thoughts on adding a sane range for rain (e.g. 0...40mm or something) to the default configuration of StdQC?

tkeffer:
Merged.

Adding rain to StdQC seems reasonable to me.

PR #51, opened 2015-07-10
Add an optional TCP transport for WMR100.

maniac103:
Since my weather station doesn't reside in proximity of the device I run weewx on (and I'm not able to place it that way), I connected the weather station to my router running OpenWRT and wrote a simple tool ([1]) which processes the USB HID reports and generates a TCP stream out of them. The code in this pull request then allows the WMR100 driver to read the data from the TCP stream instead of the USB interface. Everything is still set up for USB by default, the TCP interface is purely optional.

I'm not sure whether you'll be willing to accept this as it depends on having an additional piece of software running, but I figured it doesn't hurt to ask anyway ;-)

[1] https://github.com/maniac103/wmr-collector/blob/master/forwarder/src/forwarder.c

### DIFF
--- a/bin/weewx/drivers/wmr100.py
+++ b/bin/weewx/drivers/wmr100.py
@@ -25,6 +25,7 @@
 
 import time
 import operator
+import socket
 import syslog
 
 import usb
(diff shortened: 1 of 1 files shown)


tkeffer:
Hello, and thank you for your pull request!

Unfortunately, this feels very specialized. Indeed, you're the first person I've encountered that's wanted to do this. 

I can see the utility of a generalized TCP transport layer that all the drivers could use, but I'm not enthusiastic about adding this complexity to every driver on the chance that a user would want to access the hardware through a TCP connection. Can we come up with a more general solution? Perhaps a "TCP driver" that talks to piece of hardware, using some on-the-wire protocol?

maniac103:
When introducing a generic abstracted TCP transport, one would need to do the parsing in the USB-to-TCP daemon though, which is what I wanted to avoid.

No big deal though, replacing the single file in the installation is no real problem. Good thing weewx is written in Python :)

matthewwall:
no need to replace the file.  just put the modified file as bin/user/wmr100.py then change one line of your weewx.conf:  driver = user.wmr100 instead of driver = weewx.drivers.wmr100

that makes it much easier to see how your customizations differ from stock, even across upgrades.

PR #53, opened 2015-07-22
Implement $span tag

gjr80:
Implemented $span tag to give stats over some timespan immediately before time of report. Length of timespan is set by parameters $time_delta (in seconds), $hour_delta (in hours), $day_delta (in days) and $week_delta (in weeks) (I know this is slightly different to what you suggested Tom but since these additional parameters are well defined and easily implemented it seemed right to add them - I guess they may or may not be publicised). Parameters are additive. Unsure what $span should return when all parameters are 0 (ie timespan is zero length); current record or None?. Have implemented current record but it is easy enought to change.


tkeffer:
Gary,

Can you add a test for `$span`? It should test all the time deltas you added, as well as what happens if it's given a non-existent observation type (e.g., `$span.foobar.min`).

Add the tags to `/bin/weewx/test/test_skins/StandardTest/index.html`, and the expected results to `/bin/weewx/test/expected/index.html` and `bin/weewx/test/expected/metric/index.html`.

gjr80:
Ok, never looked at any of the test suite before. I can see what to do in test_skins but what of the expected results? Seems the test suite just runs a simulator for 48hrs, won't the expected results vary (except the deliberate errors) throughout the run or are the test results as at 48hrs or are the figures just indicative? Or of course have I not read enough :)

tkeffer:
The simulator test run always uses the same start time, so it always ends up with the same results. Having said that, there may be a glitch with timezones. 

Give them a try. From the main directory type `make test`.

gjr80:
`make test` errors aside I have put together a first cut table for $span. It does what is expected; however, cheetah aborts the template when I use `$span($time_delta=3600).foobar.max`. I initially looked at the 'Nonexistent type:' row under `$day` but I see that only does `$day.heatindex.max`, not quite the same (more on this later). I then realised a better comparison is `$hour`, though there is no equivalent 'foobar' test for `$hour`. I put in `$hour.foobar.max`. Same result as `$span` with foobar; it causes cheetah to abort. 

Looking at the code I see that `$span, $hour, $day` etc end up calling getAggregate in their manager, the difference being since `$span` and `$hour` cannot use the daily summaries they use Manager.getAggregate whereas `$day` uses DaySummaryManager.getAggregate. I gather DaySummaryManager.getAggregate is a little more graceful in its handling of unknown observation types? If nothing else I guess I can say `$span` handles unknown ob types equally as well as `$hour` :)

Getting back to 'Nonexistent type:' in the `$day` table; is `$day.heatindex.max` better described as an existing type but with no data? As far as I could see heatindex exists in the test archive and has a daily summary.

tkeffer:
All true. OK, forget the $span.foobar.max test. That will require the
introduction of an intermediate object.

-tk

gjr80:
Must have been a touch of dyslexia, PYTHONPATH now working fine! Have added some indicative `$span` tests to `index.html.tmpl` and expected results files, the tests exercise all of the parameters particular to `$span`. Found the templates test was giving me errors and if my interpretation of the test results was correct there were a few issues with `\r\n` appearing/not appearing in the `index.html`, `bymonth.txt` and `bymonth.txt` expected results. Have included updated versions that now pass `test_templates.py`. Not sure if the files were indeed in disagreement or if it is me, if me I guess you can kill the updates.

tkeffer:
One last task needs to be done: Include the new tag $span in the documentation.

gjr80:
Happy to put some words in the docs, realise it's just HTML but is there any particular app or method for editing it?

tkeffer:
Matthew and I have been using WebStorm recently. You can use other tools
(emacs?), but WebStorm does a nice job of formatting.

-tk

Fat-fingered from my Android

PR #60, opened 2015-08-05
Added rsync patch from Kenneth Baker

W0CHP:
- Option to specify SSH key to use for rsync
    - Option to enable ssh/rsync compression
    - Example weewx.conf [[RSYNC]] skin/stanza config options:
        compress = True
        ssh_options = -i /home/weewx/weewx-id_dsa

### DIFF
--- a/bin/weeutil/rsyncupload.py
+++ b/bin/weeutil/rsyncupload.py
@@ -1,7 +1,10 @@
 #
-#    Copyright (c) 2009-2015 Tom Keffer <tkeffer@gmail.com>
+#    Copyright (c) 2012 Will Page <compenguy@gmail.com>
+#    Derivative of ftpupload.py, credit to Tom Keffer <tkeffer@gmail.com>
 #
 #    See the file LICENSE.txt for your full rights.
+#
+#    $Id: rsyncupload.py 2766 2014-12-02 02:45:36Z tkeffer $
 #
 """For uploading files to a remove server via Rsync"""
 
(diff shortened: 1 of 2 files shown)


tkeffer:
The test will only work if the configuration file has `ssh_options` defined

tkeffer:
Can you fix the problem with `ssh_options`?

W0CHP:
Yep - I'll work on that! My family just moved, and we're getting everything settled over here ;-)

tkeffer:
Any progress?

W0CHP:
Sorry for the delay Tom.

OK - so what would you like the desired behavior to be, if the `sshd_options` are not defined? Bail and print an error, or continue as always without the options defined, where the `sshd_options` are simply appended to the  `cmd = ['rsync']` only if defined?

tkeffer:
Sorry, I wasn't very clear with my comment. 

My concern is the test program at the end of the file. The lines

    rsync_upload = RsyncUpload(
                           rsync_dir,
                           config_dict['StdReport']['RSYNC']['path'],
                           config_dict['StdReport']['RSYNC']['server'],
                           config_dict['StdReport']['RSYNC']['user'],
                           config_dict['StdReport']['RSYNC']['port'],
                           config_dict['StdReport']['RSYNC']['ssh_options'],
                           config_dict['StdReport']['RSYNC'].as_bool('compress'))
    rsync_upload.run()

will raise an exception unless all options are defined in `weewx.conf`. One way I've dealt with this elsewhere is to use:

    rsync_upload = RsyncUpload(
                           rsync_dir,
                           **config_dict['StdReport']['RSYNC'])
    rsync_upload.run()

W0CHP:
Ahhhh so the `**kwargs` idiom is alright with you, rather than hard-defining possible config dictionary items?

W0CHP:
Tom, see commit fad4956 on this pull request. :-)

tkeffer:
The unpacked dictionary idiom is used throughout weewx. It allows a user to pass arguments weewx is not aware of to underlying libraries.

Thanks for adding SSH support to this, Ken!

PR #64, opened 2015-08-24
Add '--transfer' option to wee_database utility

gjr80:
Add '--transfer' option to wee_database to transfer a weewx 'archive' from one database to another database of any type supported by weewx. In essence added experimental transfer_db.py capability to wee_database. '--transfer' requires mandatory '--dest-binding' parameter. Included '--dry-run' parameter to do everything except the actual transfer. 
Also reformatted usage display along functional lines as in wee_extension.

### DIFF
--- a/bin/wee_database
+++ b/bin/wee_database
@@ -14,20 +14,42 @@ import time
 
 import user.extensions      #@UnusedImport
 import weedb
+import weeutil.weeutil
 import weewx.manager
 import weewx.units
 import weecfg
+import weewx
+
+from weeutil.weeutil import TimeSpan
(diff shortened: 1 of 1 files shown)


tkeffer:
Looks good. The only suggestion I have is that we should probably change `wee_database` so it can only do one verb at a time. If it fails, it should abort by calling `sys.exit("error msg")`.

gjr80:
Good idea, thought I could do it pretty easily with a couple of lists but optparse is not that co-operative. Nonetheless I have come up with a solution that does not seem too bad. Refer #65.

PR #65, opened 2015-08-25
wee_database now only accepts 1 'verb' at a time

gjr80:
Refer #64. Not as clean as I would like due to optparse limitations but
seems to do the job.

### DIFF
--- a/bin/wee_database
+++ b/bin/wee_database
@@ -55,6 +55,11 @@ usage = """wee_database --help
 epilog = """If you are using a MySQL database it is assumed that you have the
 appropriate permissions for the requested operation."""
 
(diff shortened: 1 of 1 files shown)


tkeffer:
Ooooh, that's some mighty fancy Python-foo!

gjr80:
I'll tell that to the folks at Stackoverflow :)

gjr80:
Was always told the job's not over till the paperwork is done so I guess I had best update the two wiki entries. Do I need to be given permission to edit the wiki or is having a GitHub account sufficient?

tkeffer:
No permission necessary. Just edit away.

-tk

PR #75, opened 2015-10-18
weewx debugging report generator

gjr80:
v0.2. Changes include:
- combined all reporting into a single report
- simplified command line usage
- added installed extension reporting

### DIFF
--- a/bin/wee_debug
+++ b/bin/wee_debug
@@ -1,0 +1,378 @@
+#!/usr/bin/env python
+#
+#    Copyright (c) 2009-2015 Tom Keffer <tkeffer@gmail.com> and
+#                            Matthew Wall
+#
+#    See the file LICENSE.txt for your rights.
+#
+""" Generate weewx debug info """
+from __future__ import with_statement
+
+import optparse
+import os
+import os.path
+import platform
+import sys
+import syslog
+
+from configobj import ConfigObj
+
+# Find the install bin subdirectory:
+this_file = os.path.join(os.getcwd(), __file__)
+this_dir = os.path.abspath(os.path.dirname(this_file))
+bin_dir = os.path.abspath(os.path.join(this_dir, os.pardir))
+
+# Now that we've found the bin subdirectory, inject it into the path:
+sys.path.insert(0, bin_dir)
+
+# Now we can import some weewx modules
+import weewx
+import weecfg
+
+from weeutil.weeutil import TimeSpan, timestamp_to_string
+from weewx.units import unit_nicknames
+from weecfg.extension import ExtensionEngine
+
+VERSION = weewx.__version__
+WEE_DEBUG_VERSION = '0.2'
+
+# keys/setting names to obfuscate in weewx.conf. Can add additional string
+# elements to list if required
+KEYS_TO_OBFUSCATE = ["password", "user", "username", "apiKey"]
+
+# weewx archive field to use as the basis of counting number of archive records.
+# Can be any field that is supported by manager.getAggregate() count aggregate
+# ie must be an obs_type, cannot be 'dateTime', 'usUnits'. 'outTemp' chosen as
+# it is almost always used/available.
+COUNT_FIELD = 'outTemp'
+
+# Redirect the import of setup (needed to get extension info)
+sys.modules['setup'] = weecfg.extension
+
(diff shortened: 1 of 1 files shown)


tkeffer:
Thanks, Gary.

I don't think it's necessary to "inject" the `bin` subdirectory in the `sys.path`. The utility `setup.py` needs to do this because it is not in the `bin` subdirectory. By contrast, all the other utilities, are in the `bin` subdirectory, so they can easily find their modules.

I would use `dateTime` for your `COUNT_FIELD`.

Your "case" statement for verbosity prints out "verbosity=2" for the second case, when I think you mean "verbosity=1"

Small style issue, but your function `obfuscateKey()` actually modifies its first argument, the configuration dictionary. It should make a copy first, then change that, or build a new dictionary. I know that in this case it's unlikely to cause a conflict (because `config_dict` is not used again), but someone could like what you've done and use it elsewhere.

gjr80:
>I don't think it's necessary to "inject" the bin subdirectory in the sys.path. The utility setup.py needs to >do this because it is not in the bin subdirectory. By contrast, all the other utilities, are in the bin >subdirectory, so they can easily find their modules.
My bad, teach me to develop a utility for /bin in /bin/user !

>I would use dateTime for your COUNT_FIELD.
Had dateTime originally but the method (getAggregate) I used to count records failed because my 'manager' was of class DaySummaryManager and used the daily summaries to 'count', hence the less the optimal 'outTemp'. Your comment rang a bell in my brain that I need to use the getAggregate method from class Manager, the parent of DaySummaryManager. So super() fixed that, though it is slower (to be expected but not overly slow on my 280000 odd records in my test archive, not sure how it will be with some of the bigger archives but guess this is aimed at those juststarting out - and besides this is a standalone utility not weewx mainstream so a little delay can be tolerated). Would appreciate a look over my 'super()' code, new territory for me.

>Your "case" statement for verbosity prints out "verbosity=2" for the second case, when I think you >mean "verbosity=1"
The proof reader has been fired!

>Small style issue, but your function obfuscateKey() actually modifies its first argument, the >configuration dictionary. It should make a copy first, then change that, or build a new dictionary. I >know that in this case it's unlikely to cause a conflict (because config_dict is not used again), but >someone could like what you've done and use it elsewhere.
That's a bit embarrassing, I knew that but still did it anyway. As for a solution, I have just been sitting here working on a couple of options. (1) using copy.deepcopy, seems to be a bit slow and feels a little like a cheats way out. (2) building the result element by element as we go, does not seem a simple process as we need to build a config_obj not a plain vanilla dict so we can use the .write method in the wee_debug output and (3) leave it as it is, not really a solution. Not sure what to do..

tkeffer:
I don't know it for sure, but I'd bet that `count(dateTime)` would be pretty fast because `dateTime` is indexed. 

Nothing wrong with deepcopy. 

BTW, speed is not a big concern with a utility like this.

gjr80:
Ok, took on board earlier comments and have rolled in deepcopy. There is a defined pause for a few seconds while deepcopy does it thing but its liveable. On another matter, put a few time checks in and dateTime on archive was actually quicker than outTemp on daily summaries!

My git is not that good, think this pull request has been correctly updated with v0.3, if not let me know. Am in the process of adding a few words to user manual but that will come later.

tkeffer:
A deep copy of the `config_dict` structure takes a few seconds? No way! There must be something else happening.

gjr80:
Ok, perhaps I have lept to conclusions here, the only effective change was doing the deepcopy of config_dict and there is a definite pause for seconds so I don't have any hard evidence. My weewx.conf is a big bigger than the norm but not substantially so. I will sit down tonite and put a few timers in the code and see what I can come up with.

gjr80:
Hmm, right symptoms, wrong cause... obfuscateKey() is recursive and calls itself around 33 times on a plain vanilla weewx.conf. obfuscateKey() takes 4 odd seconds to process a plain vanilla weewx.conf, my bloated weewx.conf was taking 13 odd seconds.

tkeffer:
This really surprises me. 33 calls seems tiny. But, I don't have any contradictory data!

gjr80:
It recurses every time it finds a nested dict and I just flicked through the standard weewx.conf and counted 32... so seems about right. Clearly far more than 32 entries though.

PR #80, opened 2015-11-24
Fixed errors in obfuscateKey()

gjr80:
Fixed logic errors, added comments and fixed some plain bad programming!

### DIFF
--- a/bin/wee_debug
+++ b/bin/wee_debug
@@ -28,12 +28,14 @@ from weewx.manager import DaySummaryManager
 from weewx.units import unit_nicknames
 
 VERSION = weewx.__version__
-WEE_DEBUG_VERSION = '0.3'
+WEE_DEBUG_VERSION = '0.4'
 
-# keys/setting names to obfuscate in weewx.conf. Can add additional string
-# elements to list if required
-KEYS_TO_OBFUSCATE = ["password", "user", "username", "station", "apiKey", 
-                     "api_key"]
+# keys/setting names to obfuscate in weewx.conf, key value will be obfuscated if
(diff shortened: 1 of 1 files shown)


matthewwall:
you should probably obfuscate anything called 'token' as well (emoncms, seg)

for twitter, obfuscate these:
        app_key = APP_KEY
        app_key_secret = APP_KEY_SECRET
        oauth_token = OAUTH_TOKEN
        oauth_token_secret = OAUTH_TOKEN_SECRET

for thingspeak: 'api_key'

there is also the case where a url is specified, e.g. the mqtt extension.

mqtt://username:password@example.com:1883/

weathercloud uses 'id' and 'key'

gjr80:
Done, added mesoWx keys as well. For mqtt I added server_url so I could keep filtering on dict keys only rather than dict values; server_url does not seem to be used elsewhere. After adding many fake keys to my weewx.conf it seems to pick them all up ok. I guess if there are false positives then we ask the user for more info; false negatives then the user should check before they post.

ZXant:
ThingSpeak Extension weewx.conf install patch and Extension howto/wiki has mistakes:

1) weewx.conf, wee_extension install writes:

[StdRESTful]

[[ThingSpeak]]
     TOKEN = XXXXXXXX

2) Extension howto/wiki:

[StdRESTful]
    [[ThingSpeak]]
        api_key = API_KEY


also

[StdRESTful]
    [[ThingSpeak]]
        api_key = TOKEN



So, 3 versions mixed. While the right context is the latest, ie:


[StdRESTful]
    [[ThingSpeak]]
        api_key = TOKEN #(Write to Token)

PR #86, opened 2015-12-18
Improve WS1 driver code

Talon1024:
Add code to the WS1 driver that allows the station to be connected via TCP/UDP, and separates the Station class into classes for data parsing, serial connection, and TCP/UDP connection.

This new code may need further testing and modification before it is ready to be added to the main code base.


Talon1024:
It should be noted that this code is already in use on our weather station, and I am getting no errors so far.

matthewwall:
kevin, nice work so far.  

apparently the ws1 do not behave *exactly* like an ultimeter (see the comments in the ultimeter.py code for some details about how the ultimeter differs from the ws1).  also, the ws1 sometimes sends CR-LF, other times just CR (or just LF - i do not recall). anyway, as with most hardware, its good to let things run a good while.

what kind of serial-to-tcpip hardware are you using?  is it doing udp broadcasts? or is it in tcp server mode? or tcp client mode?

m

Talon1024:
Thanks for the comment.

Our Serial-to-TCP/IP setup is described in detail [here](https://groups.yahoo.com/neo/groups/ads-ws1/conversations/messages/1850).

My changes to the driver code make it operate in TCP (or UDP) client mode. The serial-to-ethernet adapter operates in TCP server mode.

Talon1024:
I believe my code is stable and robust at this point. We've had weewx running for some time with the code in its current state, got a few timeouts here and there, an unexpected buffer length error once, but it hasn't caused weewx to crash or throw error after error. In addition, I fixed a bug in the original driver code, which failed to consider negative Fahrenheit indoor/outdoor temperatures.

tkeffer:
Thanks, Kevin.

Matthew is going to have to take a look at this one. Unfortunately, he is without internet for a couple weeks, so it may take a little time.

bin/weewx/drivers/ws1.py
```diff
@@ -33,8 +34,11 @@ def confeditor_loader():
 METER_PER_FOOT = 0.3048
 MILE_PER_KM = 0.621371
 
-DEFAULT_PORT = '/dev/ttyS0'
-DEBUG_READ = 0
+DEFAULT_SER_PORT = '/dev/ttyS0'
+DEFAULT_TCP_ADDR = '192.168.36.25'
+DEFAULT_TCP_PORT = 3000
+PACKET_SIZE = 50
+DEBUG_READ = False
```
matthewwall: DEBUG_READ should be an integer, not a boolean.  the convention has been to use an integer so that verbosity levels can be specified, e.g., 0 for nothing, 1 for some, 2 for more, etc.

bin/weewx/drivers/ws1.py
```diff
@@ -51,26 +55,45 @@ def logerr(msg):
...
-        self.port = stn_dict.get('port', DEFAULT_PORT)
+        con_mode = stn_dict.get('mode', 'serial').lower()
+        if con_mode == 'serial':
+            self.port = stn_dict.get('port', DEFAULT_SER_PORT)
+        elif con_mode == 'tcp' or con_mode == 'udp':
+            self.port = stn_dict.get(
+                'port', '%s:%d' % (DEFAULT_TCP_ADDR, DEFAULT_TCP_PORT))
+        else:
+            # exit(3)
+            pass
```
matthewwall: consider this instead:

if con_mode == 'tcp' or con_mode == 'udp':
    self.port = ...
else:
    self.port = stn_dict.get('port', DEFAULT_SER_PORT)

if you do not like that kind of quiet backward compatibility, then fail hard with an exception indicating that 'mode' is a required parameter.

bin/weewx/drivers/ws1.py
```diff
@@ -221,6 +203,209 @@ def _decode(s, multiplier=None, neg=False):
...
+                    except (socket.error, socket.timeout), ex:
+                        raise weewx.WeeWxIOError(ex)
+                    if DEBUG_READ: loginf("buf: %s" % ' '.join(
+                        ['%02X' % ord(bc) for bc in buf]))
+            try:
+                buf += self.net_socket.recv(
+                    PACKET_SIZE - len(buf), socket.MSG_WAITALL)
+            except (socket.error, socket.timeout), ex:
+                raise weewx.WeeWxIOError(ex)
+        if DEBUG_READ: loginf("buf: %s" % buf)
```
matthewwall: any DEBUG_READ messages should be logdbg, not loginf.  also, prolly safer to keep it on two lines instead of inlining a conditional.

Talon1024:
So, where would the messages outputted from logdbg appear? The last time I remember trying to use logdbg, the debug messages didn't appear in the system message log. (/var/log/messages)

bin/weewx/drivers/ws1.py
```diff
@@ -221,6 +203,209 @@ def _decode(s, multiplier=None, neg=False):
...
+                StationData.validate_string(buf)
+                return buf
+            except (weewx.WeeWxIOError), e:
+                loginf("Failed to get data for some reason: %s" % e)
+                self.rec_start = False
+
+                # WeeWx IO Errors may not always occur because of invalid data
+                # DEBUG_READ = True
+
+                # This causes an exception
```
matthewwall: not sure what you mean here.  is get_readings or validate_string missing an exception?  or is the 'variable referenced before assignment' coming from StationInet.get_readings_with_retry?

Talon1024: WeeWx IO errors could occur because of socket errors. (timeout, address issue, or other issues)

matthewwall:
nice work here kevin.  a few more things:

- perhaps put 'import serial' into the two methods that actually use the serial interface (open and get_readings_with_retry).  that way someone with a socket-connected device does not have to install any serial packages.  (the vantage driver does this)

- i was hoping to figure out some way to abstract the ethernet-to-serial device into a class that we could put into weewx core so that it could be used with *any* device with a serial interface.  maybe at some point in the future...

btw, as tom noted i will have no internet again for awhile, so apologies in advance for delays in response.  we're at cucumber beach (belize) getting water at the moment, but soon we'll be back out to the reef and cayes.

m

bin/weewx/drivers/ws1.py
```diff
@@ -51,26 +55,45 @@ def logerr(msg):
...
         loginf('driver version is %s' % DRIVER_VERSION)
-        loginf('using serial port %s' % self.port)
+        loginf('using %s port %s' % (con_mode, self.port))
         global DEBUG_READ
         DEBUG_READ = int(stn_dict.get('debug_read', DEBUG_READ))
-        self.station = Station(self.port)
+        if con_mode == 'serial':
+            self.station = StationSerial(self.port, timeout=timeout)
+        elif con_mode == 'tcp' or con_mode == 'udp':
+            self.station = StationInet(self.port, con_mode, timeout=timeout)
```
matthewwall: for backward compatibility, please consider this instead:

if con_mode == 'tcp' or con_mode == 'udp':
    self.station = StationInet(...)
else:
    self.station = StationSerial(...)

bin/weewx/drivers/ws1.py
```diff
@@ -51,26 +55,45 @@ def logerr(msg):
...
+            self.port = stn_dict.get(
+                'port', '%s:%d' % (DEFAULT_TCP_ADDR, DEFAULT_TCP_PORT))
+        else:
+            # exit(3)
+            pass
         self.max_tries = int(stn_dict.get('max_tries', 5))
         self.retry_wait = int(stn_dict.get('retry_wait', 10))
         self.last_rain = None
+        timeout = int(stn_dict.get('timeout', 3))
         loginf('driver version is %s' % DRIVER_VERSION)
```
matthewwall: i usually put the 'driver version is ...' line as the very first thing in __init__
that way if there are any failures you always know what version someone is using

bin/weewx/drivers/ws1.py
matthewwall: declare buf outside the try block, e.g., buf = ''.  be sure to wrap the socket errors into WeeWxIO errors otherwise weewx will die.  there should be a generic socket.error or some such which you can catch then raise WeeWXIOError(e)

matthewwall:
set debug=1 in weewx.conf to see the logdbg messages

Talon1024:
OK, I made some more changes to my driver code as per @matthewwall's suggestions.

Issue #12, opened 2015-03-11
$day tag should be iterable


matthewwall:
$week, $month, and $year tags are iterable (over the daily summaries)

$day tag should be iterable as well, but this will require iteration over data in the archive table

tkeffer:
Implemented in commit 48b7fda232044df6b5d6dd0a4b82f300781312f8, thanks to user Julen.

PR #87, opened 2015-12-24
Implement hours iterable

Zokormazo:
Implements https://github.com/weewx/weewx/issues/12


tkeffer:
This is a wonderful addition, something that's been requested many times.

There are just three parts missing:

1. A test that exercises the new feature in a template. It should do an aggregation over the hour period. 

    Something like

    ```
    #for $hour in $day
    $hour.dateTime $hour.outTemp.max
    #end for
    ```

    You can either add the test to `bin/weewx/test/test_skins/StandardTest/index.html.tmpl`, or create a separate template, much like `bin/weewx/test/test_skins/StandardTest/bymonth.txt.tmpl`. The expected results should be included in the `expected` directory, for both US Customary and Metric.

2. I suspect this will reveal a second problem: you have used a time context of `hour`, but there is no matching entry in `skin.conf`, under sub-section `[[TimeFormats]]`.

3. Documentation. Add the new iteration to the list of iteration types in section [Iteration](http://weewx.com/docs/customizing.htm#Iteration)

I'll take care of the Change doc.

Thanks in advance!

Zokormazo:
Ok, I'll code the rest ASAP.

Zokormazo:
PR updated with skin_test.

I don't know how to add the iterable on docs since there is no specific doc about the rest of iterables, just a generic months iterable example

tkeffer:
Excellent!

Thanks so much, Julen.

I'll take care of the rest of the docs.

-tk

PR #93, opened 2016-01-16
Update restx.py

camattin:
Indoor temp and humidity should be reported to wunderground.

### DIFF
--- a/bin/weewx/restx.py
+++ b/bin/weewx/restx.py
@@ -658,6 +658,8 @@ class AmbientThread(RESTThread):
                 'barometer'   : 'baromin=%.3f',
                 'outTemp'     : 'tempf=%.1f',
                 'outHumidity' : 'humidity=%03.0f',
+                'inTemp'      : 'indoortempf=%.1f',
+                'inHumidity'  : 'indoorhumidity=%2.0f',
                 'windSpeed'   : 'windspeedmph=%03.1f',
                 'windDir'     : 'winddir=%03.0f',
                 'windGust'    : 'windgustmph=%03.1f',


tkeffer:
Hi, Chris

This has been discussed before, but not everyone thinks it's a good idea, because it is a strong hint whether someone is home or not.

For that reason, it should be user selectable. Perhaps something in the `[StdRESTful]` section that allows a user to exclude certain observation types. 

```
[StdRESTful]
  [[Excludes]]
    inTemp = True
    inHumidity = True
```

This would exclude `inTemp` and `inHumidity` for *all* uploaders (not just Wunderground).

Let me know what you think.

camattin:
Being optional is quite fine.  Should it be in [[Excludes]] or have it as a flag in [[Wunderground]]?  Thinking something along the lines of:

[StdRESTful]
...
    [[Wunderground]]
        enable = true
        station = KNCHIGHP18
        rapidfire = True
        reportindoor = True

tkeffer:
When making a change, it's nice to generalize things as much as possible. 

If this is a privacy issue with Wunderground, it's likely to be a privacy issue for other uploaders as well. Hence, the motivation to put the exclusions in their own, special, section, where they apply to _all_ uploaders. This generalizes to additional uploaders.

No need to make a separate option `reportindoor`, when we have the perfectly good key `inTemp` and `inHumidity` at hand. This generalizes to additional types.

It would also be useful to override the override. Something like:

```
[StdRESTful]
  [[Excludes]]
    inTemp = True
    inHumidity = True
  [[Wunderground]]
    enable = true
    station = KNCHIGHP18
    [[[Excludes]]]
      inTemp = False
```

camattin:
Gotcha.  I'll have to leave it someone else to code for the Excludes, though. :)

matthewwall:
tom, it would be helpful to establish/make-obvious the pattern for inherited sections such as this 'Excludes'.  perhaps something like this for restful service 'foo':

    # set defaults for everything the 'foo' needs
    foo_dict = {}
    # get values from the base class configuration
    foo_dict.merge(config_dict['StdRestful'].get('Excludes', {}))
    ...
    # override with anything for the foo
    foo_dict.merge(config_dict['StdRestful'].get('foo', {})

the pattern shows up in StdReport and its progeny, restful and its progeny, and probably elsewhere.  currently it is not done consistently.

we should also consider whether unrecognized variables shoud cause a failure/warning or be silently ignored.

matthewwall:
also, be sure that any configuration options that specify database fields are generic.  some people use just the wview schema, but others have extended the wview schema, and some use a schema that is not at all related to wview (e.g., the klimalogg schema or schemas for power monitoring).

the Excludes section is generic enough for this.  (however it means that you cannot name a restful service 'Excludes')

tkeffer:
Chris,

Did you want to pursue this? Or, should I close the PR?

-tk

tkeffer:
Inactive, so I'm closing this. Can reopen if necessary.

camattin:
Was out of town and unable to respond earlier.

This issue should be significant to anyone that reports data to WU and also uses the Wunderstation app on an iPad. Perhaps this sliver of folks is smaller than I imagined it might be.

That said, I don't have the expertise to code this in properly, seems I'll just need to keep track of my modified restx.py during upgrades.

PR #98, opened 2016-02-05
New data broadcasting extension and hardware driver

kaklik:
Hello, here is my network broadcasting plugin and data viewer utility. 
The system was tested with real Anema WS981 weather station hardware.


tkeffer:
Looks very interesting.

The driver looks like it can be immediately incorporated into the code base. Nice that you included a configuration editor with it! However, there are several places where it needs to be documented.
- All config options should be documented in the User's Guide under [`[Station]`](http://weewx.com/docs/usersguide.htm#[Station])
- How it works with [`wee_device`](http://weewx.com/docs/usersguide.htm#configuring_hardware) should be documented
- What types it emits should be listed in the [Station data appendix](http://weewx.com/docs/customizing.htm#station_data) of the Customization Guide

The weeview utility looks interesting. I've always liked curses! I'm not familiar with PyMeteostation. A short README would be useful so your prospective users would know where to use it.

The BroadcastService extension is something that others have asked for. However, it has two problems:
- It looks like you have hardwired in the IP address ('224.1.1.1'). This should be selectable from the config file.
- As written, the broadcast to the network is being done in the main thread. This has the risk of slowing down the thread. I know you have a timeout in there, but it is a very long one. Even if you made it very short --- a short second or two --- that would be too long to wait. Instead, you should launch a separate thread, connect a queue between the main thread and the new thread, and then put the LOOP packets in the queue. This is the basic architecture all the RESTful uploaders use. No reason why yours would be any different. Take a look in the file `weewx/restx.py` for many examples.

Your PR has got me wondering. Traditionally, extensions have been documented on the Wiki, with a link to a download site. They have not been included with the code base. What I'm wondering is whether we should have a separate weewx GitHub repository just for contributed extensions, such as yours. 

Any thoughts (from anyone) on the topic?

matthewwall:
jakub,

you can get rid of _augment_packet.  the driver should not set windDir to None; StdWXCalculate enforces that policy.

you might want to make the driver install to the user directory instead of the drivers directory (at least until we figure out a policy for when to incorporate extensions).

probably no need for 'manual' instructions in the readme any more.

as for contributed extensions...

the approach so far has been to keep extensions out of the weewx distribution.  but how do we decide when an extension (driver, uploader, skin?) should be assimilated and become part of the core?

for drivers we could have a popularity contest.  when X instances of the driver show up on the map for at least a month, then put the driver into the core. :)

i'm inclined to avoid a central repository for extensions.  but if someone committed to manage it that opinion might change.

m

tkeffer:
jakub

Did you want to pursue this?

-tk

tkeffer:
Nothing heard. Closing off the PR, but can reopen if it shows signs of life.

PR #105, opened 2016-02-21
Mobile Reading Improved

davies-barnard:
CSS and template updates to make the docs easier to read on different devices.   I have also moved the JS to the footer so content loads first.   All CSS is now in the css file.

Attempt 2 - Hope this is correct this time.   Not sure why it happened last time because I had pulled the master only a couple of days earlier!


tkeffer:
I appreciate the effort you are putting into this, Chris. The docs were built on the assumption that they were going to be  viewed on a bigger screen --- supporting mobile devices is an important addition.

But, honestly, I think we're improving readability on mobile devices at the expense of the desktop. 

Specifics:
- The font size in the tables is now smaller. This makes them harder to read, and mismatches the rest of the text.
- Same with the TOC.
- The TOC frame is also much skinnier, causing lines to wrap that previously did not have to.
- The highlighted entry in the TOC does not always reflect where the main body is. Try this: change from "more detail" to "some detail," then scroll. The highlight pops up to the top, then stays there.

I am not a UI expert, but the right way to do this is to use responsive design principles, selecting things like font and table sizes on the basis of browser width, etc.

The docs were largely written with semantic HTML markup in mind, so it should be a matter of fiddling with the CSS files and getting them right. 

If you're not sick of this project, would you mind taking one more look?

davies-barnard:
Tom,

Happy to do so but would appreciate a closure on the discussion regarding minimising user docs before I do.   There must be an all round better solution that better fits weewx as it is today.  In saying this I make no criticism of the system used until today as they have always sufficed.    

I can use media queries to adjust between mobile and desktop no problem but it might be nice from a keeping documents up to date easily point of view to separate content from structure a bit more?  Anyway, I welcome your thoughts.

Chris

tkeffer:
I don't see the docs changing in any major way soon. 

Is there any particular place where you see content needs to be separated from structure? I thought we did a pretty good job of that. Looking through your PR, it looks like the only changes within the docs were moving where the JS files get read and adding a couple of minor classes (e.g., `main` and `sidebar`). It didn't seem to require any content changes.

How about proceeding in two phases:
1. Change the structure of the docs. This would be moving the JS file reads, adding the classes you need. If there is any place where content and structure needs to be separated, this would be when to do it.
2. Fiddle with CSS.

This way, once No. 1 is done, development work can proceed without fear of stepping on your efforts.

davies-barnard:
Ok,

If you don't see the docs changing anytime soon I'm happy.   I should be able to resolve the issues you raised with a couple of media queries.   I'll try and do it in a bit.

davies-barnard:
Hi Tom,

I'd be interested to know what people think now.  

http://weewx.davies-barnard.uk/docs/upgrading.htm

It resizes for smaller screens and then vanishes for very small screens.   Fonts are all done via %ages now.   Fixed most issues with the not keeping up with scrolling but don't believe its any different to the current version.  Been off work this week ill so have not had time or energy to look at it.

tkeffer:
I like the link back to the docs index.

The highlighted TOC entry now moves with the text.

Displays very nicely (albeit not very smoothly) on the tiny 4.5" screen of my Moto X, as well as on my Nexus 7.

But, why the decision to use smaller fonts in the tables (specifically, in `<tr>`)?

davies-barnard:
Hi Tom,

One of the points raised in feedback from the user group was tables that overflowed the page and requiring horizontal zoom so I have been trying to limit that.    I fixed the worst one (but I don't think you have noticed what I have done yet ;)  http://weewx.davies-barnard.uk/docs/customizing.htm#The_database

I have just adjusted the table font size to 90% normal for large and medium screens reducing it to 80% (same as sub menus) for small and very small screens.   You can see the media queries at the top of the CSS file.

tkeffer:
Well, if I have to choose, and I guess I do (!), I'd go for scrolling, especially if that's the only offending table. The font size transition is just too jarring for my tastes. 

Another option is to take out a column or two of that table: they are there just for illustrative purposes.

davies-barnard:
Ok,  

I have removed font size adjustment from the larger screen sizes and kept it at 80% for screens under 961px wide i.e small.

I have left the table for you to edit as you feel fit at a later date.

tkeffer:
Really close.

The remaining issue is that the default font size _for the body_ is quite large. It's set to `100%`, which on my desktop works out to a really large 16pt. It was 11pt. 

I would suggest leaving it at 11pt. The media scalings should scale it appropriately for smaller screens.

Then again, I'm not a CSS expert.

davies-barnard:
Well,  I'm not convinced by leaving it as 11pt because its not the done thing ...  https://www.w3.org/Style/Examples/007/units.en.html

I have set it to 90% which equals 14.4px or 14.667 pt but this should also adapt to screen sizes.

tkeffer:
LOL! Ok, you're the CSS expert. :-)

But, I'm curious, 90% of what?

davies-barnard:
I would not consider myself a CSS expert. I've just a had enough battles over browser, screen, device compatibility.

90% of your browser/system default so it should fit in with your screen settings etc.

tkeffer:
Ah, now I get it.

Have you pushed your final changes to your pull request?

-tk

matthewwall:
chris, is there a place where i can view your changes?

matthewwall:
when i try to connect to this:
http://weewx.davies-barnard.uk/docs/

i get this:

Requested Range Not Satisfiable

None of the range-specifier values in the Range request-header field overlap the current extent of the selected resource.
Apache/2.4.7 (Ubuntu) Server at weewx.davies-barnard.uk Port 80


macosx 10.6.8
firefox 39.0.3
(i know, i'm way behind the latest shiny)

matthewwall:
double-scrollbars

i see double scrollbars

tkeffer:
No problems with scroll bars on the four combos I tested:

Ubuntu 12.04: with Chrome 48.0.2564.116 and Firefox 44.02
Mac OS X El Capitan 10.11.3: with Chrome 48.0.2564.116 and Safari 9.0.3

I did notice, however, that the docs index page
<http://weewx.davies-barnard.uk/docs/index.html> has the title "Upgrading
weewx"

-tk

davies-barnard:
Morning,

So my tests are the same as Toms except I can add Firefox 44.   

Matt,  I have prevented the scroll on the sidebar so you should only see it on the inner toc contents now.

Tom, I have updated the title and added favicon support which should work once I change permissions on my server and as default on yours.

I have updated my fork of weewx to the docs.   So I'm guess that once you pull it will bring those versions?

tkeffer:
Merged.

I'm not quite sure how the new index.htm file will interact with the website, but we can figure that out later.

davies-barnard:
Thanks Tom.     

And a big thank you for your support and patience.

PR #107, opened 2016-02-25
Add some basic support for NMEA-0183 hardware.

jof:
Specifically, this was added to support the New Mountain Ultrasonic Weather Station.


tkeffer:
This is a great addition.

However, could I talk you into packaging it as an extension, then posting a link to it in the Wiki? There are two reasons
1. I love the idea of supporting NMEA, but, at this point, this is not a common enough station to include it in the weewx distribution, with all the testing, etc., that involves.
2. It would add an additional requirement, `pynmea2`.

Thanks for taking this on.

tkeffer:
jof?

tkeffer:
Closed for lack of activity. Can reopen if necessary.

PR #119, opened 2016-04-29
Fix missing humidity of the second sensor TE923

ngulden:
I use three outside sensors on my TFA Nexus. All sensors report temperature and
humidity. But the humidity of the second sensor is not recorded in weewx. I
found it missing in the driver and fixed the error there.

This fix works for my station and the humidity for the second sensor is back
and working.


matthewwall:
thanks for catching the typo.  fixed at commit 67bb6be

PR #120, opened 2016-05-05
Standard2 Beta Skin for comparison

davies-barnard:
Hi,

Here is Standard2 skin as it stands.   So you can pop around its structure etc.   I don't consider it ready so no worries about not merging it yet.


tkeffer:
A few comments. Keep in mind, I am *not* a UI expert!

1. Part of what makes a good looking page is clever use of fonts. The fonts used in the "Standard" page, particularly for the images, are kind of clunky. 

2. To my eye, the Almanac does not as attractive as the "Standard" almanac. Something about the black border.

3. Only the "current" page has links to the other pages.

4. When served from the file system, the "NOAA" reports seem to use an absolute URL, instead of a relative URL. I get
    ```
        file:///NOAA/NOAA-2015-08.txt not found
    ```

Looks good on mobile devices!

davies-barnard:
Nope this is all good.

Takes an age to poke around and bug fix everything. More hands make light
work of it and you all gonna have an opinion anyway so I might as well put
you to work doing something useful ;)

Fonts on images/charts are the standard ones - I am not playing with those
atm.

davies-barnard:
Hi Tom,

Ok,  had a quick poke around in reference to your comments.

1) it seems that the font being used by skin.conf for the images does not exist on my system!   Need to have a deeper look but I have switched the HTML font to helvetica for now - let me know what you think.
2) The border is no more.
3) Can you explain - I cannot see this.
4)  Fixed - sorry.  Had only checked remote paths.

davies-barnard:
Ok.  On the mac which I am using for testing PIL and fonts are a bit ...   

http://leancrew.com/all-this/2008/11/truetype-fonts-for-the-python-imaging-library/

I'll need to have a look on the old Pi when I get home.

tkeffer:
No activity.

PR #121, opened 2016-05-07
disable ssl verification in wunderfixer

vinceskahan:
This fixes wunderfixer to disable strict ssl verification (python 2.7.9)

### DIFF
--- a/bin/wunderfixer
+++ b/bin/wunderfixer
@@ -17,6 +17,9 @@ Weather Underground with the missing data.
 
 CHANGE HISTORY
 --------------------------------
+1.1.0 5/08/16
+Suppress strict ssl checking for wunderground http=>https
+
 1.0.0 8/16/15
 Published version.
 
@@ -61,6 +64,27 @@ from weeutil.weeutil import timestamp_to_string
 import weewx.manager
 import weewx.units
 import weewx.restx
+
+# wunderground.com quietly redirects http=>https and
+# python 2.7.9 turns on strict certificate verification
+# which breaks the original wunderfixer 0.5.2
+#
+# long discussion of the internals is at
+#     http://stackoverflow.com/questions/27804710/python-urllib2-ssl-error/27826829#27826829
+#
+# the following code is taken verbatim from
+#     https://dnaeon.github.io/disable-python-ssl-verification/
+
+import ssl
+try:
+    _create_unverified_https_context = ssl._create_unverified_context
+except AttributeError:
+    # Legacy Python that doesn't verify HTTPS certificates by default
+    pass
+else:
+    # Handle target environment that doesn't support HTTPS verification
+    ssl._create_default_https_context = _create_unverified_https_context
+
 
 usagestr = """%prog CONFIG_FILE|--config=CONFIG_FILE
                   [--binding=BINDING]
--- a/bin/wunderfixer
+++ b/bin/wunderfixer
@@ -111,7 +111,7 @@ proposes !!
        -->>      or corruption of your data !       <<-- 
 """
 
-__version__ = "1.0.0"
+__version__ = "1.1.0"
 
 # The number of seconds difference in the timestamp between two records
 # and still have them considered to be the same: 


gjr80:
Just wondering should our default action be to bypass verification? I realise we are only talking about WU and PWS data but... As was apparent from the forums yesterday some people are successfully verifying the certificate, some aren't (I suspect anyone on a Debian based district will have issues for a while yet). It would be a simple manner to add another command line parameter to wunderfixer, say --ignore_ssl_verification (I am known for inappropriate parameter names), that when used would bypass verification. That way those for whom verification works would likely continue to use it and for those for which it does not work there is a workaround. Same end result but a slightly different path.

matthewwall:
wget uses `--no-check-certificate`

matthewwall:
curl uses `--insecure` to do the same thing.  i prefer `--no-check-certificate`

matthewwall:
oh, and sshd has its own camelCase way of doing it, such as `NoHostAuthenticationForLocalhost`

vinceskahan:
Well wunderfixer only works with WU so I'm leaning toward it's maybe OK,
but I wouldn't object to a switch making it  optional either.

tkeffer:
I was finally able to reproduce the error by using a [Docker image of Jessie running Python 2.7.9](https://gist.githubusercontent.com/tkeffer/6ba9c52ab1bc8f35bdc73c6b26de6e5a/raw/9c5fadddea66012e5ee3f5e06e2ea75a4a01df5f/Dockerfile).

Then I was able to fix the error by doing the following

```shell
cd /usr/local/share/ca-certificates
sudo wget  http://www.tbs-x509.com/GTECyberTrustGlobalRoot2018.crt
sudo update-ca-certificates
```

I then tried a Docker image of Stretch, and wunderfixer worked fine out of the box. I then added the above certificate to Stretch, and wunderfixer continued to work fine.

Personally, I feel better about asking users to fix their certificates, rather than patching wunderfixer to ignore certificate validation.

vinceskahan:
Tom - that works great on a pi3 running the latest Raspbian.....
I'd be interested in learning how you figured out which cert was missing,
walking those trust chains is a bit painful.

I'd be a little uncertain about whether actually 'fixing' the problem is a
good idea however, personally I'd lean against some app's utility messing
with my operating system's certs, but how about if wunderfixer printed out
the steps to follow for the user perhaps ?

Would it be possible to perhaps add a "--wutest" switch that checks whether
your system can connect to WU via https ok, printing out the instructions
for how to fix the problem from the wunderfixer code ?   That would get you
out of the inevitable documentation hell perhaps.....

The exception is coming from the first URLError part here, but I couldn't
find a way in urllib2 to look for the cert verification error specifically,
looks like they just say "nope" and call it a URLError...

       try :
            # Hit the weather underground site:
            _wudata = urllib2.urlopen(_url)
        except urllib2.URLError, e :
            print >>sys.stderr, "Unable to open Weather Underground station
" + self.station, " or ", e
            raise
        except socket.timeout, e:
            print >>sys.stderr, "Socket timeout for Weather Underground
station " + self.station
            raise





On Sun, May 8, 2016 at 9:19 AM, Tom Keffer <notifications@github.com> wrote:

> I was finally able to reproduce the error by using a Docker image of
> Jessie running Python 2.7.9
> <https://gist.githubusercontent.com/tkeffer/6ba9c52ab1bc8f35bdc73c6b26de6e5a/raw/9c5fadddea66012e5ee3f5e06e2ea75a4a01df5f/Dockerfile>
> .
>
> Then I was able to fix the error by doing the following
>
> cd /usr/local/share/ca-certificates
> sudo wget  http://www.tbs-x509.com/GTECyberTrustGlobalRoot2018.crt
> sudo update-ca-certificates
>
> I then tried a Docker image of Stretch, and wunderfixer worked fine out of
> the box. I then added the above certificate to Stretch, and wunderfixer
> continued to work fine.
>
> Personally, I feel better about asking users to fix their certificates,
> rather than patching wunderfixer to ignore certificate validation.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/weewx/weewx/pull/121#issuecomment-217730576>
>



-- 
-----  vinceskahan@gmail.com ----

tkeffer:
Oh, I agree completely. The certification "fix" is something that I'd expect the user to do explicitly.

We could either put instructions in the Wiki or, as you suggest, have `wunderfixer` detect and offer up the fix instructions.

tkeffer:
By the way, if you go to https://www.digicert.com/help/ and type in www.wunderground.com, it shows that the WU SSL certificate is not signed by a trusted authority. I don't claim to understand all this SSL certificate mumbo-jumbo, but my conclusion is that this could be a WU problem, rather than a Jessie problem.

davies-barnard:
Hi Tom,

From my basic understanding of doing it with web servers for https that
suggests they have self signed the certificate rather than used an
authority although I find that rather surprising for wunderground so would
basically agree with your conclusion.

Chris

tkeffer:
If they have self-signed, then why would introducing a new root certificate fix things?

Or, are there two different problems?

-tk

davies-barnard:
I've no idea as you've reached the end of my knowledge but when I have seen
that error before self signing has been the cause.

vinceskahan:
My experience has been most recently puppetlabs.com using a cert windows
systems don't trust by default, but Linux systems do. Guessing they just
went cheap. Same solution.... download the missing cert and install it at
the os level

gjr80:
Again my limit of knowledge is well and truly exceeded but I thought one of the answers in this thread gave quite a plausible dissection and it would explain what has been see no I far, including the Jessie and Stretch behaviour. http://stackoverflow.com/questions/27804710/python-urllib2-ssl-error/27826829#27826829

wotid:
My even more limited knowledge.  A Google search lead me to a Mozilla developers thread where they discussed removing the certificate too. They removed GTE certificate because it was a 1024b certificate, broke a lot of https access to a lot of sites, put it back, contacted a lot of sites then removed the certificate again. 

I don't know how or why it affects my OS, but it appears to me that WU rely on a certificate that Mozilla canned a couple of years ago.

tkeffer:
Closed due to inactivity. We can reopen if this becomes an issue again.

Issue #17, opened 2015-03-24
implement cron-like report generation


matthewwall:
Look at Gary's (?) changes to allow cron-like report generation.

(from the TODO file)

gjr80:
Just wondering if there is still an appetite for cron like report generation. Attached email exchange is the last correspondence I have on the topic.
[Re weewx implement cron-like report generation (#17).txt](https://github.com/weewx/weewx/files/139156/Re.weewx.implement.cron-like.report.generation.17.txt)

To refresh, I developed some additional code that used 3 parameters that could be specified in `skin.conf`/`weewx.conf` against each report being generated. The parameters were `generate_dom`, `generate_dow`, `generate_time` that allowed the user to specify the day of month, day of week and time that a report will be generated. The parameters were cumulative and each could include multiple comma seperated values. I believe the net effect is to mimmick the standard cron file format but in a 'wordy' manner.

To summarise the state of the code. After the 1st cut of the code I refactored and put the logic that determines whether to generate the report or not in a seperate method (in `weeutil.py` for want of a better place) and added about 10 extra lines to `cheetahgenerator.py`. The code was soak tested for about 6 weeks on a simulator system running 25 odd reports testing various combinations of the 3 parameters and I believe it to operate without error.

I am not sure what weewx version I tested on but given it was April-May 2015 I am guessing 3.1, definitely v3.x. I don't know that there has been any great change to weewx report generation since this time, so I expect the code will still be functional. If there is still a desire to implement this I will go ahead and revalidate the code and submit.

tkeffer:
If it does not dramatically increase the complexity of the code and user experience, then I think there is still interest.

davies-barnard:
Hi,

Is there scope to include this in some refactoring of the stdreport/restful
sections so that skins are not compulsory etc?   All my ISSWatch, METWatch
and Spacewatch extensions would benefit from only being run once in every 1
or less.

Chris

gjr80:
@tk. Ok, I don't believe it does dramatically increase the code complexity. The complexity is in the parsing of the 'cron like parameters' but that is pushed down to a seperate method. The core report generator code (method name escapes me) is not substantially altered. I will integrate ithe code into a current clone, redo the soak testing and offer up a pull request.

@cdb. I am not overly familiar with your extensions, and I could not find them given a quick search this morning. CheetahGenerator already supports a stale_age parameter which runs a reports every x seconds. Maybe that is appropriate.

gjr80:
[...]

PR #122, opened 2016-05-15
Implement CRON like report generation

gjr80:
Refer issue #17, in particular later comments.

Renamed variables/class name to remove reference to the word CRON. Located class ReportTiming (formerly class ReportCron) in reportengine.py.


bin/weewx/reportengine.py
```diff
@@ -25,55 +26,92 @@
...
-    
+
     See below for examples of generators.
     """
-    
-    def __init__(self, config_dict, stn_info, gen_ts=None, first_run=True):
-        """Initializer for the report engine. 
-        
+
+    def __init__(self, config_dict, stn_info, record=None, gen_ts=None, first_run=True):
```
tkeffer: Would suggest adding the record parameter to the end, on the remote chance someone has called the initializer without using keywords

tkeffer:
this looks really great! Much cleaner approach than where we started.

tkeffer:
This will also set us up nicely for reports that need direct access to the last archive record, without going through the database.

matthewwall:
very nicely done, gary.  i took the liberty of making a few editorial fixes to the docs - at commit ce949f2.

PR #125, opened 2016-05-16
Fix for DMPAFT not working properly after summertime change

lheijst:
When the system has changed to summer time and the vantage console is presenting it's latest record in winter time there will be a difference of one hour when the records of the first hour of the summer change are stored to the weewx database, DMPAFT will ask for all records since the latest timestamp stored in the weewx database, but a record with this time stamp can't be found by the vantage console, resulting in presenting all available records starting with the eldest record. DMPAFT on his turn thinks it's job is finished and in fact nothing happened. The next archive period the same process will repeat until eventually all elder records than start summer time are 'scrolled out' or 'deleted from' the vantage logger memory. This fix implies that during the first hour of the summertime period no vantage records are saved to the weex database.

### DIFF
--- a/bin/weewx/drivers/vantage.py
+++ b/bin/weewx/drivers/vantage.py
@@ -615,9 +615,25 @@ class Vantage(weewx.drivers.AbstractDevice):
                                      weeutil.weeutil.timestamp_to_string(_last_good_ts)))
                     syslog.syslog(syslog.LOG_DEBUG, "vantage: Catch up complete.")
                     return
-                # Set the last time to the current time, and yield the packet
-                _last_good_ts = _record['dateTime']
-                yield _record
+                # Fix for start summertime problem
+                # When the system clock is changed to summer time and the vantage console present the data of a record
+                # with a time stamp in winter time, the first hour of the summer time the hour presented by the console
+                # doesn't match with the hour presented by the system (console hour = system hour - 1)
+                # When such a record would be stored in the database the DMPAFT mechanism won't be working anymore
+                # until all 2560 records are flushed because the console can't find a record with the requested
+                # time stamp. Instead the console sends all records with the eldest record first which causes the
+                # DMPAFT routine to end. The time check itself is done in routine _archive_datetime which will
+                # return 0 when a 'hour mismatch' is detected.
+                # Check for 'hour mismatch'
+                if _record['dateTime'] == 0:
+                    # First hour of summertime time stamps
+                    syslog.syslog(syslog.LOG_ERR, "vantage: DMPAFT got record with first hour of summertime; skip record")
+                    # set the time of the console clock (both systems then will have the same DST flag again)
+                    self.setTime()
+                else:
+                    # Set the last time to the current time, and yield the packet
+                    _last_good_ts = _record['dateTime']
+                    yield _record
 
             # The starting index for pages other than the first is always zero
             _start_index = 0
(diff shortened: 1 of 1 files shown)


bin/weewx/drivers/vantage.py
```diff
@@ -1481,6 +1502,14 @@ def _archive_datetime(datestamp, timestamp):
                       0, 0, -1)                     # have OS guess DST
         # Convert to epoch time:
         ts = int(time.mktime(time_tuple))
+        # fix for hour mismatches at begin of summertime
+        # calculate a time tuple from the UTC time stamp
+        utc_time_tuple = time.localtime(ts)
```
tkeffer: this time tuple will hold the _local time_, not UTC time.

lheijst: That's is where the problem lies, Tom. If the Vantage console would report its timestamps in UTC we wouldn't had a problem. What we do is convert vantaghe time stamp to UTC and then back (to the Vantage's format). During the first hour of the summertime these times won't be the same but differ 1 hour. No solution ever to fix this either, we have to accept this. Because of this one hour difference the stored record in the weewx database (in UTC) never can be translated back to the time stamp the vantage has of this record (one our differnce , yoy know).

lheijst: Ok, I now got what you mean:
The name of the variable should be something like: 'vantage_local_time_tuple' instead of 'utc_time_tuple'.

tkeffer:
My interpretation of the situation is very different.

First, we can both agree that the logger holds the time in local time, with no indication of whether it is daylight savings time or not. This is a design flaw on the part of Davis.

Second, with all times held in local time, this is what the logger memory is supposed to look like (using 15 minute archive intervals):

```
...
0130
0145
0200
0315
0330
...
```

Unfortunately, I think what actually happens depends on which page the 0315 record is put. In the examples in the weewx-user archive, it seemed to get put in the very first page, or the second page. This completely messes up the `DMPAFT` command in a way that I don't understand.

I really wish someone had done a `--logger-summary` dump after experiencing this problem.

bin/weewx/drivers/vantage.py
tkeffer: Right.

Also, see my note to the PR. 

There are no "bad" timestamps. All records represent real data. We just don't know whether the times are in DST or not.

lheijst:
Well, I tried my best, Tom, But nobody reacted on a request of one named 'Luc' to dump these records!

In my oppinion it doesn't matter which 'page' is concerned. Weewx translates the most recent time stamp to a local time stamp that doesn't exist in the vantage logger (as we now know why). So the records can't be found and the record won't be found. Simple as that!

bin/weewx/drivers/vantage.py
lheijst: Weewx also doesn't know if the time is in DST or not. From your comments in the driver:
        # Construct a time tuple from Davis time. Unfortunately, as timestamps come
        # off the Vantage logger, there is no way of telling whether or not DST is
        # in effect. So, have the operating system guess by using a '-1' in the last
        # position of the time tuple. It's the best we can do...

lheijst: Note: This 'best guess' most of the time works. But not when the computer already switched to summer time and the Vantage console presents i'ts most recent record with a time stamp in winter time (without as we know any information if the time in question is winter time or summer time).

lheijst:
I tried it myself, Tom. Select a timestamp in the first hour of the summer time period. Translate it to UTC and translate it back to local time (all with the commands the vantage driver also uses). You will notice the outcome will differ one hour.

bin/weewx/drivers/vantage.py
lheijst: Tom, you stated: "nothing should be thrown away". I agree. The current situation is: the user notices weewx is 'storing nothing anymore to the weewx database since summert time change'. At last he/she decides to clear the logger memory. As by 'magic' weewx is working again! Hurray!
How many weewx records are lost in this process do you think? (Please don't answer!).

tkeffer:
I don't think so. It's true that

```
time.localtime(time.mktime((2016,3,13,2,0,0,0,0,0)))
```

gives 0300 DST, but that's actually correct. 0200 ST is the same as 0300 DST.

Doing the query for the first hour of DST
```
time.localtime(time.mktime((2016,3,13,3,0,0,0,0,1)))
```
gives the same answer, 0300 DST.

Daylight savings time is a stupid idea. I wish they'd drop it...

lheijst:
Tom, you don't think so, but keep in mind:
1. The problems five weewx users reported about the same subject: are they all fake?
2. Please tell me again: how do we read the summer/wintertime status from the Vantage console?

I totally agree with you about the daylight savings idea is stupid! On the other hand: it gives paid programmers hands full of work. Unfortunately non-paid programmers have to solve these problems too! Sigh!

tkeffer:
> The problems five weewx users reported about the same subject: are they all fake?

Of course not. But, there were only two users who experienced the problem (Phil and Mark, although there are probably others who did not bother to report in). All others were using something other than the Vantage console. 

I'm trying to avoid affecting the 99+% of users who did not experience a problem. A patch without a reasoned, tested, theory of what it is trying to fix is likely to lead to more problems, not less.

I've been trying to get around this flaw in the Davis consoles for nearly 10 years. I've tried lots of things and I've learned there is unlikely to be an easy fix. The only sure fire way around the problem is to ignore the logger and just rely on LOOP packets.

Frankly, spring is the easy one. It's the fall transition off of DST that is the tough one: in that case, you have _two_ records with the same timestamp. Then it becomes extremely hard to unravel which is which.

lheijst:
Tom, do as you please. No problem with me! Suriname has no winter(time): it's always summer here!

Please be so honest the next time people report this item to tell them they are only a smart part of the weewx community who report this and you didn't want to solve it on behalve of the majority and please don't play hide and seek.

Discussion closed.

Luc

I'm not pissed of, only disappointed.

Issue #128, opened 2016-06-06
rtfreq should not be fixed at 2.5


matthewwall:
rick pointed out that the rtfreq in the wunderground restful service is hard-coded to 2.5:

https://groups.google.com/forum/#!topic/weewx-user/DeTlOPVgCNc

although 2.5 is probably close enough for vantage stations, it is not correct for many others.

how do we fix this?

i'm pretty sure we don't want rtfreq to show up in any driver.  like many other wunderground issues (e.g., dealing with partial packets), the fix for this probably belongs in the wunderground restful service.

one option is to make the wunderground uploader remember the timestamp of the last rapidfire upload.  when another rapidfire packet comes in, set rtfreq to the difference between the current and last.

we'll quickly see whether wunderground actually does anything with rtfreq, and whether they can deal with non-constant values...

matthewwall:
btw, this is rick's suggestion:

> So, feature request, in order of convenience for the user:
> derive the actual rtfreq by timestamping packets
> allow a configurable rtfreq entry in the wunderground section of weewx.config
> only set the 'rtfreq' record if it is unset, allowing driver authors to set it in the driver

tkeffer:
It's probably not a good idea to leave `rtfreq` hard coded, but before we start patching code, we should know what it does.

Does it represent an aspiration on the part of the hardware?

Or, an average frequency of packets?

And, what does WU do if posts come in more frequently than `rtfreq`? Less frequently?

Let's be sure we understand what problem we are trying to solve before "fixing" this.

whorfin:
As I mentioned in the group posting [but I guess not with the checkin, sorry] rtfreq appears to only be used by the Flash-based "View full-screen RapidFire" window on the wunderground site.
It is also only used once per load of the Flash code, ie "Frame 1".
To elaborate, the RapidFire window will refresh every "the first value of rtfreq seen" seconds.
Subsequent values of rtfreq, if changed by the station, are passed along by wunderground for each
packet, but ignored by the Flash code.

Rather obviously to those of us with stations which refresh less frequenty than 2.5 seconds,
wunderground does nothing in particular if posts come in less frequently than rtfreq.
It appears to be a simple attempt to guide the Flash window refresh without bothering to compute anything server-side.

The regular non-Flash javascript dashboard appears not to use rtfreq.
The iOS mobile app appears to behave similarly to the javascript dashboard, though it gets the station update timestamp grossly wrong if you ask it.

[...]

PR #129, opened 2016-06-08
Change "rtfreq" in RapidFire mode of WUnderground publish to be automatic

whorfin:
Rather than hardcoded 2.5 sec, a moving average over packet timestamps
is computed and sent.
This addresses [#128]


tkeffer:
No activity

PR #130, opened 2016-06-09
Change the StdWXCalculate service to wrap a WXCalculate object

gjr80:
Added a layer of abstraction so that the `StdWXCalculate` service wraps the new, non-service `WXCalculate` object. This makes the (former) `StdWXCalculate` service capabilities more easily used by other code/utilities through use of the wrapped `WXCalculate` object without the need for the overheads that go with invoking a weewx service. 
This change will help facilitate a `wee_import` utility and perhaps a 'backfill-calculations' capability for `wee_database`.


tkeffer:
Makes sense.

PR #132, opened 2016-06-19
implement caching for wu

matthewwall:
this has been tested with a wmr300 station, which emits partial packets, and standard wu (not rapid-fire).  the wmr300 emits a pressure packet every 15 minutes, so only every third archive record has pressure.  this patch retains the pressure value so packets sent to wu always have pressure.  this patch has no effect on data saved to the database.

rain is still an issue.  wu uses an accumulated rain value (dayRain) instead of a delta rain value (rain).  so, caching the dayRain across day boundary might result in bogus data.

### DIFF
--- a/bin/weeutil/weeutil.py
+++ b/bin/weeutil/weeutil.py
@@ -1275,6 +1275,50 @@ class ListOfDicts(dict):
     def extend(self, new_dict):
         self.dict_list.append(new_dict)
 
+
+class ObservationCache(object):
+    """Associate timestamps with observations.  Designed to be used with
+    packet/record dictionaries that include a 'dateTime' field that applies
+    to all values in the packet/record.
+
+    To create and populate a cache:
+    cache = ObservationCache()
+    cache.add_record(packet)
+
+    Update the cache values:
+    cache.add_record(packet)
+
+    To get the most recent cached observations:
+    record = cache.get_most_recent()
+
+    Optionally specify a staleness dictionary with per-observation ages:
+    packet = cache.get_most_recent(stale_dict={'UV': 10, 'pressure': 120})
+
+    Optionally specify a list of observations to exclude:
+    packet = cache.get_most_recent(excludes=['rain', 'hail'])
+    """
+
+    def __init__(self):
+        self.values = dict()
+        self.timestamps = dict()
+
+    def add_record(self, record):
+        for obs in record:
+            if record[obs] is not None:
+                self.values[obs] = record[obs]
+                self.timestamps[obs] = record['dateTime']
+
+    def get_most_recent(self, stale_dict=dict(), excludes=[]):
+        record = dict()
+        if 'dateTime' in self.values:
+            ref_time = self.values['dateTime']
+            for obs in list(set(self.values.keys()) - set(excludes)):
+                if (obs not in stale_dict or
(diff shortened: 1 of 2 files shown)


tkeffer:
Are you looking for me to approve a pull request from you?

matthewwall:
the patch solves the problem for wmr300, should solve it for any other station with partial packets, and should not affect stations with full packets.  so i'm just looking for critical feedback, and want to be sure i have not missed anything.

gjr80:
One thing that sprung to mind last night was what would be the impact of differing behaviours between weewx WU posts using cached obs and wunderfixer (which would effectively being using uncached obs). 

Considering non-RF use; I guess the issue will only arise when the original weewx post fails (or the subsequent wunderfixer get from WU fails to produce the record). If barometer is the field of concern, Wunderfixer in its current state would not post barometer and you would end up with the pre-cache situation of having a some random WU behaviour for barometer in the record concerned. Wunderfixer uses `Manager.getRecord()` to query the archive so I guess by making use of the `max_delta` parameter in `getRecord()` you could effectively institute an 'archive' cache though it would effectively be a backward and forward cache given how `max_delta` is implemented.

I know little of exactly how WU uses RF posts to have any idea of the impact if using RF.

tkeffer:
OK. Here's what immediately pops into my mind:

I guess I'm not following the logic for `ObservationCache`. Why is `ref_time` set to the last time a record was inserted, rather than the last time a specific observation value was refreshed? Is the assumption that a use pattern will be followed where a packet will be inserted, immediately following which you'll ask the cache for the most recent value? Seems like it would be safer to pass `ref_time` into `get_most_recent()`.

For the `restx.py` code, why cache archive records? Haven't the accumulators already done that? Or, are you trying to cover the case where a user sets such a short archive interval, that the partial packets may not have populate all the observation types in an archive record? If so, this is getting close to making up data. Many users treat the WU PWS "archive" data as a kind of data-of-record and, indeed, Gary is writing a utility `wee_import` that would formalize this. 

Or, another way of looking at it, is what problem in the WU PWS protocol are we trying to fix?

Personally, I think this patch should stick to making the broken Rapidfire work and not mess with the PWS protocol.

tkeffer:
Matthew?

matthewwall:
here are the related use cases:

1) wmr300 has partial packets and at least one reading (pressure) is not reported within each archive interval.  pressure is reported once every 15 minutes.  as a result, anyone posting to wu (rapidfire or otherwise) will see sporadic pressure readings.  this is not unique to wu.

2) acurite, oregon scientific, and other stations send partial packets.  when posting to wu using non-rapidfire, this is not a problem, since each archive record (usually) contains a complete packet.  wu rapidfire, on the other hand, does not handle partial packets properly.

this pull request was written to address (1), but i think the implementation is too bespoke and needs to be generalized, since (1) does not apply only to wu.

part of the issue is a user education problem.  some users want to fabricate data in order to see smooth plots; for some people, pretty plots and tables with no empty spaces matter more than accurate data.

tkeffer:
OK.

As far as your last point, let us say that weewx comes down firmly on the side of not fabricating data. If an instrument emits erratically, then you can expect plots with gaps.

matthewwall:
caching for rapidfire only was implemented at commit 4ed8c78069a60fbb4968679beb3766ac57f5fde1

PR #138, opened 2016-07-29
Fix typo

gjr80:
You don't want to know how long it took to find this one!


tkeffer:
I had to stare at it for a few seconds to see the change!

wotid:
I feel your pain.

Pirates would be lost without Arrr!

Issue #97, opened 2016-02-03
consider making a 'wee_import' utility to import archive records from an external source


gjr80:
A number of users have sought advice on how to import Weather Underground PWS data, usually as a last resort disaster recovery measure. Currently this is achieved manually by obtaining the WU data in CSV format then massaging it into a format where it can be imported using an SQL query or some other method.

This process should be able to be automated with a 'wee_xxxx' like utility. Further more the utility should be generic in nature so that additional sources (eg WOW, CWOP or some non-web based source) can be used.

gjr80:
M Wall [comments](https://groups.google.com/d/msg/weewx-development/SUOZmDkw-ww/taWcK-xwAwAJ) of 2 Feb (summary):
- use data from an arbitrary source, not just wunderground
- define a mapping between source schema and weewx schema (which may not be the default wview schema)
- schema mapping could probably be done in a config file, so the code to support multiple sources would be minimal
- specify what to do about derived quantities (e.g., apply StdWXCalculate to each record)

gjr80:
A few questions/points of clarrification:

1. **Field 'interval'**. 'interval' is an INTEGER NOT NULL field in the weewx schema. I believe weewx accepts 'interval' values that can change from record to record but currently weewx does not take this into account when calculating averages etc. Obs sourced from WU (and I suspect the same for other sources) may have irregular intervals between successive records, depending on the capabilities of the PWS and the configuration of weewx, or simply because there are missing records or the vagaries of WU. What is the preferred approach for handling 'interval'? A few possibilities spring to mind:
a. calculate the interval as the difference between the timestamps of successive records.
b. use the interval from weewx.conf or PWS
c. sample a day of observation and use the mode 'interval' (ie the most common 'interval')
a. would seem the most logical but could have some wild consequences if there are numerous missing records (could limit the max 'interval' I guess). b would seem to introduce some inconsistencies in the resulting archive as would c. Also, 'interval' is defined in whole minutes so any resulting interval would need to be rounded. Looking for some guidance here?

2. **WU WXDailyHistory.asp units**. Does WXDailyHistory.asp return metric units or am I seeing some biased results because I am in Australia? Would appreciate it is someone in USA would confim what units they see in WXDailyHistory.asp query.

3. **UV**. I do not see my UV results appearing in a WXDailyHistory.asp query. Am I special or does this the same for others as well? I have queried other stations I know have UV sensors and see the same resu

tkeffer:
[...]

PR #140, opened 2016-08-19
A wee_import utility

gjr80:
In response to #97.

I think I have taken on board feedback to date at #97. Started with a CSV import that imports from a single CSV file with headers, extended that to a WU import that imports multiple days of WU data. In response to a user request (nothing heard since) have added a Cumulus import that imports from multiple Cumulus monthly log files.

I do not claim to have tested every conceivable combination/permutation but I think I have given the utility some fairly thorough testing.

It is complex and the command line help instructions/options are lengthy as are the comments in `wee_import.conf`. I have not started to document the use of `wee_import` other than in `wee_import.conf` comments and the pre-amble comments in `wee_import`. I think we are talking about more than a few paragraphs, is this best added to the already burgeoning _Customisation Guide_ or perhaps a wiki entry? Happy to do either.


tkeffer:
I'm a little concerned about introducing a second config file. Users might get confused over which is which. 

Where would it live in a DEB install? In `/etc/weewx`? 

This config file plays a different role than `weewx.conf` or, for that matter, all the other configuration files in `/etc`. It's more of a convenient way of collecting all the options together that might otherwise be command line parameters to `wee_import`.

But, I don't have any better ideas what to do!

Your thoughts?

The `except...as` construct was introduced in Python 2.6. We are trying to target 2.5.

gjr80:
>I'm a little concerned about introducing a second config file. Users might get confused over which is which. 

Understand the concern about another config file. I think the 'inexperienced user' situation is mitigated as follows:

- if a user edits `wee_import.conf` at least they will not doing anything destructive to their existing weewx install
- the word 'weewx' is not used in the file name
- the support headache of a user actually editing `wee_import.conf` when he said/was told to edit `weewx.conf` could be mitigated by greater use of `wee_debug` for when asking the user for details of his config
- could put some words at the top of `wee_import.conf` or the docs but we know most don't read them.

>Where would it live in a DEB install? In `/etc/weewx`? 

I guess that was my first thought. Creating another folder seems overkill, putting it in the same folder as `wee_import` is an option, but that seems to break the mold of keeping config separate to executables.

>This config file plays a different role than `weewx.conf` or, for that matter, all the other configuration files in `/etc`. It's more of a convenient way of collecting all the options together that might otherwise be command line parameters to `wee_import`.

I guess, but in the extreme could you not say the same of `weewx.conf`? :)

>But, I don't have any better ideas what to do!

>Your thoughts?

I guess this is the money question. Some quick brainstorming:

- put it on the command line - some judicious setting of defaults makes this more pallatable but not really practical for mapping fields
- put it in `weewx.conf` - would work but I don't think you really want to pollute `weewx.conf`
- put it in `wee_import` - I don't think we really want users editing `wee_import` and possibly changing the code
- call it something else - could drop the '.conf' I guess, I don't know enough linux/programming to know how big a line this is to cross!
- put it in the source data file - again some judicious setting of defaults makes this more pallatable and would probably reduce it down to just the field maps, which at the moment only really apply to CSV. I guess though the point of `wee_import` was to try to get away from making the user do a whole lot of editing of source data files so I am not too keen on this one. One thing I have discovered is that some systems use a monthly file, so having the user edit many files to import say a few years of data does not really make for a useful tool.

That was about all I could think of, maybe Matthew has some good ideas? I don't think any are really viable solutions. To me it really comes down to the mapping, though mapping is only really an issue for CSV imports. WU and Cumulus 'sources' are well defined and the mapping is implicit in the data. I did do some playing around with VWS and depending on what you use as a source(there monthly tabulated files or the user can create CSV export) it can be ill-defined (or perhaps 'variable' is a better choice of words). I would like in the fullness of time to extend `wee_import` to WD but have not even looked at that yet.

>The `except...as` construct was introduced in Python 2.6. We are trying to target 2.5.

Noted, to sort of quote somebody around here 'a little refactoring of the code should fix it'. Note to self - get a 2.5 install for future testing.

tkeffer:
That's a great analysis and I have to agree with your conclusions. 

If we end up having a whole raft of config files, we'll deal with it later.

> > The except...as construct was introduced in Python 2.6. We are trying to target 2.5.

>Noted, to sort of quote somebody around here 'a little refactoring of the code should fix it'. Note to self - get a 2.5 install for future testing.

It's simpler than that: I use [PyDev](http://www.pydev.org/) in Eclipse which allows you to specify which version of Python you're targeting. It then flags anything that falls short.

Put the V2.5 patch in then check in your additions.

gjr80:
Any thoughts on documentation(when I write it), find somewhere to put it in the Customization Guide?

tkeffer:
Given the pattern so far, that would be the logical place for it.

But, I'm wondering if it's time to refactor the documentation to include a separate "utilities" document. Or, at least, a separate appendix in the Customizing Guide. There are now a lot of utilities, and the present organization requires the reader to scan a lot of documentation just to find them all.

gjr80:
Ok, I'll write it in the style of the Customization Guide and it can be moved around when/if required.

tkeffer:
Thanks, Gary. Once you get it done and checked in, I'll take on the refactor. It shouldn't be a big job. Just a bunch of mindless shuffling around HTML.

PR #146, opened 2016-09-02
Allow calcRainRate to calculate rainRate from archive records if no l…

gjr80:
`WXCalculate.calcRainRate()` only calculates `rainRate` if loop data is available. This is done using retained loop rain data and is based on a sliding window of user definable length. This works for a connected weather station (or any other supported source) that provides loop data but does not work with `wee_import` where there is no loop data. In this case `rainRate` is always 0.

This change retains archive rain data in parallel with loop rain data and if there is no loop rain data then `rainRate` is calculated using the same sliding window algorithm, but based on a sum of archive rain not loop rain. All (currently supported) weather stations provide loop data so this change does not affect operation with a connected weather station, in these cases `rainRate` is calculated exactly as before. If a connected weather station does not provide loop data, or should `WXCalculate` be used externally without loop data (eg via `wee_import`), then `rainRate` will be calculated using the same user defined sliding window but based on the sum of archive rain instead.

This change will allow `wee_import` to provide `rainRate` data from sources that provide rain but not `rainRate` data.

### DIFF
--- a/bin/weewx/wxservices.py
+++ b/bin/weewx/wxservices.py
@@ -159,6 +159,7 @@ class WXCalculate(object):
         self.temperature_12h_ago = None
         self.ts_12h_ago = None
         self.rain_events = []
+        self.archive_rain_events = []
 
         # report about which values will be calculated...
         syslog.syslog(syslog.LOG_INFO, "wxcalculate: The following values will be calculated: %s" %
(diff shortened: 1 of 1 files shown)


gjr80:
>Looking through the original `calc_rainRate`, It seems odd to me that knowledge of the source of the rain events would appear in the algorithm. Shouldn't it just be a pure algorithm that calculates rain rate, irregardless of the origin of the data?

I understand what you are saying but I cannot see a way around it. Effectively we have 2 cases; case 1 is where we have loop packets (ie a weather station) and case 2 is where we don't have loop packets but rather just archive records (wee_import). Unfortunately, we have archive records arriving as well in case 1 meaning there needs to be a way of distinuishing between the two. My brain is not very switched on today, I will give it some more thought though.

tkeffer:
I'm probably just over-analyzing it.

One could factor out the algorithm, then use it for both the loop and archive data. That's probably what I'm thinking. 

Don't worry about it.

gjr80:
I could move the smell from wxservices to wee_import by putting things back as they were and subclassing WXCalculate and then overriding calcRainRate()....

PR #148, opened 2016-09-09
Restructured wee_import

gjr80:
This PR contains a restructured PR given @matthewwall comments circa 29 August at issue #97.

Points of note:
- A `bin/weeimport` directory has been created.
- The core `wee_import` code has been moved to `weeimport.py` in the `bin/weeimport` directory.
- Child classes supporting different sources have been moved to individual `xxximport.py` files in `bin/weeimport`.
- `util/import` directory has been created.
- Example import config files for each supported source now reside in `util/import`.
- `wee_import` has been reduced to setting up optparse and then kicking off the import.
- wee_import command line options have been reduced significantly, `--import-config` is mandatory, `--date` and `--log` have been retained.
- `wee_import` log was intially moved to a separate log file using the python logging module but this has been reverted to the weewx log using syslog. The only `--log` option is `--log=-` to turn off logging to file.
- Added quality checks to imported data using the `weewx.conf`` StdQC` min/max settings via `ImportQC` class in `weeimport.py`. I don't believe I could use the `StdQC` service code directly given its binding to `NEW_LOOP_PACKET` and `NEW_ARCHIVE_RECORD` so there is some duplication though it is only a handful of lines and I would not have thought it likely to change much into the future. This is a big help for WU imports.
- Revised Customization Guide given `wee_import` changes.

Usual caveat applies - I expect any option/parameter names I created to be changed!

Not sure what is easiest for you to look at `wee_import` now; this PR or the branch in my clone. This PR came from my `wee_import_QC` branch at commit https://github.com/weewx/weewx/commit/1fc28903251ddc8d018db9923462e08b2c65e4c2.


tkeffer:
I'm liking this organization better.

Question: why does a sample `wee_import` configuration file contain stanzas for something other than the import its designed for? That is, why does `csv-example.conf` include a `[Wunder]` section?

matthewwall:
a few things:

- you should call QC code (analogous to WXCalculate) instead of duplicating it.  at some point QC will be more than just min/max filtering.

- the logging stuff is smelly (not your fault).  we need to be consistent throughout weewx about what 'verbose' and 'debug' and log levels mean.  its ok to have a custom logging class in there right now (at least until we figure out syslog vs python logging throughout weewx), but we need to get the semantics of using it right.  for example, we don't want to end up with one utility that has --debug and other that has --verbose.  in weewx.conf, debug=1 and debug=2 increase verbosity.  how does that fit with a simple 'verbose' flag?  to sort this out we need to outline the known use cases (across all weewx utilities) then sketch out the logging/feedback solution that makes them all work consistently.

- i would rename wunderimport.py to wuimport.py.  also wu-example.conf.  basically use WU as you use CSV.  otherwise you'll have case-sensitivity issues all over the place with wunder.

- each example.conf should contain only the top level stanza for that example.  wu-example.conf would have only [WU]

- there is no need to shout in the log messages.

- prefer to reduce code in logging conditionals, and keep the log/feedback messages succinct.  makes it easier to fine tune the logging for translations (not a big deal here), general use (definitely a big deal), and log analysis (e.g., logwatch).  

for example, instead of this:

```
if self.calc_missing:
    print "Any missing derived observations WILL be calculated."
else:
    print "Any missing derived observations WILL NOT be calculated."
```

i would do this:

```
if self.calc_missing:
    print "Missing derived observations will be calculated."
else:
    print "Missing derived observations will not be calculated."
```

only log/feedback for exceptional behavior.  for example, instead of this:

```
if self.dry_run:
    print "This is a dry run, imported data WILL NOT be saved to archive."
else:
    print "This is NOT a dry run, imported data WILL be saved to archive."
```

i would do this:

```
if self.dry_run:
    print "This is a dry run; imported data will not be saved."
```

gjr80:
Sorry, long comment.

>tk:
>Question: why does a sample wee_import configuration file contain stanzas for something other than the import its designed for? That is, why does csv-example.conf include a [Wunder] section?

>mw:
>each example.conf should contain only the top level stanza for that example. wu-example.conf would have only [WU]

Ok, perhaps I have misunderstood the conversation at #97 on 29 Aug:

>mw:
>but that still leaves us with at least two options for structuring the config file:

>option 1: one, and only one, top-level stanza in the conf. the name of the top-level stanza determines the source, and also determine which code will be loaded to parse the stanza and do the import.

>option 2: any number of top-level stanzas in the conf, with a root-level 'source=(CSV|WU|Cumulus) that determines which stanza will actually be used and which code will do the parsing and import. only one source can be 'active' per run.

>all options would go within the top-level stanza, even if the options are common to multiple sources. at some time in the future we might want to permit common options to be inherited and overridden (that is the pattern in other weewx configs), so the structure and behavior we define now should not preclude that.

>for simplicity we could go with option 1, then extend that later to option 2 if we find it easier to test with a single config and/or if a use case for single config with changing source arises.

>the use cases i am designing for are (1) one-off import, e.g., moving from weatherdisplay or cumulus or heavyweather to weewx, and (2) automated, repeated import, e.g., pull data from vws or wu every day/week/month.

>gr:
>So happy with option2?

>mw:
>i think so. 

I guess you are wanting to stick with the use case mentioned. I will remove the redundant sections in the example config files.

>you should call QC code (analogous to WXCalculate) instead of duplicating it. at some point QC will be more than just min/max filtering.

Just looked at `StdQC` again and it seems obvious now (red face), my only defence is that for some reason the binding to the events threw me.

>the logging stuff is smelly (not your fault). we need to be consistent throughout weewx about what 'verbose' and 'debug' and log levels mean. its ok to have a custom logging class in there right now (at least until we figure out syslog vs python logging throughout weewx), but we need to get the semantics of using it right. for example, we don't want to end up with one utility that has --debug and other that has --verbose. in weewx.conf, debug=1 and debug=2 increase verbosity. how does that fit with a simple 'verbose' flag? to sort this out we need to outline the known use cases (across all weewx utilities) then sketch out the logging/feedback solution that makes them all work consistently.

This sounds like a bit of work. Given the different approaches across weewx, the utilities and the drivers do I/we (1) stop and sort the bigger issue out, (2) leave it is until we come back to this or (3) adopt a different approach (in `wee_import`)?

>i would rename wunderimport.py to wuimport.py. also wu-example.conf. basically use WU as you use CSV. otherwise you'll have case-sensitivity issues all over the place with wunder.

OK, done.

>there is no need to shout in the log messages.

Ok, but I know a lot of people don't listen very well :)

>prefer to reduce code in logging conditionals, and keep the log/feedback messages succinct. makes it easier to fine tune the logging for translations (not a big deal here), general use (definitely a big deal), and log analysis (e.g., logwatch). 

Ok, verbosity has always been an issue for me, you would think 28 years of military service would have rid me of that! I would say the thought I had in my mind was always that the 'output' was serving two purposes; (1) to populate the log to allow suitable forensics to occur and (2) to inform the user what was/was not happening. I have always believed that (2), whilst needing to be succinct, does need a few more words than (1). Nonetheless, I will critically review each piece of 'output'.

Some more work to do, one question for you (either). I see this comment at the bottom of the PR:

>Add more commits by pushing to the wee_import_QC branch on gjr80/weewx.

So this PR remains open and I just continue to commit to my branch and then advise when i think it needs further review ?

gjr80:
Re `StdQC`, following on from the way we approached `StdWXCalaculate` is this appropriate? If so, didn't know if it was appropriate to put class `QC` where it is or tuck it away somewhere else so as to not interfere with the layout of `engine.py`. Also, I put a parameter int here `parent` to keep the 'engine' word in the log output but also to give another app the freedom to change it to something appropriate. Is there a better way?

```python
#==============================================================================
#                    Class StdQC
#==============================================================================

class StdQC(StdService):
    """Service that performs quality check on incoming data."""

    def __init__(self, engine, config_dict):
        super(StdQC, self).__init__(engine, config_dict)

        # Get a QC object to actually apply the QC checks to our data
        self.qc = QC(config_dict)
        
        self.bind(weewx.NEW_LOOP_PACKET, self.new_loop_packet)
        self.bind(weewx.NEW_ARCHIVE_RECORD, self.new_archive_record)
        
    def new_loop_packet(self, event):
        """Apply quality check to the data in a loop packet"""
        
        self.qc.apply_qc(event.packet)

    def new_archive_record(self, event):
        """Apply quality check to the data in an archive record"""
        
        self.qc.apply_qc(event.record)

#==============================================================================
#                    Class QC
#==============================================================================

class QC(object):
    """Class to apply quality checks to a record."""
    
    def __init__(self, config_dict, parent='engine'):
        
        # Save our 'parent' - for use when logging
        self.parent = parent
        
        # If the 'StdQC' or 'MinMax' sections do not exist in the configuration
        # dictionary, then an exception will get thrown and nothing will be
        # done.
        try:
            mm_dict = config_dict['StdQC']['MinMax']
        except KeyError:
            syslog.syslog(syslog.LOG_NOTICE, "engine: No QC information in config file.")
            return

        self.min_max_dict = {}

        target_unit_name = config_dict['StdConvert']['target_unit']
        target_unit = weewx.units.unit_constants[target_unit_name.upper()]
        converter = weewx.units.StdUnitConverters[target_unit]

        for obs_type in mm_dict.scalars:
            minval = float(mm_dict[obs_type][0])
            maxval = float(mm_dict[obs_type][1])
            if len(mm_dict[obs_type]) == 3:
                group = weewx.units._getUnitGroup(obs_type)
                vt = (minval, mm_dict[obs_type][2], group)
                minval = converter.convert(vt)[0]
                vt = (maxval, mm_dict[obs_type][2], group)
                maxval = converter.convert(vt)[0]
            self.min_max_dict[obs_type] = (minval, maxval)
        
    def apply_qc(self, data_dict):
        """Apply quality checks to the data in a record"""

        for obs_type in self.min_max_dict:
            if data_dict.has_key(obs_type) and data_dict[obs_type] is not None:
                if not self.min_max_dict[obs_type][0] <= data_dict[obs_type] <= self.min_max_dict[obs_type][1]:
                    syslog.syslog(syslog.LOG_NOTICE, self.parent + ": %s LOOP value '%s' %s outside limits (%s, %s)" % 
                                  (weeutil.weeutil.timestamp_to_string(data_dict['dateTime']), 
                                   obs_type, data_dict[obs_type], 
                                   self.min_max_dict[obs_type][0], self.min_max_dict[obs_type][1]))
                    data_dict[obs_type] = None
```

matthewwall:
gary, regarding the conf files, my intent was this:

wu-example.conf
```
source = WU
[WU]
    station = xxx
    ...
```

csv-example.conf
```
source = CSV
[CSV]
    file = xxx
    ...
```

so if someone wanted to, they could do this:

mwall-imports.conf
```
source = WU
[CSV]
    file = xxx
    ...
[WU]
    station = xxx
    ...
[VWS]
    ...
```

and eventually, specify the 'source' option at the command line to indicate which section will apply.  parameters that apply to all sources could be specified at the root level (in no stanza), and/or in each source stanza.

matthewwall:
regarding logging: i'd rather get 3.6 out the door (the te923 stuff is killing me!), so now is probably not a good time to do a weewx-wide audit/refactoring of logging.  but we should be careful to not introduce any user-facing options or behavior that will result in backward compatibility that we don't want to maintain.

gjr80:
Ok x 2. Regarding 3.6, I would like to have any `wee_import` pre-requisites in 3.6 if possible, not fussed when `wee_import` is included. At the moment the only 2 pre-reqs are the new `StdWXCalculate` structure and any changes required for `StdQC`. `StdWXCalculate` has already been committed, if my `StdQC` approach above is OK I will get a PR for it submitted.

tkeffer:
I agree in V3.6. It is long overdue. Don't worry about logging. Just log to the syslog.

But, to answer the meta-question: to me, the primary job of logging is to allow remote debugging and support.

PR #149, opened 2016-09-11
Add a layer of abstraction to StdQC

gjr80:
Taking a leaf out of the 'Ask for forgiveness, rather than permission' mantra...

Following on from @matthewwall #148 comments of 9 Sep, this PR adds a layer of abstraction to the `StdQC` service to allow other apps to hook into the underlying weewx QC system.

### DIFF
--- a/bin/weewx/engine.py
+++ b/bin/weewx/engine.py
@@ -25,6 +25,7 @@ import daemon
 import weedb
 import weewx.accum
 import weewx.manager
+import weewx.qc
 import weewx.station
 import weewx.reportengine
 import weeutil.weeutil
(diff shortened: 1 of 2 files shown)


tkeffer:
Can we move class `QC` into its own file so other things that want to use it don't have to import `engine.py`?

gjr80:
Did mention that in my #148 comment but somehow missed that in this PR. You know my naming choices are poor, I don't see any other obvious candidates files/directories. I might need some direction here.

tkeffer:
Ummm, ... `qc.py`? In the `bin/weewx` directory.

gjr80:
Hmm, that was simple. I'll fix it now.

tkeffer:
Thanks, Gary!

PR #154, opened 2016-09-16
Implement format_data() in RESTThread

dcapslock:
Implements format_data() in RESTThread to allow a subclass of RESTThread to use REST APIs which require posting of data while still using the rest of the base class functions.

### DIFF
--- a/bin/weewx/restx.py
+++ b/bin/weewx/restx.py
@@ -364,11 +364,12 @@ class RESTThread(threading.Thread):
         _us_record = weewx.units.to_US(_full_record)
         # ... format the URL, using the relevant protocol ...
         _url = self.format_url(_us_record)
+        _data = self.format_data(_us_record)
         # ... convert to a Request object ...
         _request = urllib2.Request(_url)
         _request.add_header("User-Agent", "weewx/%s" % weewx.__version__)
         # ... then, finally, post it
-        self.post_with_retries(_request)
+        self.post_with_retries(_request, _data)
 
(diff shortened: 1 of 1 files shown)


dcapslock:
While that works it means that they specialising class needs to unnecessarily deal with creating the request etc. Using the logic that default process_record can deal with get and post cases, without handling process_record in the specialising class makes it cleaner. Otherwise, why not do all this work in get_record?

tkeffer:
That makes sense. Thank you!

tkeffer:
Resurrecting this PR. 

Are you actually using this capability? The present implementation only allows urlencoded POSTs. I would like to extend it to allow JSON POSTs.

dcapslock:
Yes. The OpsGenie extension is using this. I can update that as required.

________________________________

tkeffer:
It would not be hard to provide backwards compatibility, but if you're the only one using it (true, as far as I can tell), I'd rather keep it simple.

tkeffer:
So long as they have their own version of process_record, it should not be a problem.

tkeffer:
OK, try this one on for size: https://github.com/weewx/weewx/compare/1f99010b2f493ba25fb9092aaf7510d0b7760f43...development

To provide a POST body, you override function `get_post_body()`. It should return a 2-way tuple (body, mime_type).

The default version calls `format_data()` and packages any result as a MIME-type of `application/x-www-form-urlencoded`, thus maintaining backwards compatibility

PR #158, opened 2016-09-23
Documented wunderfixer

gjr80:
>may i suggest that we follow the pattern used in wee_import for all of the other wee_* sections in utilities.htm?

>specifically, that means an intro with the --help output, then one section called "command-line options" with a brief description and possibly syntax/table for each option, then zero or more narrative sections that illustrate how to use the utility to perform a specific objective.

Following on from Matthew's comments above...


tkeffer:
You are a documentation animal!

gjr80:
That one was easy after wee_import!

gjr80:
Meant to say that once this PR is accepted/3.6 released the Wunderfixer link at the bottom of the weewx home page should be removed or updated since the linked readme is at odds with the new Utilities Guide (plus there is a dead link there). I will leave this up you guys - getting a bit beyond my comfort zone.

PR #161, opened 2016-09-27
Add Previmeteo to restful services

previmeteo:
Hi,

Would it be possible to add our service to the standard restful services ?

Thanks,
Xavier


tkeffer:
Thanks for your PR.

However, this addition is better suited as an extension, rather than part of the core weewx distribution, at least until Previmeteo becomes more popular.

Weewx was designed for these kinds of extensions. They are easy to create and easy to install. Please give it a try, and let's see how it goes.

PR #162, opened 2016-09-28
Add Previmeteo restful service to extensions

previmeteo:
Hi,

As suggested in your answer to our previous pull request, we have created an extension. It would be great to have it into the extensions dir of the official repo. 

Thanks !


tkeffer:
Thanks for your contribution.

Extensions are not generally packaged with the general distribution. While there are a few exceptions (`fileparse`, `pmon`, `xstats`), these are very general.

Host your extension somewhere (GitHub is popular), then include a link to it in the [uploaders](https://github.com/weewx/weewx/wiki#uploaders) section of the Wiki.

PR #168, opened 2016-10-11
Almost forgot the docs, minor reformat of wee_debug description

gjr80:
Updated wee_debug --help output.
Added --config under Actions and options. I know there has been a lot of standardisation of the documentation for the utilities but of those that use a --config option some document its use and other do not. Thought it was justified, if we don't want it let me know and I will remove remove it from Actions and options.
Minro reformat of wee_debug --help Description text.

### DIFF
--- a/bin/wee_debug
+++ b/bin/wee_debug
@@ -58,11 +58,11 @@ usage = """wee_debug --help
 
 Description:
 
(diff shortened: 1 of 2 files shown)


tkeffer:
gary, don't feel like you need to submit these minor changes as PRs!

PR #188, opened 2016-12-13
Add files via upload

richterb1:
Detect OS uptime for FreeBSD11

### DIFF
--- a/bin/weewx/station.py
+++ b/bin/weewx/station.py
@@ -3,11 +3,13 @@
 #
 #    See the file LICENSE.txt for your full rights.
 #
+#$Id: station.py 1117 2016-12-13 01:30:53Z richterb $
 """Defines (mostly static) information about a station."""
-import time
 
+import time
 import weeutil.weeutil
 import weewx.units
+import os, sys, datetime
 
 # For MacOS:
 try:
@@ -103,19 +105,46 @@ class Station(object):
         # Get the OS uptime. Because this is highly operating system dependent, several
         # different strategies may have to be tried:
         os_uptime_secs = None
-        try:
-            # For Linux:
-            os_uptime_secs = float(open("/proc/uptime").read().split()[0])
-        except (IOError, KeyError):
-            try:
-                # For MacOs:
-                os_uptime_secs = CACurrentMediaTime()
-            except NameError:
-                pass
 
-        return weewx.units.ValueHelper(value_t=(os_uptime_secs, "second", "group_deltatime"),
-                                       formatter=self.formatter,
-                                       converter=self.converter)
+	try:
+    	    import ctypes
+    	    class timespec(ctypes.Structure):
+     			_fields_ = [
+        		('tv_sec', ctypes.c_long), ('tv_nsec', ctypes.c_long)
+     			]
+
+    	    if sys.platform.startswith('freebsd'):
+                CLOCK_MONOTONIC = 4# see < time.h >
+                dylib = 'libc.so.7'
+    	    #else :
+    	    #    sys.platform.startswith('linux')
+    	    #    CLOCK_MONOTONIC = 1# see < linux / time.h >
+    	    #    dylib = 'librt.so.1'
+
+    	    syslib = ctypes.CDLL(dylib, use_errno = True)
+    	    clock_gettime = syslib.clock_gettime
+    	    clock_gettime.argtypes = [ctypes.c_int, ctypes.POINTER(timespec)]
+    	    t = timespec()
+
+    	    if clock_gettime(CLOCK_MONOTONIC, ctypes.pointer(t)) != 0:
+                errno_ = ctypes.get_errno()
+         	raise OSError(errno_, os.strerror(errno_))
+            os_uptime_secs = t.tv_sec + t.tv_nsec * 1e-9
+    
+	except ImportError:
+    	    try:
+            # For Linux:
+                os_uptime_secs = float(open("/proc/uptime").read().split()[0])
+            except (IOError, KeyError):
+                try:
+                # For MacOs:
+                 os_uptime_secs = CACurrentMediaTime()
+                except NameError:
+                 pass
+
+        return weewx.units.ValueHelper(value_t=(os_uptime_secs, "second", "group_deltatime"), 
+                                        formatter=self.formatter,
+                                        converter=self.converter)
 
     def __getattr__(self, name):
         # This is to get around bugs in the Python version of Cheetah's namemapper:
--- a/bin/weewx/station.py
+++ b/bin/weewx/station.py
@@ -3,13 +3,12 @@
 #
 #    See the file LICENSE.txt for your full rights.
 #
-#$Id: station.py 1117 2016-12-13 01:30:53Z richterb $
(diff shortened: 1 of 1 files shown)


richterb1:
Update os_uptime for FreeBSD, a simpler version

tkeffer:
I appreciate the effort to provide an uptime port for FreeBSD.

However, this version depends on running an external process and collecting the results. This command will run in the main weewx thread, potentially delaying the thread. 

Your original strategy of loading an OS-dependent shared library and then making a system call is a better one. I would suggest modifying it slightly to do the load when station.py is imported. Thereafter, you just make the call to `gettime`.

Keep going!

richterb1:
Yes, I pondered that, simplicity has it's downside.  I also don't want
to have to have additional dependency modules.  Let me work further on
the first option.  I was distracted by loading the newskin Standard
template. 

On 2016-12-13 06:11 AM, Tom Keffer wrote:

> I appreciate the effort to provide an uptime port for FreeBSD. 
> 
> However, this version depends on running an external process and collecting the results. This command will run in the main weewx thread, potentially delaying the thread. 
> 
> Your original strategy of loading an OS-dependent shared library and then making a system call is a better one. I would suggest modifying it slightly to do the load when station.py is imported. Thereafter, you just make the call to gettime. 
> 
> Keep going! 
> 
> --
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].
 

Links:
------
[1] https://github.com/weewx/weewx/pull/188#issuecomment-266747919
[2]

richterb1:
I think I updated the request, this time I'm using libc.sysctlbyname and find_library so no hardcoded paths.

tkeffer:
This is getting close, but there are two minor issues

1. Please remove the `$Id` stuff. They are not used by git.
2. There are a few hard coded tabs in there. These should be replaced by four spaces.

richterb1:
1. Duh, I forgot to remove that. 

2. Found two hard coded tabs, removed. 

3. uploaded again. 

On 2016-12-14 02:23 PM, Tom Keffer wrote:

> This is getting close, but there are two minor issues 
> 
> * Please remove the $Id stuff. They are not used by git.
> * There are a few hard coded tabs in there. These should be replaced by four spaces.
> 
> --
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].
 

Links:
------
[1] https://github.com/weewx/weewx/pull/188#issuecomment-267175852
[2]

tkeffer:
Those pesky little details.

It now runs on my Mac, but unfortunately gives the wrong answer. 

The code block intended for freeBSD is standard Python code, so it is run instead of the Mac code. Unfortunately, it returns the wrong answer. In particular, if I include a print statement to give the values:

~~~python
                os_uptime_secs = time.time() - float(buf.value)
                print "times=", time.time(), float(buf.value), os_uptime_secs 
~~~

it returns

    times= 1481757273.49 0.0 1481757273.49

So, `buf.value` is always zero.

This can be fixed by reordering the try clauses to test for MacOS first. I've taken the liberty of modifying your code to do this (attached).  This works on the Mac, but I can't test freeBSD. Perhaps you could give it a try?


[station.zip](https://github.com/weewx/weewx/files/653373/station.zip)

richterb1:
Yes, that works for me.   I ponder though if it's worth the bother to
add the flag in weewx.conf for an OS option. But I'm guessing the
overhead is not the great using this method.  

On 2016-12-14 03:24 PM, Tom Keffer wrote:

> Those pesky little details. 
> 
> It now runs on my Mac, but unfortunately gives the wrong answer. 
> 
> The code block intended for freeBSD is standard Python code, so it is run instead of the Mac code. Unfortunately, it returns the wrong answer. In particular, if I include a print statement to give the values:
> 
> os_uptime_secs = time.time() - float(buf.value)
> print "times=", time.time(), float(buf.value), os_uptime_secs 
> 
> it returns
> 
> times= 1481757273.49 0.0 1481757273.49
> 
> So, buf.value is always zero. 
> 
> This can be fixed by reordering the try clauses to test for MacOS first. I've taken the liberty of modifying your code to do this (attached). This works on the Mac, but I can't test freeBSD. Perhaps you could give it a try? 
> 
> station.zip [1] 
> 
> --
> Reply to this email directly, view it on GitHub [2], or mute the thread [3].
 

Links:
------
[1] https://github.com/weewx/weewx/files/653373/station.zip
[2] https://github.com/weewx/weewx/pull/188#issuecomment-267188828
[3]

tkeffer:
> Yes, that works for me.   I ponder though if it's worth the bother to
add the flag in weewx.conf for an OS option. 

It's definitely not. This solution is much more robust and makes weewx.conf OS independent. 

If you're concerned about run time, you could factor out the loading of the DLL so it is done only once. That's likely to be the slowest part by far.

richterb1:
Yes.  I'm thinking another option in porting instructions, let the
porter enable the right test when making a port. Basically now all the
OSes should be 'covered'.  But at the same time might not be worth it. 
I like it even if my OS is the last in the test. 

On 2016-12-14 04:52 PM, Tom Keffer wrote:

>> Yes, that works for me. I ponder though if it's worth the bother to
>> add the flag in weewx.conf for an OS option.
> 
> It's definitely not. This solution is much more robust and makes weewx.conf OS independent. 
> 
> If you're concerned about run time, you could factor out the loading of the DLL so it is done only once. That's likely to be the slowest part by far. 
> 
> --
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].
 

Links:
------
[1] https://github.com/weewx/weewx/pull/188#issuecomment-267203909
[2]

matthewwall:
if you don't like nested try/except, you could always conditionally branch based on platform.platform()

tkeffer:
Yes you can, but that requires explicitly enumerating platforms. The present algorithm is more general, not  to mention more Pythonic. It should also be quite fast.

In any case, the database lookups will swamp any minute differences in timings.

richterb1:
I spent more time trying to shave a few CPU cycles than coming up with
the final solution. I'd guess the test overhead is simple.  I'd also
think if there was a note in the porting instructions it could suggest
that the correct routine be enabled for the package/port. Those building
from source would in theory be aware of the 'option' to streamline a
bit.  

Most importantly it works for me and I'm happy. 

On 2016-12-14 05:56 PM, Matthew Wall wrote:

> if you don't like nested try/except, you could always conditionally branch based on platform.platform() 
> 
> --
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].
 

Links:
------
[1] https://github.com/weewx/weewx/pull/188#issuecomment-267214670
[2]

richterb1:
Agreed.  

On 2016-12-14 05:59 PM, Tom Keffer wrote:

> Yes you can, but that requires explicitly enumerating platforms. The present algorithm is more general, not to mention more Pythonic. It should also be quite fast. 
> 
> In any case, the database lookups will swamp any minute differences in timings. 
> 
> --
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].
 

Links:
------
[1] https://github.com/weewx/weewx/pull/188#issuecomment-267215107
[2]

PR #189, opened 2016-12-15
Add more documentation for configuration of WOW upload

alfh:
The WOW website where you enter a new site does not
have information on the username and password to
use when uploading to WOW.

So it should help to document that you need to read
another WOW page on how to find the station id, and
that you need to set the password yourself for the
site, and it should be a 6 digit pin code.


alfh:
I also sent an email to WOW to get weewx added to the list of supported software on 
http://wow.metoffice.gov.uk/support/dataformats#automatic

PR #194, opened 2017-01-11
prevent KeyError and update to txTempBatteryStatus

manos:
I can't tell if I'm getting txTempBatteryStatus instead of txBatteryStatus simply because my station is an accurite.. but I figured I'd ask the question in the form of a pull request.

I'm also seeing sensor_battery, so I'm less confident now...
Jan 11 14:01:34 weatherasp weewx[11567]: packet received: {'sensor_id': 645, 'outHumidity': 61, 'maxSolarRad': None, 'altimeter': None, 'heatindex': 40.5, 'inDewpoint': None, 'channel': None, 'barometer': None, 'windchill': 31.740679430262393, 'dewpoint': 28.123793591658, 'humidex': 40.5, 'pressure': None, 'sensor_battery': 0, 'rxCheckPercent': 100.0, 'rainRate': 0, 'usUnits': 1, 'txTempBatteryStatus': 0, 'appTemp': 26.

### DIFF
--- a/examples/lowBattery.py
+++ b/examples/lowBattery.py
@@ -108,7 +108,7 @@ class BatteryAlarm(StdService):
         """This function is called on each new LOOP packet."""
 
         # If the Transmit Battery Status byte is non-zero, an alarm is on
-        if event.packet['txBatteryStatus']:
+        if event.packet.get('txTempBatteryStatus', 0):
 
             self.alarm_count += 1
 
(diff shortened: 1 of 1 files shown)


matthewwall:
thanks for picking this up!  however, the bug is in the acurite driver - the key should be outTempBatteryStatus, not txTempBatteryStatus.  that bug has been lurking since 15jan2015!

you should change txTempBatteryStatus to outTempBatteryStatus in acurite.py

fixed at commit 5b65d4e

until the development branch is merged back into master, you can try the new lowBattery example from here:

https://github.com/weewx/weewx/blob/development/examples/lowBattery.py

manos:
Awesome, thanks @matthewwall !

PR #196, opened 2017-01-14
Add last_<x> tag periods.

dcapslock:
Add tag time periods of last_hour, last_day, last_week, last_month and last_year. Add month_delta and year_delta to span.

You can iterate over those periods. See http://carlingfordweather.sydney/TEST/tag-test.txt for Top Temperatures over the last immediate week.


dcapslock:
This change will assist in easily creating cheetah templates to create data files (json or other) to have data that allows client graphing to parallel the current ImageGenerator. A further change will be to allow iteration of all records in a TimeBinder.

tkeffer:
Wow! It's rare when we get a PR so nicely put together!

Give me a chance to look this over. Should not take long. We should also include it in the test suites.

tkeffer:
Thank you. If you get a chance, and you can follow the logic of the template test suite, it would be nice to complement this with some tests.

PR #199, opened 2017-01-14
Implement .spans(interval=seconds) tag timespan iterator.

dcapslock:


dcapslock:
Allows for implementing data file generation for client graphing which will parallel current Image generation which supports an aggregation interval.

tkeffer:
Hello, again, and so soon. 

We're missing 3 things:

1. You need to define the semantics of `roundTS`, include it in the docstring, and provide a couple of tests in the doctest. It's not obvious to me.

2. You also need to document `$spans`, in the customizing guide.

3. Finally, include a test. See `bin/weewx/test/test_skins/StandardTest` for the other iterator tests. Follow the same pattern: define a `byspan.txt.tmpl`, include it in the provided `skin.conf`, then include US Customary and Metric "expected" values in the subdirectories below `bin/weewx/test/test_skins/expected/StandardTest`.

dcapslock:
I have added documentation to intervalgen and a test in the weeutil test. I am having trouble firing up the general weewx tests as I do not have MySQL installed. I need to do this another day. Please consider merging this PR for now.

(Any more info on running tests?)

dcapslock:
OK. I have created a test case and tested with sqllite, commenting out MySQL. 

I note that index.html failed for me due to various use of .format("%x %X") which is different due to month order of my local date format. I have not attempted to fix this but will raise an issue - all will be well as long as Unit testers always have M/D/Y local date format.

tkeffer:
Please note that most development is occurring off the `development` branch now.

tkeffer:
I found several problems with this. Basically, I regard a tag like `$last_month` to mean the entire prior month. For example, for this moment, `$last_month` was the prior month, that is, December 2016. Instead, the code regarded it as the previous 30 days. Depending on the length of a month, this may not even end up on the same day.

I have pushed my code to `development`. Unfortunately, it does not pass the test suites, probably because of this difference in semantics. I haven't finished sorting this out. Perhaps you could take a look.

dcapslock:
Perhaps the semantics of the tag name is the main question. The use case for the tags, $last_week, $last_month etc. is to be able to use cheetah generator to pull out data to use for client generated graphs that would be able to parallel the plots in standard skin. Perhaps $prior_<x>?

matthewwall:
currently we have: $hour, $hours_ago($hours_ago=h), $day, $days_ago($days_ago=d), $yesterday, $week, $month, $year, $rainyear.

is the intent to add $weeks_ago($weeks_ago=w), $months_ago($months_ago=m), $years_ago($years_ago=y)

btw, $yesterday is the same as $days_ago($days_ago=1), right?

then $prior_month (or $previous_month) would be synonymous with $months_ago(1).  (i do not think that synonyms are necessary)

dcapslock:
My intent is to have a method to get the immediate prior time span, not the last ordinal timespan. Yes, could have $weeks_ago etc. but it is not the intent.

Take today Sat 21/1, $weeks_ago($weeks_ago=1) would be (assuming week start of Sunday) Sun 8/1 til Sat 14/1. What I am after is Sat 14/1 til Sat 12/1 (last data).

tkeffer:
Would the present `$span` work?

dcapslock:
Was thinking so Tom. Use $span, with the changes, but remove $last_ as I can see the confusion.

tkeffer:
Instead of introducing a whole bunch of `$last_xxx` tags, we can alternatively introduce `ago` parameters. For example, instead of `$last_week`, we would have `$week($weeks_ago=1)`. And so on. That generalizes things so we can get to the week-before-last, etc.

tkeffer:
Let me work on this a bit.

dcapslock:
I am happy remove $last_, update the test cases and documentation.

tkeffer:
I have removed the `$last_xxx` tags, and brought the test suites up to date. I have not updated the documentation.

Please let me know if the present tag interface fulfills your needs!

dcapslock:
Mostly. The $span tag will allow to match what plots do for a data span. However, removing the rounding feature of $spans & intervalgen means that it will not be possible to get, say, hourly spans that lie on ordinal boundaries. The image generator first standardises its time span (plot x axis) before calling intervalgen. There is no way to do this in the tag system.

What I am trying to achieve.

1. Get a span for prior week (plot 'week'). Yes, can do with $span.
2. Aggregate data for every hour. Yes, can do with $spans BUT will be on boundaries of the time span which will at most times not be ordinal.

Say the current time is Sun 6:05am. If I am using the image generator to get a week plot, aggregate 1 hour, I will get Sat (-1w) 00:00 til Sun (today) 24:00, with aggregate spans 00:00-01:00 etc.

With the current code I can't get that with the tag system. I would get Sun (-1w) 6:05am til Sun (Today) 6:05am, with aggregate spans 6:05-7:05 etc.

I am happy to leave this for now to get 3.7 out. It is not so important as to delay a release.

PR #213, opened 2017-02-05
Reworked check_strings() output

gjr80:
Submitted this as a PR since I have deviated substantially from the old `check_strings()` - thought this way it would be easier kick this into touch if its no good. Still a few other minor things to tweak but this implements the main changes to `check_strings()` output.

Reworked `check_strings()` screen and syslog output:
- now gives progress à la `--rebuild-daily`
- syslog is silent for both `--check-strings` and `--fix-strings` when used with `--dry-run`
- left 'Preparing' (rather than 'Starting') but added 'this may take a while' as there is a significant delay in `dbmanager.genBatchRows()` initialising at line 619 (well there was some 30 seconds for my 400k records). I think this is long enough we need to tell the user something but I don't see anything else we can do other than the 'this may take a while'

### DIFF
--- a/bin/wee_database
+++ b/bin/wee_database
@@ -21,7 +21,7 @@ import weedb
 import weewx.manager
 import weewx.units
 
-from weeutil.weeutil import TimeSpan
+from weeutil.weeutil import TimeSpan, timestamp_to_string
 
 usage = """wee_database --help
        wee_database --create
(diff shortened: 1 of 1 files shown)


tkeffer:
Just wondering: why do you and Matthew have an aversion to starting sentences with a capital letter?

gjr80:
Oh, Matthew is much more hard core than me, for comments if I have more then one sentence then I will capitalise otherwise lowers. And my emails use initial capitals!

PR #216, opened 2017-02-18
wunderfixer: add -r option to allow a range of days to be checked

Jafaral:
wunderfixer works on one day (specific date) at a time. I had numerous occasions
where a station I manage goes offline for a few days before it comes back online.
I used scripts/cron to take care of this but I thought adding support to the tool itself
is more convenient and would be the best option on the long run. If -r is not  present
(The default for -r is 1) the tool works as before, otherwise a loop is run to cover all
days from the given date going back the specified number of days. Examples:

wunderfixer -r 1  :  check today , i.e same as before wunderfixer -d "today's date"
wunderfixer -r 7  : check all of the past week

wunderfixer -d 2017-01-31 -r 7  : check the last week of January of 2017
wunderfixer -d 2017-02-3 -r 3   : check the first 3 days of February of 2017

Signed-off-by: Jafar Al-Gharaibeh <to.jafar@gmail.com>

### DIFF
--- a/bin/wunderfixer
+++ b/bin/wunderfixer
@@ -113,7 +113,11 @@ def main() :
                       "Default is to take from configuration file.")
     parser.add_option("-d", "--date", type="string", dest="date", metavar="YYYY-mm-dd",
                       help="Date to check as a string of form YYYY-mm-dd. Default is today.")
-    parser.add_option("-e", "--epsilon", type="int", dest="epsilon", metavar="SECONDS", 
+    parser.add_option("-r", "--range", type="int", dest="dayrange",  metavar="DAYS",
+                      default=1,
+                      help="The number (range) of days to be checked."
+                      "Starting from the date specified by --date going backward")
+    parser.add_option("-e", "--epsilon", type="int", dest="epsilon", metavar="SECONDS",
                       default=120,
                       help="Timestamps within this value in seconds compare true. Default "
                       "is 120.")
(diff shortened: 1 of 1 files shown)


Jafaral:
sample output:

./wunderfixer /etc/weewx/weewx.conf -v -r 3
Using configuration file /etc/weewx/weewx.conf.
Using database binding 'wx_binding', which is bound to database 'archive_sqlite'

Weather Underground Station:   SOMESTATION
Date to check:                  2017-02-17
Number of days:                 3

=== Day 0 === 
Date to check:                 2017-02-17
Number of archive records:     272
Number of WU records:          271
Number of missing records:     1

Missing records:
2017-02-17 22:25:00 CST (1487391900); 29.606";  43.1F;  83%; 3.4 mph; 280 deg; 4.7 mph gust;  38.2F; 0.00" rain  ...published.

=== Day 1 === 
Date to check:                 2017-02-16
Number of archive records:     287
Number of WU records:          287
Number of missing records:     0

=== Day 2 === 
Date to check:                 2017-02-15
Number of archive records:     288
Number of WU records:          288
Number of missing records:     0

tkeffer:
Thanks for your contribution. I am out of the country right now, but I'll take a look at your PR when I get back.

joanpau:
May I suggest to use the [ISO 8601][1] format to represent the desired date or time interval?
Then there would be no need for a new option argument, the current `-d` option would suffice
(and it would be backward compatible):
  - `wunderfixer -d 2017-03-01`: check all records in the given day (same meaning as now )
  - `wunderfixer -d 2017-02-01/2017-03-01`: check all records in the given period (1month)
  - `wunderfixer -d 2017-02-01P1M`: check all records in the given period (same as above)

  [1]: <https://en.wikipedia.org/wiki/ISO_8601>

Jafaral:
@joanpau Thanks for your input. I actually thought about that initially; extending -d to support the new feature but after looking at the source code I ended up adding "-r" because it  is more concise at the command line and was less disruptive coding-wise.  There is also no backward comparability concern either way. ISO 8601 options are an overkill for  wunderfixer use cases. If people prefer to go  with it (extending -d) I probably could do that.

tkeffer:
With the utility [`wee_database`](https://github.com/weewx/weewx/blob/master/bin/wee_database) we've introduced a notation to reflect a range of dates. Take a look at it. Basically, you can do either

    --date=2017-03-08

or

    --from=2017-03-02 --to:2017-03-05

but not both.

Jafaral:
Do you suggest I change the code to use that format and do a new PR?

~~~
--from=2017-03-02 --to:2017-03-05
~~~

tkeffer:
Yes. Let's not introduce a new notation.

1. If no date is provided, it defaults to today's date.

2. If the `--date` option is used, then that single date would be used.

3. Alternatively, both `--from` and `--to` can be provided. In this case, the inclusive dates would be used.

Option 2 or option 3 can be specified, but not both.

tkeffer:
Some other comments:

1. Most users would find it counterintuitive that you start with the last date, then work backwards in time.

2. The line

        date_tt = time.strptime(str(datetime.date.today()), "%Y-%m-%d")

    is an inefficient way of getting today's date as a timetuple. Why not just use

        date_tt = time.timetuple()

3. The expression `range(0, drange)` can be more simply stated as`range(drange)`

4. Instead of

        date_date = datetime.date(date_tt[0], date_tt[1], date_tt[2]) - datetime.timedelta(day)

why not just precalculate a starting `datetime` (which the original wunderfixer had), then subtract (or, better, add), the timedelta?

Jafaral:
Thanks @tkeffer , I will update the code to use the notation you described. And yes, I will fix those python issues, new to the language :-) !

tkeffer:
No activity.

PR #228, opened 2017-03-12
Qc development

dcapslock:
QC Development including updated simulator driver to generate spikes.

TODO: Test suite.


tkeffer:
This looks promising.

However, the existing test suites use the existing simulator. Either the generation of spikes should be made optional, or their pattern should be deterministic and the expected results of the test suite changed accordingly.

The latter is probably a better solution, as it will test the rest of the weewx machinery in the presence of spikes.

Because it will break the test suites, let's wait to include this PR until the suites are updated.

dcapslock:
Test suite for simulator would not have been broken as spike defaults to not active. However, I have updated test_sim.py and simgen.conf so that test simulator does both a general test and then a test with spikes. NOTE: The test case with spikes will only match to one decimal place due to the test simulator run having the non spike values included.

dcapslock:
Any update on considering this PR?

tkeffer:
Sorry. I haven't forgotten about you! I got distracted by V3.7, plus your patch requires some careful thinking.

### General comments:

Detecting anomalies in real time is hard. Even in hindsight, patching up a timeseries can be difficult.

You're implicitly adopting a model of how you expect these anomalies to show up: there will be a relatively smooth series of points, then a jump in value, then (presumably) a jump back. But, what if there was a step function: a jump in value, but no jump back? How would the algorithm respond? Or, what if there was a hat function: a jump up in value for two or more data points, then a jump back?

It would be useful to list the possibilities and what you expect the algorithm to do.

Along the same line, you're doing spike detection for both LOOP packets and archive records. But, remember, archive records are aggregate values of the LOOP packets. If you properly filter the LOOP packets, there should be no spikes in archive records.

The algorithm you are using is not terribly complicated, but it gets obscured by a lot of unnecessarily wordy code. In particular, you introduce temporary variables that are not really necessary, but their presence requires some cognitive understanding.

Finally, a few more comments about what you're doing would be welcome. I got lost here and there.

### Specific comments:

1. Rather than one giant `__init__` function, the code would be easier to follow if it were broken up in to separate functions to initialize for min-max detection, and for for spike detection. Same with the function `apply_qc`.

2. Because you are using the database solely to "prime the pump" and supply values for `cache_record`, it's not necessary to have the database appear in the initializer. Why not just require that `cache_record` be passed in to the initializer? ("Push policy up, implementation down") This way you decouple the database from the algorithm. To make it easier for the user of class `QC` to supply the cache record, you can include a helper function that includes the logic of fetching the record from the database.

3. If you get error `weedb.NoDatabaseError` can you really just `pass` (line 89). I would think you'd have to abandon detecting a spike, but I may be misunderstanding what's going on.

4. Sometimes your code is more wordy than it needs to be. For example, the section

    ~~~python
                if db_record is not None and db_record.has_key(obs_type) and db_record[obs_type] is not None:
                    rec_data = db_record[obs_type]
                    rec_dateTime = db_record['dateTime']
                else:
                    rec_data = None
                    rec_dateTime = None

                cache_record[obs_type] = {}
                cache_record[obs_type]['data'] = rec_data
                cache_record[obs_type]['dateTime'] = rec_dateTime
    ~~~
 
   can be more simply expressed as:  

    ~~~python
                cache_record[obs_type] = {}
                if db_record is not None and db_record.has_key(obs_type) and db_record[obs_type] is not None:
                    cache_record[obs_type]['data'] = db_record[obs_type]
                    cache_record[obs_type]['dateTime'] = db_record['dateTime']
                else:
                    cache_record[obs_type]['data'] = None
                    cache_record[obs_type]['dateTime'] = None
    ~~~

5. Along the same line, and assuming you really want to filter archive records, instead of

    ~~~python
            self.last_records['LOOP'] = cache_record
            self.last_records['Archive'] = cache_record
    ~~~

    why not just  

    ~~~python
            self.last_loop = cache_record
            self.last_record = cache_record
    ~~~

    or, if you don't filter archive records, you can flatten things even more;  

    ~~~python
            self.last_loop_value[obstype] = db_record[obstype]
            self.last_loop_time[obstype]  = db_record['dateTime']
    ~~~

6. Line 145: if you follow my suggestion to separate out the min-max logic from the spike logic, rather than have a huge indented block conditional on whether `data_type` is not a null string, you can just return from the "spike" function.

7. No need to use `deepcopy` to copy a record. Just use

    ~~~python
    cache_record = dict(last_record)
    ~~~
 
8. Not sure what you're trying to do with:

    ~~~python
                        spikecheck = data_dict[obs_type] - last_record[obs_type]['data']
                        try:
                            spikecheck = abs(spikecheck / (data_dict['dateTime'] - last_record[obs_type]['dateTime']))
                        except ZeroDivisionError:
                            continue
    ~~~
 
    Why not just:  
 
    ~~~python
                        try:
                            spikecheck = abs(spikecheck / (data_dict['dateTime'] - last_record[obs_type]['dateTime']))
                        except ZeroDivisionError:
                            spikecheck = data_dict[obs_type] - last_record[obs_type]['data']
    ~~~
 
    This makes your intentions more explicit.

dcapslock:
Tom, thanks for your review. Much appreciated in helping me understand the best structure for Weewx and non wordy Python code (only learnt Python to be able to program Weewx)

> You're implicitly adopting a model of how you expect these anomalies to show up: there will be a relatively smooth series of points, then a jump in value, then (presumably) a jump back. But, what if there was a step function: a jump in value, but no jump back? How would the algorithm respond? Or, what if there was a hat function: a jump up in value for two or more data points, then a jump back?

> It would be useful to list the possibilities and what you expect the algorithm to do.

Tried to make it uncomplicated. It is as per name, 'Spike' check. Although I understand your query on a jump up with no jump back. The code would never allow a jump up above the threshold and will keep dropping the observation until a jump back - so I will implement a configurable timeout. For a hat function, the hat obs will be dropped until a no hat obs comes through.

> Along the same line, you're doing spike detection for both LOOP packets and archive records. But, remember, archive records are aggregate values of the LOOP packets. If you properly filter the LOOP packets, there should be no spikes in archive records.

Correct me if I am wrong, but I believe that hardware can send through 'Archive' packets if the hardware supports it? In my custom driver this is what I do.

> Rather than one giant  __init__  function, the code would be easier to follow if it were broken up in to separate functions to initialize for min-max detection, and for for spike detection. Same with the function  apply_qc .

Good point. I was just following the current code. Improving the code is a good idea, especially with the adding in of another QC element.

> Because you are using the database solely to "prime the pump" and supply values for  cache_record , it's not necessary to have the database appear in the initializer. Why not just require that  cache_record  be passed in to the initializer? ("Push policy up, implementation down") This way you decouple the database from the algorithm. To make it easier for the user of class  QC  to supply the cache record, you can include a helper function that includes the logic of fetching the record from the database.

Good idea. At the current time, having everything in qc.py has meant easier testing. However for clean code I agree on your approach.

> If you get error  weedb.NoDatabaseError  can you really just  pass  (line 89). I would think you'd have to abandon detecting a spike, but I may be misunderstanding what's going on.

Yes. Though I see it is not clear. What will happen in this case is that there will be no Spike checking on the first packet, and the first packet value will then be the reference, rather than the database read.

So from here I will:

1. Implement a timeout for spike checking to allow for a jump that never returns.
2. Restructure the reading of cache from Database
3. Include some more commenst to explain code (like pass on weedb.NoDatabaseError - also some debug output in this case woule be worthwhile.
4. Clean up wordy code

There are a handful of users testing all with good results so I am keen to get this in a release soon. The most common use cases are A. wireless hardware glitching without so much a reason. B. Wireless hardware getting corruption from HAM radio operation.

Probably won't be til next weekend that I have time to tidy this up.

Thanks again

Darryn

dcapslock:
OK. Code updated. Please review again. Thanks

- QC now relies on QC object classes with QcMinMax and QcSpike classes. This structure would also allow a user QC object without making a full service. This allows some flexibility as well as the user determining the order in which QC objects are applied (spike first of MinMax first)
- Weeimport updated with the QC objct changes
- Manager now has a lastGoodRecord() function which StdQC Engine uses to grab a db_record to pass to the QC objects
- Wordy code reduced.
- Upgrade will need to handle [StdQC]objects=weewx.qc.QcMinMax to be backward compatible. Note sure how to handle
- Will update doco once this PR is merged with a new PR. As well as to handle any needs for upgrading the [StdQC] section.

tkeffer:
Hello again.

1. Pretty close, but I think you're making this a bit more complicated than necessary. In particular, there are no real advantages to loading the QC modules dynamically. You're basically replicating the weewx "service module" architecture. If you think that your spike detector is just the beginning of a suite of possible QC modules, then why not just make them independent weewx service modules? Personally, I don't think that's a good idea. I would just modify the `StdQC` module to do min/max and spike detections and leave it at that.

2. Functions `lastGoodRecord()` and `firstGoodRecord()` are unnecessary. Just use
   ~~~python
   manager.getRecord(manager.lastGoodStamp(), max_delta) if timestamp else None
   ~~~

3. No longer passes the test suites. In particular, I got a bunch of errors of the sort:

   ~~~
   ======================================================================
   ERROR: test_archive_data (__main__.TestSqlite)
   ----------------------------------------------------------------------
   Traceback (most recent call last):
     File "bin/weewx/test/test_sim.py", line 77, in setUp
       engine = weewx.engine.StdEngine(self.config_dict)
     File "/home/tkeffer/git/weewx/bin/weewx/engine.py", line 76, in __init__
       self.loadServices(config_dict)
     File "/home/tkeffer/git/weewx/bin/weewx/engine.py", line 140, in loadServices
       self.service_obj.append(weeutil.weeutil._get_object(svc)(self, config_dict))
     File "/home/tkeffer/git/weewx/bin/weewx/engine.py", line 431, in __init__
       self.qc_obj.append(weeutil.weeutil._get_object(obj)(config_dict, db_record=db_record))
     File "/home/tkeffer/git/weewx/bin/weewx/qc.py", line 157, in __init__
       cache_record[obs_type]['data'],
   KeyError: 'data'
   ~~~

   and a permission error:

   ~~~
   PermissionError: (1044, "Access denied for user 'weewx1'@'localhost' to database 'test_sim_spike'")
   ~~~

tkeffer:
Anything new on this?

dcapslock:
Finding time to rework back to not using service and the extra manager functions. Please let me know of anything else so this can my last work on this.

tkeffer:
> Please let me know of anything else so this can my last work on this.

Well, I can't promise that. Once it goes into distribution, I can guarantee that someone will have some issue that we didn't foresee! :-)

dcapslock:
Yes, always will be issues. What I wanted to explore is if there was further to tweak to meet you desire or standards.

tkeffer:
Any progress on this?

tkeffer:
Closing due to inactivity. We can always reopen.

PR #232, opened 2017-03-14
move aggregated plot points to the middle of the aggregation interval

paolobenve:
The point is moved only if aggregation type isn't bar

I needed to redefine the `__sub__` method for ValueTuple because the first item in the tuple could be a list

### DIFF
--- a/bin/weewx/imagegenerator.py
+++ b/bin/weewx/imagegenerator.py
@@ -18,6 +18,7 @@ import weeutil.weeutil
 import weewx.reportengine
 import weewx.units
 from weeutil.weeutil import to_bool, to_int, to_float
+from weewx.units import ValueTuple
 
 #===============================================================================
 #                    Class ImageGenerator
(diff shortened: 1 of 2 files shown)


tkeffer:
Thanks. A few questions:

What problem are we trying to solve? Is the position of the x variable causing a problem for you?

If you've redefined `__sub__` to allow subtraction from a list or tuple, why do you need to explicitly invoke it at lines 151 & 152?

You are assuming that the aggregation type is average, but the user could have specified something else. This will require pushing knowledge of the aggregation type into imagegenerator. Unless you can find a simple way of doing this as well as a compelling reason to do so, I'm inclined not to make this change.

paolobenve:
Ok, I restricted the correction to avg, min and max. I think the correction is reasonable at least for those aggregations. Anyway, it's easy to extend it to other aggregation types.

I reset units.py and moved to imagegenerator.py all the stuff that manages lists in value tuple.

                         * * *

The problem I want to solve: without this patch, lines from aggregated values are plotted `aggregate_interval / 2` to the rigth, out of phase with non aggregated graphs.

See the difference on a daily-aggregated average in weekly graph:

with the patch: the avg values are located in the middle of the aggregation time period, their value reflects the average of the values around the time where the point is plotted

weektempprova

without the patch: the average curve is plotted 12 hours later
weektempprova 1

3-hours aggregated average:

with the patch: the average graph is in sync with the non-averaged values plot
weektempprova 3

without the patch: out-of-phase average plot
weektempprova 2

The example plot in http://www.weewx.com/docs/customizing.htm#aggregation_types shows the out-of-phase too:
daytemp_with_avg

tkeffer:
That's a pretty compelling point. You're (slowly) convincing me. 

Could you try something for me? Instead of explicitly testing for a `tuple` or `list`, follow Python's EAFP pattern and use an exception that tests for a `TypeError`. Something like

~~~python
try:
    start_vec_t = ValueTuple([x - aggregate_interval / 2 for x in start_vec_t[0]], start_vec_t[1], start_vec_t[2])
except TypeError:
    start_vec_t = ValueTuple(start_vec_t[0] - aggregate_interval / 2, start_vec_t[1], start_vec_t[2])
~~~

This way, any kind of iterable object can be supported.

paolobenve:
what about now?

tkeffer:
Merged into `development` branch at commit https://github.com/weewx/weewx/commit/6be61c058689de4fa93730b15b3781708450a781

Simplified logic at commit https://github.com/weewx/weewx/commit/a2bdabebdff7f37d0791666e84b06f9c3227b812. There's no reason why the start and stop vector would ever be anything besides a vector.

Added note to change log.

Thanks, Paolo.

paolobenve:
thank you! :-)

PR #236, opened 2017-03-16
Permite the user to plot some line in the grid with a different colors

paolobenve:
I added 3 new variables:

- chart_gridline_bold_color, inspired to chart_gridline_color, with default value 0x545454 (a darker gray): the color for plotting some other line, as specified by the spacing defined by the following 2 variables
- chart_gridline_bold_x_spacing: every how many grid lines the vertical darker line is to be plotted
- chart_gridline_bold_y_spacing: every how many grid lines the horizontal darker line is to be plotted

chart_gridline_bold_x_spacing and chart_gridline_bold_y_spacing default to 1, which means no darker lines

An example:

daytemp

The docs update is still missing, I'm going to do it afterwards

### DIFF
--- a/bin/weeplot/genplot.py
+++ b/bin/weeplot/genplot.py
@@ -44,7 +44,11 @@ class GeneralPlot(object):
         self.image_background_color = weeplot.utilities.tobgr(config_dict.get('image_background_color', '0xf5f5f5'))
 
         self.chart_background_color = weeplot.utilities.tobgr(config_dict.get('chart_background_color', '0xd8d8d8'))
-        self.chart_gridline_color   = weeplot.utilities.tobgr(config_dict.get('chart_gridline_color',   '0xa0a0a0'))
+        self.chart_gridline_color        = weeplot.utilities.tobgr(config_dict.get('chart_gridline_color',   '0xa0a0a0'))
+        self.chart_gridline_bold_color   = weeplot.utilities.tobgr(config_dict.get('chart_gridline_bold_color',   '0x545454'))
+        self.chart_gridline_bold_x_spacing = int(config_dict.get('chart_gridline_bold_x_spacing', 1))
+        self.chart_gridline_bold_y_spacing = int(config_dict.get('chart_gridline_bold_y_spacing', 1))
+
         color_list                  = config_dict.get('chart_line_colors', ['0xff0000', '0x00ff00', '0x0000ff'])
         fill_color_list             = config_dict.get('chart_fill_colors', color_list)
         width_list                  = config_dict.get('chart_line_width',  [1, 1, 1])
@@ -271,8 +275,12 @@ class GeneralPlot(object):
 
         drawlabelcount = 0
         for x in weeutil.weeutil.stampgen(self.xscale[0], self.xscale[1], self.xscale[2]) :
-            sdraw.line((x, x), (self.yscale[0], self.yscale[1]), fill=self.chart_gridline_color,
-                       width=self.anti_alias)
+            # plot bold color lines if requested
+            fill=self.chart_gridline_color
+            if self.chart_gridline_bold_color and self.chart_gridline_bold_x_spacing > 1 and drawlabelcount % self.chart_gridline_bold_x_spacing == 0 :
+                fill=self.chart_gridline_bold_color
+            
+            sdraw.line((x, x), (self.yscale[0], self.yscale[1]), fill=fill, width=self.anti_alias)
             if drawlabelcount % self.x_label_spacing == 0 :
                 xlabel = self._genXLabel(x)
                 axis_label_size = sdraw.draw.textsize(xlabel, font=axis_label_font)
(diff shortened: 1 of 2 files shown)


matthewwall:
i would suggest that you avoid styles in the parameter names.  for example, use 'major' instead of 'bold'.

you should not have to specify spacing - just use the spacing of the axis labels, otherwise you must duplicate all of the axis logic (xscale, yscale) for gridlines.

you will also need a mechanism to turn this feature on/off.  not everyone uses/wants grids in their plots.

i would punt the 'chart_' prefix - what other kind of gridline is there?

so you should be able to do what you propose with only two parameters:

```
# if non-none, then draw major gridlines on top of grid; pixels
gridline_major_width = None
# if none, use the gridline color
gridline_major_color = None
```

i would make the default behavior to be None for both of the parameters, then make the day and week plots have non-none (analogous to the day/night bands).  otherwise the plots will be too busy.

paolobenve:
The two spacing parameters are needed, because the user may want different spacings on the two axis.

Besides that a darker grid that doesn't coincide with the labels is one more oppurtunity too; making it coincide with the labels makes no sense when labels spacing is 1 (labels on all the grid lines).

The feature is turned off with all the variables to their default value. I put 1 as spacings default value, which means no darker grid. None could be put as default too, the code would only have some more line.

I used the `chart_` prefix following the code conventions, namely the existing variable `chart_gridline_color`: the prefix cannot be punt from that variable too, because that would break the users skins.

Defining a different width for the darker grid: at first I had experimented with it, but the result is unpleasant, even with a 2-pixels width, so I decided to work only on the grid color.

paolobenve:
The variables names are now:

- chart_gridline_bold_color
- chart_gridline_bold_x_spacing
- chart_gridline_bold_y_spacing

matthewwall:
paolo,

we *can* change the name of existing chart_gridline parameter if necessary.  if we did that, then we simply make chart_gridline_foo do the same thing as gridline_foo for backward compatibility.  imho, that is preferable to adding useless baggage to every additional gridline parameter.

what are the units of major_x_spacing and major_y_spacing?

what happens if someone specifies bogus spacing values?  what are bogus spacing values?

what other things should a user be able to do to the axes?  to the tic marks?  to the grid?  to the grid labels?  to the axis labels?

what if i want to show only horizontal gridlines?  only vertical gridlines?

what if i want no gridlines at all?

i do not understand why a value of 1 would mean no major gridlines.  that does not seem intuitive.

i still think that you should not force people to enter a spacing parameter - it should just work with no configuration required.

if grid labels fall on every grid line, then the plot is probably too busy anyway, and the grid label algorithm probably has to be revisited.

i am not convinced that it cannot be done with no spacing parameter.

imho, this kind of change to the plot configuration options should not be done in isolation.  lots of ad-hoc changes result in a user-unfriendly syntax and backward-compatibility nightmare.  better to lay out a broader plan/design, then make sure these changes fit.

paolobenve:
Changing the name of existing and new parameter: no problem!

the units of major_x_spacing and major_y_spacing are grid lines: a value of 2 means that 1 line every 2 must be major

values for major_x_spacing and major_y_spacing and converted to integers and used only if > 1

you're right that the default value of 1 is not intuitive: I'm going to change it to None

people are not forced to enter a spacing parameter, the skin just work as before if this parameters are not used

The grid labels cannot fall on every grid line, because the spacing parameters are used only if > 1

paolobenve:
I'm not changing the name of the `chart_gridline_color` parameter, I am not experienced enough

tkeffer:
The code

~~~python
        self.gridline_major_x_spacing = config_dict.get('gridline_major_x_spacing', None)
        if self.gridline_major_x_spacing:
            self.gridline_major_x_spacing = int(self.gridline_major_x_spacing)
~~~

can be simplified to

~~~python
from weeutil.weeutil import to_int

    ...
        self.gridline_major_x_spacing = to_int(config_dict.get('gridline_major_x_spacing')
~~~

I don't understand why you are passing in the gridline spacing through class `PlotLine`. You are implying that each line can have its own spacing. They can't. The gridline spacing has nothing to do with individual lines.

I disagree with Matthew on names. The pattern has already been set. We have `chart_gridline_color`, so it's only logical that we would have `chart_gridline_color_major` or something similar.

Finally, this is the *second* time you have submitted a pull request that simply does not run at all. Sorry for being so brutally blunt, but In the future you must check your work or I am going to stop looking at it. You can't just type and submit.

paolobenve:
I ended running development branch on my system, and the two files in this pull request reflect the changes from development branch.

I removed x/y_label_spacing and chart_gridline_major_* variables from PlotLine.

I used to_int function on possibly None value, as adviced.

@tkeffer, I beg your pardon for submitting code which didn't work, now on my system everything seems working fine...

The only thing I don't understand is why the patch that plots avg, min and max in the center of aggregation interval is shown here among the difference between my branch and weewx development branch, since I had understood that that patch had already been merged into weewx's development branch...

tkeffer:
24 commits and 13 files changed? Way too complicated.

The patch that plots in the center of the aggregation interval was done in commit https://github.com/weewx/weewx/commit/a2bdabebdff7f37d0791666e84b06f9c3227b812

paolobenve:
I think I have messed up my branches.... it was only two files

PR #241, opened 2017-03-20
Provide system information for FreeBSD 

richterb1:
I'm struggling to get  my local copy in sync with the master, I've not had much success.  But this is the output from wee_debug --info on FreeBSD11.

System info

  Operating system: ('FreeBSD', 'freebsd1.9500years.com', '11.0-RELEASE-p8', 'FreeBSD 11.0-RELEASE-p8 #8: Mon Mar  6 01:04:05 PST 2017     root@freebsd1.9500years.com:/usr/obj/usr/src/sys/RPI-B', 'arm', 'armv6')
  Release:   11.0-RELEASE-p8
  Platform:  FreeBSD-11.0-RELEASE-p8-arm-armv6-32bit-ELF
  Processor: armv6
  System:    FreeBSD
  1 minute load average:  0.41
  5 minute load average:  0.27
  15 minute load average: 0.20


tkeffer:
I sympathize how tough it can be to get an old branch back in sync with the current master. However, this PR would touch 134 files! That's just way too many to review.

You may have to retire your old fork, do a fresh fork, then apply the changes to that.

richterb1:
Yes, agree. I was going to do that as a last resort.  Let me jiggle git
again. 

On 2017-03-20 04:56 AM, Tom Keffer wrote:

> I sympathize how tough it can be to get an old branch back in sync with the current master. However, this PR would touch 134 files! That's just way too many to review. 
> 
> You may have to retire your old fork, do a fresh fork, then apply the changes to that. 
> 
> --
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].
 

Links:
------
[1] https://github.com/weewx/weewx/pull/241#issuecomment-287739255
[2]

richterb1:
This makes the changes suggested by mwall and should only touch  wee_debug

tkeffer:
I had to add execute permission to `wee_debug`, and change the first line back to

    #!/usr/bin/env python

to get it to run. Don't know if this means it will no longer work on BSD.

Committed to the `development` branch at commit https://github.com/weewx/weewx/commit/58417a4eb71447573f1f4afadf232ae1de72d751

richterb1:
Yes, it will work I forgot to change the #!/usr/local/bin/python back
when I updated the VERSION.  It appears that when setup.py installs
weewx in /home/weewx/bin, it does that.  

On 2017-03-20 12:09 PM, Tom Keffer wrote:

> I had to add execute permission to wee_debug, and change the first line back to
> 
> #!/usr/bin/env python
> 
> to get it to run. Don't know if this means it will no longer work on BSD. 
> 
> Committed to the development branch at commit 58417a4 [1] 
> 
> --
> Reply to this email directly, view it on GitHub [2], or mute the thread [3].
 

Links:
------
[1]
https://github.com/weewx/weewx/commit/58417a4eb71447573f1f4afadf232ae1de72d751
[2] https://github.com/weewx/weewx/pull/241#issuecomment-287866197
[3]

tkeffer:
> It appears that when setup.py installs weewx in /home/weewx/bin, it does that.

Does what? Change the first line? If it does, that's a bug. It should not change any Python files. Only configuration files.

matthewwall:
tom, i think that is a standard python behavior, not a result of anything we do in setup.py

i've seen it happen on other python projects - when you use setup.py (or anything that uses setup.py to do the install), something in the python plumbing modifies the first line of each script to use whatever python installation you used to install.

richterb1:
Yes, it changes the first line.

I see all the weewx executables on Fedora, which I just installed for testing and FreeBSD all use #!/usr/local/bin/python

I used ./setup.py build and ./setup.py install from the weewx-3.7.0 tarball.

tkeffer:
I'll be darned. I just checked an install I did and, sure enough, it modified the first line of every executable. Never noticed before. I really wish it wouldn't monkey with my files unless I ask it to. 

There may be a hidden switch somewhere to turn this behavior off, but I guess it doesn't harm anything so long as we keep installed files separate from repository files. It does complicate debugging, though.

richterb1:
I'll bet you already found this.

Scripts are files containing Python source code, intended to be started from the command line. Scripts don’t require Distutils to do anything very complicated. The only clever feature is that if the first line of the script starts with #! and contains the word “python”, the Distutils will adjust the first line to refer to the current interpreter location. By default, it is replaced with the current interpreter location. The --executable (or -e) option will allow the interpreter path to be explicitly overridden.

richterb1:
I think this is already in a response to you.  I usually get email right
away but haven't.  I continue my NOOB learning curve with GIT, and
GoogleGroups.  I think I have a handle on updating my repository on
github, I also have a clone of weewx on my FreeBSD system.  

Scripts are files containing Python source code, intended to be started
from the command line. Scripts don't require Distutils to do anything
very complicated. The only clever feature is that if the first line of
the script starts with #! and contains the word "python", the Distutils
will adjust the first line to refer to the current interpreter location.
By default, it is replaced with the current interpreter location. The
--executable (or -e) option will allow the interpreter path to be
explicitly overridden. 

On 2017-03-20 12:27 PM, Tom Keffer wrote:

>> It appears that when setup.py installs weewx in /home/weewx/bin, it does that.
> 
> Does what? Change the first line? If it does, that's a bug. It should not change any Python files. Only configuration files. 
> 
> --
> Reply to this email directly, view it on GitHub [1], or mute the thread [2].
 

Links:
------
[1] https://github.com/weewx/weewx/pull/241#issuecomment-287870938
[2]

PR #252, opened 2017-04-24
added missing mentions of "min_ge"

clmcavaney:
I found that min_ge was missing from a few files.  These changes fix that.


tkeffer:
Thanks. Don't know how that got missed.

Well, I do know: It seems to be missing from the test suites as well. If you've got the interest, perhaps you could add it? Just follow the pattern of one of its mates.

clmcavaney:
Yep - will do.

clmcavaney:
Done. Test file updated - not sure how to test the test files though.

tkeffer:
Just type `make test` from the main weewx directory.

You'll have to supply the "correct" expected results as well.

clmcavaney:
How do I modify the test values to ensure I get something other than a count of 0 for a month?
i.e. at least one minimum greater than 20 degrees C

clmcavaney:
Righto - I think the tests are good and working.

tkeffer:
Thanks, Chris! Nicely done.

I changed the "mild" condition from 65F to 59F, so that there would be some non-zero test values.

PR #258, opened 2017-06-01
Show difference of total daylight in minutes and seconds

tsoodla:
Show difference of total daylight in "human readable" format (e.g., 3 minutes, 5 seconds more than yesterday).

### DIFF
--- a/skins/Seasons/celestial.inc
+++ b/skins/Seasons/celestial.inc
@@ -28,12 +28,12 @@
     #set $seconds %= 3600
     #set $minutes = $seconds//60
     #set $seconds %= 60
-    #if $difference > 0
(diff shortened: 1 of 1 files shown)


tkeffer:
This is a matter of personal preference. Personally, near the solstices when the amount of daylight is not changing much, I don't like the text referring to minutes. 

How about removing the minutes if they are zero?

tsoodla:
Removing the minutes if they are zero makes sense. Fixed in [9ce4a1a](https://github.com/weewx/weewx/pull/258/commits/9ce4a1a445d2adc74422a73b36c45b317ebfe18f)

tkeffer:
Committed, but that's a pretty complicated Cheetah expression! Perhaps you can simplify it a bit...?

PR #263, opened 2017-07-04
Fixed: Barometer calibration changed when setting altimeter value

dl1rf:
The old code changes the barometer calibration while setting new
altimeter value to station.
This code will preserve barometer calibration while setting new
altimeter value to station.

### DIFF
--- a/bin/weewx/cheetahgenerator.py
+++ b/bin/weewx/cheetahgenerator.py
@@ -65,6 +65,7 @@ import configobj
 import Cheetah.Template
 import Cheetah.Filters
 
+import weedb
 import weeutil.weeutil
 import weewx.almanac
 import weewx.reportengine
@@ -476,7 +477,8 @@ class Almanac(SearchList):
         # weather database. The database might not exist, so be prepared for
         # a KeyError exception.
         try:
-            archive = self.generator.db_binder.get_manager()
+            binding = self.generator.skin_dict.get('data_binding', 'wx_binding')
+            archive = self.generator.db_binder.get_manager(binding)
         except (KeyError, weewx.UnknownBinding, weedb.NoDatabaseError):
             pass
         else:
--- a/bin/weewx/drivers/cc3000.py
+++ b/bin/weewx/drivers/cc3000.py
@@ -81,7 +81,7 @@ import weeutil.weeutil
 import weewx.drivers
 
 DRIVER_NAME = 'CC3000'
-DRIVER_VERSION = '0.15'
+DRIVER_VERSION = '0.16'
 
 def loader(config_dict, engine):
     return CC3000Driver(**config_dict[DRIVER_NAME])
@@ -207,6 +207,7 @@ class CC3000Configurator(weewx.drivers.AbstractConfigurator):
             print "channel:", self.driver.station.get_channel()
             print "charger:", self.driver.station.get_charger()
             print "baro:", self.driver.station.get_baro()
+            print "rain:", self.driver.station.get_rain()
         self.driver.closePort()
 
     def clear_memory(self, prompt):
--- a/bin/weewx/drivers/vantage.py
+++ b/bin/weewx/drivers/vantage.py
@@ -21,7 +21,7 @@ import weewx.units
 import weewx.engine
 
 DRIVER_NAME = 'Vantage'
-DRIVER_VERSION = '3.0.10'
+DRIVER_VERSION = '3.0.10.8'
 
 def loader(config_dict, engine):
     return VantageService(engine, config_dict)
@@ -819,6 +819,26 @@ class Vantage(weewx.drivers.AbstractDevice):
         self.port.send_data("EEBWR 14 02\n")
         self.port.send_data_with_crc16(struct.pack("<h", offset))
 
+    def setWindCupType(self, new_wind_cup_code):
+        """Set the wind cup type.
+
+        new_windCup_code: The new wind cup type. Must be one of 0 or 1
+        """
+        if new_wind_cup_code not in (0, 1):
+            raise weewx.ViolatedPrecondition("Invalid wind cup code %d" % new_wind_cup_code)
+        old_setup_bits = self._getEEPROM_value(0x2B)[0]
+        new_setup_bits = (old_setup_bits & 0xF7) | (new_wind_cup_code << 3)
+
+        # Tell the console to put one byte in hex location 0x2B
+        self.port.send_data("EEBWR 2B 01\n")
+        # Follow it up with the data:
+        self.port.send_data_with_crc16(chr(new_setup_bits), max_tries=1)
+        # Then call NEWSETUP to get it to stick:
+        self.port.send_data("NEWSETUP\n")
+
+        self._setup()
+        syslog.syslog(syslog.LOG_NOTICE, "vantage: Wind cup type set to %d (%s)" % (self.wind_cup_type, self.wind_cup_size))
+
     def setBucketType(self, new_bucket_code):
         """Set the rain bucket type.
         
@@ -891,7 +911,7 @@ class Vantage(weewx.drivers.AbstractDevice):
     def setLongitude(self, longitude_dg):
         """Set the stations longitude.
 
-        latitude_dg: Must be in the closed range -180.0...180.0
+        longitude_dg: Must be in the closed range -180.0...180.0
         """
         longitude = int(round((longitude_dg * 10), 0))
         if longitude not in range(-1800, 1801):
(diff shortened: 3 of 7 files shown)


tkeffer:
Thanks for your contribution! Unfortunately, I'm away from my weather station for a few weeks, so I won't have a chance to test it. I'll get to it when I get back.

Remind me if necessary!

dl1rf:
No problem. I am only offering.
I will provide more over the next weeks.
I run an own version of the vantage driver with more than 25% of code changes.
This has grown over the past one and a half year.
Separating the changes into isolated topics will take some time, as I will always verify the changes before offering them.

dl1rf:
Hi Tom,

I stopped providing changes to my master and switched over to a dl1rf-development branch with more changes.
As I am new to Git-hub and learning, I made some mistakes, which I tried to correct.
Also I missed to update the hardware.html file at the beginning. Sorry.
But at the end everything provided here shall work.

tkeffer:
I am still away from home, so I have not had a chance to review the proposed changes.

Frankly, there are a *lot* --- hundreds of lines. This driver has been working reliably for years, so am very reluctant to make big changes.

It would help me if you could explain what your main goals are with these changes. You have gone far beyond just preserving barometer calibrations.

dl1rf:
Tom, I don't want to frighten you.

Be sure I understand that you are very reluctant to make big changes to the vantage driver.

The main running code is not changed, except one or two small changes fixing crashes found with my full modified driver version.

My offer are new commands which I use with wee_device program to setup the Weather Envoy and some changes in the layout of the --info command output. This is reflected in the hardware.htm document, so hopefully easy to look at.

My intention is to provide commands I found missing and make the commands and the info usable with the in Europe usual weather units.
I am not so familiar with 'feet', '°F' or 'inHg' but much more with 'meter', '°C' and 'hPa'.
With my full modified driver (these changes are not provided yet) I can use all setup commands not only with the US units, but with all suitable units.

For example: with my full modified vantage driver I can use 'feet' or 'meter' with the --set-altitude command to setup the stations altitude. And the --info command show all values with the units used at the console.
Why shall I manually calculate between weather units while software can do this?
Maybe this will be interesting to others too. Let me know.

With the last commits I tried to fix the open TODO of the vantage driver.
This is an idea how to do, but may need discussion as I only have verified this with my station.

If you need more info to decide to take the pull request or parts of it, let me know.

tkeffer:
Back home.

I would agree that if the user changes the altitude, the barometer calibration constant should not change. That would mimic the behavior if the altitude was changed using the buttons on the console. So, I've included commit 1bddc4e28f5f76ebe68c0562e60f49701e038409.

As for the others, it will take a while to look through them all.

tkeffer:
I've had a change to look a bit more closely at the changes. 

Command `set-transmitter-type`: please move the field `REPEATER` to the end of the list. If it is missing, it should default to zero (i.e., none). This will allow backwards compatibility.

Command `set-altitude-unit`: this is a clunky solution. Can we find a better way to do this? How about adding a unit specifier to the command? This is what `setup.py` does so users are familiar with this pattern. So, the command becomes `--set-altitude=700,foot`. If the unit specifier is left off, it defaults to `foot`, thus maintaining backwards compatibility. 

Command `set-barometer-unit`: same thing.

Command `set-temperature-unit`: same thing.

The command `set-log-average` is confusing. Perhaps change the help field to "Specify `on` to log average temperatures, or `off` to log value at end of archive interval." Or, better yet, change the command to `set-temperature-logging` and offer options `AVERAGE` or `LAST`. This way, it explicitly acknowledges that the user is changing temperature logging. Incidentally, there's a typo in the code comments. You're not changing hex location `0x2B`, but location `0xFFC`.

Despite their names, commands `set-latitude` and `set-longitude` require tenths of latitude and tenths of longitude, making it easy for users to make a mistake. There's no reason not to allow the user to specify a floating number. E.g., `--set-latitude=32.5`.

Finally, and very importantly, the document `hardware.htm` will need to be updated to reflect your changes.

These are very useful contributions for those users with an Envoy. Thanks for your work!

dl1rf:
Tom, do you know how to remove all my additional commits from this pull request?
I wish to present them more clean and clearly, step by step in separate pull requests.
I missed that my commits to my master branch where all added to my first pull request.
This was not intended.

So next I will look to the '--set-transmitter-type' command and include your recommendations and provide it newly clean and clear with the documentation in hardware.htm.

Providing again more clean and clearly is no problem to me, as I have all the steps in my own local source control now and can merge them here again, if I am able to rollback here.

dl1rf:
> Despite their names, commands set-latitude and set-longitude require tenths of latitude and tenths of longitude, making it easy for users to make a mistake. There's no reason not to allow the user to specify a floating number. E.g., --set-latitude=32.5.

Hopefully I did not make a mistake here, but the code and the documentation in 'hardware.htm' shall show that the input is float and with tenths of degree.
I use a round function in the code with 0 but multiply by ten before.

dl1rf:
> Command set-altitude-unit: this is a clunky solution. Can we find a better way to do this? How about adding a unit specifier to the command? This is what setup.py does so users are familiar with this pattern. So, the command becomes --set-altitude=700,foot. If the unit specifier is left off, it defaults to foot, thus maintaining backwards compatibility.

The intention behind 'unit' commands is first to have the possibility to set the console display units even at an Envoy. With the standard console I did this at my console. With Envoy this seems to be useless, but necessary to my next intention.

The second intention is to use these settings to provide the the displayed data at E.g. '--info' command.
I wish the output is not ever the actual showed US units but I wish to use the stations setup (Console/Envoy) units. Thats the units I am familiar to.
As a German I am more familiar with 'hPa' or 'mb' than 'inHg'.

The third (not provided yet) intention is to use the stations setup units as the default for the setting commands, if no additional unit is provided.

E.g: if stations altimeter setup unit is 'meter' than the default unit for --set-altimeter will be meter.
So the command input for my station will be: --set-altimeter=480
To use foot(feet) the command shall be entered as --set-altimeter=1575,foot(or feet??)
Or --set-altimeter=480,meter

Does this make sense to you?

dl1rf:
> The command set-log-average is confusing. Perhaps change the help field to "Specify on to log average temperatures, or off to log value at end of archive interval." Or, better yet, change the command to set-temperature-logging and offer options AVERAGE or LAST. This way, it explicitly acknowledges that the user is changing temperature logging. Incidentally, there's a typo in the code comments. You're not changing hex location 0x2B, but location 0xFFC.

You are right.
your proposal for the command is much better.
I will change take this and provide again.

tkeffer:
I misunderstood your intention with the `--set-altitude-unit` command. How about `--set-altitude-display=meter`?

The unit used for `--set-altitude` should not be dependent on external state, in this case, the state of the console. This makes it difficult to debug. Better to use a consistent default unit, or to explicitly specify a unit. This way the user can show us what command line s/he used, and we know what the command should have done without knowing anything about the console.

Thanks for your work!

dl1rf:
> The unit used for --set-altitude should not be dependent on external state, in this case, the state of the console. This makes it difficult to debug. Better to use a consistent default unit, or to explicitly specify a unit. This way the user can show us what command line s/he used, and we know what the command should have done without knowing anything about the console.

If output from `--info` is included in a debug log the used units are shown there.
But I understand your concerns. It is desired to also use explicit units for overwrite.

As this feature is not yet included in this code I would propose to postpone it to later.

dl1rf:
> Command `set-transmitter-type:` please move the field REPEATER to the end of the list. If it is missing, it should default to zero (i.e., none). This will allow backwards compatibility.

Hm, placing the REPEATER to the end of the list will make it difficult, as the list length is dynamic.
That's the reason why I placed it at the second place in the list.
But if I change the REPEATER input value from numbers 0..8 to letters N,A..H
I think that will work.

tkeffer:
> Hm, placing the REPEATER to the end of the list will make it difficult, as the list length is dynamic.

That's right, but the list length can be determined by using `len`, or you can just subscript the list, but be prepared for an `IndexError` exception.

dl1rf:
The function splitting the input string into the list needs to be changed too.

dl1rf:
Will be an interesting task.

tkeffer:
The pattern for detecting a variable length list is already there. It just needs to be extended one more element.

dl1rf:
Yes, and it need to be of type `char` instead of `int`, as the rest.

dl1rf:
Another question: All the stuff I presented here will be hard to separate if you intend not to include all.
I was to hot and euphoric with presenting the code. As cooled down I wish to revert / throw away everything here except the first accepted commit, close this pull request and than start again slowly and in separate pull requests with clear topics.
Doing so is not that much work to me. I have good tools helping with this.
I don't want to make you to much work trying to understand all changes, as some new functions are interacting with others.
I wish to present that more clear.
How are your thoughts about this?

tkeffer:
It is definitely easier to deal with several small PRs, instead of one big one.

But, the git `cherry-pick` command makes it possible to pull in individual commits, so it really doesn't matter provided that there are identifiable boundaries between the different fixes and additions.

The worst is when they are all jumbled together in one big commit. That makes it almost impossible to untangle.

dl1rf:
I restart again with my commits.
First will be two small fixes.

tkeffer:
Good idea. It will be much easier to review if they come in as separate PRs. It may be a few days before I can get to them --- I'm off to watch the 2017 eclipse.

Thanks for your persistence!

PR #265, opened 2017-08-03
Fix minor alignment issue

tsoodla:
Fix minor alignment issue in titlebar


matthewwall:
these changes have been incorporated into the newskin branch

PR #268, opened 2017-08-18
My first small fixes and changes to the vantage driver

dl1rf:
This are the first small 6 commits after reverting all unacknowledged changes.

### DIFF
--- a/bin/weewx/cheetahgenerator.py
+++ b/bin/weewx/cheetahgenerator.py
@@ -65,6 +65,7 @@ import configobj
 import Cheetah.Template
 import Cheetah.Filters
 
+import weedb
 import weeutil.weeutil
 import weewx.almanac
 import weewx.reportengine
@@ -476,7 +477,8 @@ class Almanac(SearchList):
         # weather database. The database might not exist, so be prepared for
         # a KeyError exception.
         try:
-            archive = self.generator.db_binder.get_manager()
+            binding = self.generator.skin_dict.get('data_binding', 'wx_binding')
+            archive = self.generator.db_binder.get_manager(binding)
         except (KeyError, weewx.UnknownBinding, weedb.NoDatabaseError):
             pass
         else:
--- a/bin/weewx/drivers/cc3000.py
+++ b/bin/weewx/drivers/cc3000.py
@@ -81,7 +81,7 @@ import weeutil.weeutil
 import weewx.drivers
 
 DRIVER_NAME = 'CC3000'
-DRIVER_VERSION = '0.15'
+DRIVER_VERSION = '0.16'
 
 def loader(config_dict, engine):
     return CC3000Driver(**config_dict[DRIVER_NAME])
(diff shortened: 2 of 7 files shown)


dl1rf:
It starts at: BARDATA's humidity ....

tkeffer:
One comment: in general, changes to the "master" branch are "fixes," while new features are put in the "development" branch. These changes are all fixes, so the master branch is appropriate.

But, some of your other changes are new features, so they should be put in the development branch.

dl1rf:
So pull request shall go to your development branch, right?
I will branch my forked master to my development too.

tkeffer:
> So pull request shall go to your development branch, right?

If they involve new features, then, yes. If they are fixes, then they can go to the `master` branch. 

I just wrote up some comments about the weewx git work flow. See commit 31b34eb3d219405bf1b92c56a82e533f8cdc1d7b.

dl1rf:
I will do so.
Actual I have problems with GitHub Desktop, as it shows me two weewx/master branches with different dates. I don't know what happened. The diagram shows that one weewx/master is branched from the other.
Before coming up with the other changes I will read more documents about how to use Git here.
Also I have to make some security changes to the `--set-transmitter` command. The repeater function is only available with the model type 2. The old model are using the field in a different way and we shall keep this. I don't know if anyone ever uses weewx with model type 1 as the code is fixed to model type 2.
I have developed a fix for the open TODO and with the changes the vantage driver shall be able to detect between model type 1 and 2. The code analyses the receiving channel 0 for this.

tkeffer:
Do you know when the type 2 model was introduced?

There is a similar situation with archive records. Firmware before 2002 use a "Rev A" format, afterwards a "Rev B" format. In the 8 years weewx has been public, I've only encountered a handful of "Rev A" stations.

If type 2 was introduced at the same time, I doubt there's many still out there.

dl1rf:
April 24, 2002 and later use Rev B format.

dl1rf:
The model 1 can be detected by the ID 1 transmitter type having a value below 0x8x and not 0x0x.
So model 1 ID1 transmitter type is 0x10 and the model 2 ID1 transmitter type is 0x00 or >= 0x80.
With model 2 the ID 1 can't be disabled, set to 0x0A. Trying to do so will not be accepted by the station. Maybe this is true for model 1 also. Otherwise ID 1 will be 0x0A wich is not 0x00. 
If I read it right than there are also model 1 stations build using archive "Rev B" format.
So I think the only way to distinguish between model 1 and 2 two is by checking the ID 1 transmitter type field. (upper nibble)
It's really not clearly named in the documentation of the serial protocol rev. 2.6.1 page 46.

dl1rf:
I will open a separate branch within my fork.
Than you can preview my solution of model detection.
This modification is independent.

dl1rf:
You can view my solution at: https://github.com/dl1rf/weewx/commit/20ec60a366bd88ce5cfca4b8e12866168d7e08e8
branch:
https://github.com/dl1rf/weewx/tree/vantage-TODO-fix

PR #270, opened 2017-08-25
Allow user to specify the Vantage Pro model type in weewx.conf

dl1rf:
Allow user to specify the Vantage Pro model type in `weewx.conf`.
Only allowed values are `1` for Vantage Pro and `2` for Vantage Pro2.
Vantage Vue will always overwrite `model_type` with value `2.`
If `model_type` is not defined in `weewx.conf` the default will be `2`, as it is hard coded now.
Invalid specified values in `weewx.conf` will raise an exception and stop loading the driver.
Addition:
`_determine_hardware()` function name corrected.
Duplicate `docstring` removed.
Call `_determine_hardware()` only once in `_setup()`

### DIFF
--- a/bin/weewx/drivers/vantage.py
+++ b/bin/weewx/drivers/vantage.py
@@ -455,22 +455,18 @@ class Vantage(weewx.drivers.AbstractDevice):
         Default is 4]
         
         iss_id: The station number of the ISS [Optional. Default is 1]
-        
-        model_type: Vantage Pro model type. 1 := Vantage Pro; 2 := Vantage Pro2
-        [Optional. Default is 2]
         """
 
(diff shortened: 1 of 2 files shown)


dl1rf:
I have reverted all `model_type` related code.
The `model_type` is useless.

dl1rf:
I will redo the `Allow user to specify old Vantage Pro model type in weewx.conf`.
I don't need it for my functions.
But while looking about the use of `model_type` I found that it is used in `_rxcheck()`.
I think it will make sense to complete the weewx vantage driver with the possibility to allow the user to specify the Vantage Pro model.
It will be optional to weewx.conf. Default will be 2 as hardcoded now.

tkeffer:
Thank you!

PR #271, opened 2017-08-30
Added repeater support to '-—set-transmitter-type' command and new command: '—-set-retransmit'

dl1rf:


tkeffer:
As a matter of style, instead of

```python
            if channel != 0:
                print "Proceeding will set retransmit to 'ON' at channel: %d." % channel
            else:
                print "Proceeding will set retransmit to 'OFF'."
```

we generally use

```python
            if channel:
                print "Proceeding will set retransmit to 'ON' at channel: %d." % channel
            else:
                print "Proceeding will set retransmit to 'OFF'."
```

The former is more a "C" style.

dl1rf:
Hm, ok, here we get into general coding rules.
I choose `if channel != 0` while channel is not a boolean.
With boolean I use the same as you generally use.
If you wish I will change this, as I wish to follow your coding style.

tkeffer:
No. It's not important to me.

We generally try to follow the PEP 8 coding conventions, but are not slavish about it. There are lots of exceptions.

See the section [Programming Recommendations](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) of PEP 8, down at the bottom. It's a slightly different case than what you're doing, but gives you an idea of "Pythonic" style.

dl1rf:
Ok, I will read it again and will follow more carefully.

PR #272, opened 2017-09-01
New commands: '--set-latitude', '--set-longitude', '--set-wind-cup', '--set-temperature-logging'

dl1rf:
Additional: very small cosmetic changes.


dl1rf:
Please check if the introduced `°` (\xc2\xb0) character in `--info` command will be possible at all platforms.
I only check with my Raspberry Pie Linux system.

dl1rf:
The next I have are a restructure to the `--info` command and adding barometer calibration offset setting to `--set-offset=VARIABLE,OFFSET` command.
I think these changes will rise a discussion.
I intend to create two additional branches and pull requests to allow independent discussion and maybe ignore the changes.
Is this ok for you?

tkeffer:
Thanks again!

Issue #106, opened 2016-02-23
Strategy to discourage old CWOP posts


tkeffer:
Because CWOP timestamps incoming records with the wall clock time, any catchup will result in a bunch of older records looking like new records.

Either change `stale` to something smaller, maybe 60 seconds, or not allow old records at all.

matthewwall:
set stale to 60 seconds. fixed at 2091be1..f935d2b

jcea:
I think that 60 seconds is overkill. 1800 was OK, 900 could be appropriate. 60 seconds is unusable in my situation :-). I have added a explicit "stale=900" in my configuration to cope with this change.

Related: PR #277

PR #277, opened 2017-09-30
Default for CWOP "stale" was too small. Increase it to something reasonable

jcea:
See issue #106.


tkeffer:
> I think that 60 seconds is overkill. 1800 was OK, 900 could be appropriate. 60 seconds is unusable in my situation :-). I have added a explicit "stale=900" in my configuration to cope with this change.

I tend to agree. Want to amend your PR to change the default to, say, 600 seconds?

jcea:
Done, @tkeffer . Thanks.

PR #299, opened 2018-01-19
Fixes #298

### LINKED ISSUE #298: update documentation for WOW upload
I'll submit a pull req for the userguide, but essentially the 'station' should be set to the 9 digit (pre 2016 registrations) or UUID (post 2016) visible in the URL of metoffice 'my sites', and the password is NOT your metoffice/wow login, but a user created 6 digit numerical pin

Elwell:
This should clarify the docs slightly


tkeffer:
Thanks, Elwell

PR #303, opened 2018-03-01
Maintain wind direction when speed is zero

DigitalDan05:
Even if ignore_zero_wind was set to False, the wind direction
information could get set to 90º because the x and y sum of the
wind vector would be 0.  This changes the accumulator to
return the last known direction when the sum is otherwise zero.

My anemometer isn't mounted in the best location, but I'd like to be able to tell the difference from when the wind vane isn't changing direction at all and when it is moving, but the wind is too gentle to
register on the anemometer.  I'm using an Ambient Weather ws-1401-ip.

### DIFF
--- a/bin/weewx/accum.py
+++ b/bin/weewx/accum.py
@@ -204,11 +204,15 @@ class VecStats(object):
  
     @property
     def vec_dir(self):
-        if self.dirsumtime:
+        if self.dirsumtime and self.ysum != 0 and self.xsum != 0:
             _result = 90.0 - math.degrees(math.atan2(self.ysum, self.xsum))
             if _result < 0.0:
                 _result += 360.0
             return _result
+        elif self.last:
+            # Return the last known direction when our vector sum is 0
+            return self.last[1]
+
 
 #===============================================================================
 #                             Class Accum
--- a/bin/weewx/accum.py
+++ b/bin/weewx/accum.py
@@ -204,7 +204,7 @@ class VecStats(object):
  
     @property
     def vec_dir(self):
-        if self.dirsumtime and self.ysum != 0 and self.xsum != 0:
+        if self.dirsumtime and (self.ysum or self.xsum):
             _result = 90.0 - math.degrees(math.atan2(self.ysum, self.xsum))
             if _result < 0.0:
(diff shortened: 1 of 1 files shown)


tkeffer:
This seems reasonable, although I could certainly be missing something.

However, there is at least one corner case that you missed. Your test

```python
 elif self.last:
    # Return the last known direction when our vector sum is 0
    return self.last[1]
```

will not return anything when the last direction really was zero. You probably want

```python
elif self.last is not None:
    return self.last[1]
```

Thinking through the branches of the conditional, I believe simply returning the last value would be equivalent:

```python
return self.last[1]
```

DigitalDan05:
Thanks for responding quickly. 

When the last direction really is zero, shouldn’t self.last be (0,0)?

In that case, it should still take the elif branch because this prints True for me in a Python interpreter. 
`print True if (0,0) else False`

DigitalDan05:
And if I just return `self.last[1]` without checking to see if it isn’t None, won’t that throw an exception when it is None?  Or am I misunderstanding your last comment?

tkeffer:
True enough. The expression `bool((x, y))` will always return `True` irregardless of the values of `x` and `y` because it tests length. 

I don't think there would be a problem with simply returning `self.last[1]`, even if it is `None`. After all, that was the behavior before your modifications.

tkeffer:
Sorry, I'm sitting at home with a case of the flu, and I'm just not thinking clearly. 

Shouldn't your conditional

```python
 if self.dirsumtime and self.ysum != 0 and self.xsum != 0:
```

be

```python
 if self.dirsumtime and (self.ysum != 0 or self.xsum != 0):
```

which can be simplified further to

```python
 if self.dirsumtime and (self.ysum or self.xsum):
```

DigitalDan05:
You are correct about my logic case, that should have been an "or".

DigitalDan05:
Also, I now see my mistake about self.last.  I thought it was getting initialized to None, instead of (None, None).

tkeffer:
Merged commit https://github.com/weewx/weewx/commit/b3e5b80d7e462dd8df870e4aa4849e9625b35e77

PR #304, opened 2018-03-04
Added windgustdir to WU Rapidfire, rtfreq autoconfig, SMF file for solaroid OS

whorfin:
I noticed when switching to the new AcuRite Hub that it sent windgustdir messages to WU rapidfire
I'm testing a Vantage and noticed weewx was not sending windgustdir to WU rapidfire, hence this PR
wunderground doesn't always show windgustdir i've noticed, but still why not send it, since it's supported?
It seems development is the right branch as it's hard to argue this is a bug exactly, more new feature
Also automatically determine rtfreq, which interestingly i'm noticing is more like 2.0 for VP2 with USB

[http://wiki.wunderground.com/index.php/PWS_-_Upload_Protocol](The WU Docs) say

> where rtrfreq is the frequency of updates in seconds.

After being invited to fix this and ultimately getting shot down in #128, I'm gonna try again...

Finally, added an SMF file for Solaris and Illumos OS svcadm support

...while i've got the ears of devs, i also thought i'd note that weewx works great under LXSS [unix service on Windows]

Cheers


tkeffer:
OK, you're wearing me down.

However, I do have a few requests:

None of the `collections` classes, including `deque`, are used anywhere else in weewx. As you are not using its high-performance aspect, could we just use a simple list? This will help keep memory requirements down.

The name `rtfreq_sma_window` sound very techy, but doesn't convey much information. Just call it `rtfreq_window`.

It should not appear in `weewx.conf`, as most users will not have to tune it. It should just appear in the documentation. 

Use of `assert` to check options is non-standard. The rest of weewx raises an exception of type `weewx.ViolatedPrecondition` with an appropriate error message.

Instead of 

```python
_record['rtfreq'] = str(round(self.sma(tick - self.prevtick), 1))
```

use

```python
_record['rtfreq'] = "%.1f" % self.sma(tick-self.prevtick)
```

which is a more direct way of expressing what you want: a string with a single decimal point.

I also question whether the nested class `SimpleMovingAverage` is necessary as you are not taking advantage of closure, but I don't feel strongly about the subject.

Glad to have the Solaris code (although we do not support Solaris).

Cheers

whorfin:
Heh
I don't mean to be a pain, and in fact I went back and read your objections again.
Since I agree we don't actually know what rtfreq is for, and since clearly WU could just compute the timestamp differences if was meant to be actual as opposed to expected, I'm going to modify the code and this PR to just be a simple "rtfreq" parameter.  So it won't be a hardcoded 2.5, but a default.
stand by...

whorfin:
OK, as requested - this simplifies things.
I also cleaned up realtime to be the int is really is

Cheers

tkeffer:
I definitely prefer the simplicity of this solution.  I've merged the PR, then made a few minor changes, the most significant of which was not including `rtfreq` in weewx.conf. 

The reason is that I wish we could put some guidelines in the User's Guide as to when to set `rtfreq`. What would a user look for to decide that the value should be changed? Is there any observable?

Until there is, I know from experience we will generate tons of unanswerable support questions.

PR #305, opened 2018-03-06
Expand on 'Customizing units and unit groups' section

glennmckechnie:
Following on from this discussion...
https://groups.google.com/d/msg/weewx-development/FwFtnUeMOkQ/UlSnU-3JAgAJ

Additional notes to include the above solution that bypasses the need to add new units directly to user/extension.py  It outlines the option to add the required tags directly to the new service, driver, SLE script.

typos fixed, passes https://validator.w3.org/


tkeffer:
Thanks, Glenn. I simplified it a bit before merging.

PR #307, opened 2018-03-17
Wind direction filter for fousb.py

daffhunt:
I've been using this modification for a year or so, and thought I would offer it back to the project.  Its purpose is to smooth out the readings from the physically very twitchy wind vane on my Fine Offset derived station.  I am attaching an image captured from wunderground that shows the effect; without averaging, my wind direction is all but useless.

Maybe it should be refactored to be outside the driver, but I don't have the time to do that.  Also I don't know how applicable this is to other makes of station, and don't have access to other stations to test with.

windaveraging

### DIFF
--- a/bin/weewx/drivers/fousb.py
+++ b/bin/weewx/drivers/fousb.py
@@ -215,6 +215,7 @@ import sys
 import syslog
 import time
 import usb
+import math
 
 import weewx.drivers
 import weewx.wxformulas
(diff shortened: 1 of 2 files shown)


tkeffer:
Thanks, Steve

As you may have guessed, we try not to put this kind of functionality inside a driver. The intention is to keep them as stateless as possible. Instead, this should be implemented as an optional service.

See the [Guidelines](http://weewx.com/docs/customizing.htm#General_guidelines) on writing a driver.

daffhunt:
Yes, that seems reasonable.  When I get a chance I'll try moving it into a service, and remove the device specific assumptions like quantization to 16 compass points.

Issue #318, opened 2018-04-24
Adding week number to ToDate filenames


stimpy23:
Hey, everybody!
Thanks for your great work! I'm a long time happy user... :)

Now I have question: Could you please add the ability to include week numbers in the name of the "ToDate-Files"..?

Like "NOAA-YYYY-MM" is translating to "NOAA-2018-04" but with weeks: "NOAA-WW" to "NOAA-17"
I'm working on a template that needs this feature, so I've patched it into my local weewx version, but it'ld be awesome not having to in the future...

My patch is attached and if you want to see what I need it for, check out https://www.rabenwetter.de/archive.html

Thanks in advance and cheers,
  stimpy

[cheetahgenerator.patch.txt](https://github.com/weewx/weewx/files/1942666/cheetahgenerator.patch.txt)

stimpy23:
I hope, I've done it all right... This is the pull request, consisting of 2 commits:
https://github.com/weewx/weewx/pull/319

There's a third commit in the conversation timeline that should not exist anymore and it doesn't at the "commits" tab.. So I hope, it's all right...

stimpy23:
All done. Thank you!

PR #319, opened 2018-04-24
Added weeknumber substitution for filenames

stimpy23:
According to my feature request at
https://github.com/weewx/weewx/issues/318

this patch adds the ability to substitute the string "WW" with the week number in filenames of "ToDate archive files", just like it's already doing with "DD", "MM" and "YYYY".


tkeffer:
Looks good!

Attribution added in commit 8da0f9fda9cc7abf15127c7d29b13f2da08618db

PR #324, opened 2018-06-07
Extend imagegenerator.py to output graphs for last calendar year

evilbunny2008:
It'd be nice if this or something like it would be included in weeWX.

### DIFF
--- a/bin/weewx/imagegenerator.py
+++ b/bin/weewx/imagegenerator.py
@@ -73,6 +73,34 @@ class ImageGenerator(weewx.reportengine.ReportGenerator):
                     plotgen_ts = archive.lastGoodStamp()
                     if not plotgen_ts:
                         plotgen_ts = time.time()
+                        
+                # If previous year or month is set, get the setting and use it to calculate the time span. 
+                year = datetime.date.today().year
+                month = datetime.date.today().month
+                previous_month = 0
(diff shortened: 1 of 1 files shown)


tkeffer:
Thanks for your PR!

Something like this seems reasonable, but is there any way to generalize it? I can see requests for, "But I want last month," or, "What about 2 years ago?"

evilbunny2008:
So, something like year=-1 for last year? Same with month=-1?

I've only ever known one calendar so this may be misleading to some.

evilbunny2008:
Updated to use previous_month and previous_year

tkeffer:
That's one way of doing it.

Another would be to adopt the cheetahgenerator model: allow images to be generated "By Day," "By Month," and "By Year."

evilbunny2008:
Wouldn't that add a lot of complexity if all they want is last month/year?

evilbunny2008:
How else can I get last calendar month/year images functionality without a feature like this?

evilbunny2008:
I should have been clearer in my last post, I was hoping for something like SLE type functionality that can work with exacting periods in graphs, because I think it's the wrong thing to do to have last year/month/week/day generated and stored for every time period, but running graphs covering 2 or 3 times the period isn't the best visually either.

PR #333, opened 2018-07-07
Debian package

alteholz:
some improvements for the debian packaging


tkeffer:
Just wanted to let you know that we have not forgotten about you. It's just that Matthew (who maintains the Debian package) is overseas, and I'm in and out of cell phone coverage.

PR #340, opened 2018-09-07
Fix unmatched lines in logwatch

buster-one:
Changed "Garbage" back to lowercase in bin/weewx/engine.py
"engine: Garbage collected" will be reported as unmatched line in logwatch.


tkeffer:
Bit of history here. The log messages in WeeWX have always started with an upper-case letter, but there seems to be a shortage of caps in the Cambridge, MA area, so, Matthew, as he has added things, uses lower-case letters. There is no such shortage here in Oregon, so I swap them back for upper-case.

Bottom line: the right place to fix this is in the logwatch configuration file.

buster-one:
PR updated.

Issue #346, opened 2018-10-04
Inconsistent use of defaults by `wee_config --reconfigure`


gjr80:
When the user runs `wee_config` with the `--reconfigure` action the user is presented with a number of key WeeWX config items eg description, altitude, location etc. In each case the default presented is the existing setting for that config item in `weewx.conf`. The user simply hits return to accept the current setting. Later, when presented with any driver config items the default presented is not the current setting (if it exists) from `weewx.conf`, but rather the default that is hard coded in the driver file. This is not an issue the first time the driver is configured; however, on second an subsequent uses of `wee_config --reconfigure` for driver settings the user cannot simply hit return to accept the current setting without change (unless the current setting is the same as the hard coded default in the driver file). This could lead to the user inadvertently mis-configuring the driver.

gjr80:
Email comments from @matthewwall wall on 2 October 2018:

it would require some changes, but i think it could be done with minimal disruption.

currently it looks like this:
```
    def prompt_for_settings(self):
        settings = dict()
        settings['x'] = self._prompt('x', default_value)
        return settings
```
if i understand correctly what you are asking, we would have to get the current settings into prompt_for_settings.  

here are some ways we could do what you ask:

option 1: pass the current settings to prompt_for_settings.  this would be rather disruptive as it would break every existing ConfEditor implementation that uses prompt_for_settings.

option 2: pass the current settings to the loader_function when the editor is loaded.  this is also disruptive and would break every existing ConfEditor

option 3: assign a copy of the settings as a member after the editor is loaded.  then any conf editor could use those as default value (first try the existing settings, then fallback to hard-coded default if no existing setting is found)

option 4: streamline 3 even more by embedding the default into the _prompt function.  then the behavior you are asking for would happen automatically for every existing and future ConfEditor implementation.

option 4 might look something like this:

config.py line 151:
```
        stn_info.update(weecfg.prompt_for_driver_settings(driver, stn_info)
```
weecfg/__init__.py line 1140:
```
def prompt_for_driver_settings(driver, stn_info):
    ...
    try:
        ...
        editor = loader_function()
        editor.existing_options = stn_info.get(driver_module.DRIVER_NAME, {})
    ...
```

drivers/__init__.py line 133:
```
class AbstractConfEditor(object):
    ...

    def _prompt(self, label, dflt=None, opts=None):
        if label in self.existing_options:
            dflt = self.existing_options[label]
        import weecfg
[...]

PR #347, opened 2018-10-04
Address issue #346

gjr80:
Changes as per Matthew's email of 2 October. One change I found necessary though, `weecfg.prompt_for_driver_settings()` needs a `config_dict` to obtain the driver settings rather than `stn_info`.

Tested against a driver that does prompt for driver settings and one that does not. Both behaved as expected.

PS. Sorry about all the whitespace removal, but that must be a good thing musn't  it?

### DIFF
--- a/bin/weecfg/__init__.py
+++ b/bin/weecfg/__init__.py
@@ -77,30 +77,30 @@ DEFAULT_LOCATIONS = ['../..', '/etc/weewx', '/home/weewx']
 def find_file(file_path=None, args=None, locations=DEFAULT_LOCATIONS,
               file_name='weewx.conf'):
     """Find and return a path to a file, looking in "the usual places."
-    
+
     General strategy:
 
     First, file_path is tried. If not found there, then the first element of
     args is tried.
 
     If those fail, try a path based on where the application is running.
-    
+
     If that fails, then the list of directory locations is searched,
-    looking for a file with file name file_name. 
-    
+    looking for a file with file name file_name.
+
     If after all that, the file still cannot be found, then an IOError
     exception will be raised.
-    
+
     Parameters:
 
     file_path: A candidate path to the file.
 
     args: command-line arguments. If the file cannot be found in file_path,
     then the first element in args will be tried.
-    
+
     locations: A list of directories to be searched. If they do not
     start with a slash ('/'), then they will be treated as relative to
-    this file (bin/weecfg/__init__.py). 
+    this file (bin/weecfg/__init__.py).
     Default is ['../..', '/etc/weewx', '/home/weewx'].
 
     file_name: The name of the file to be found. This is used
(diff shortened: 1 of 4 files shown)


tkeffer:
Looks good, except for one thing. There is also a call to `prompt_for_driver_settings()` in `setup.py`.

gjr80:
Hmm, a flaw in my workflow. Working off a VM with an install from a branch does not give you a `setup.py`...

Fixed but have not tested this. I am not sure I see the need, but then again I am not well versed on `setup.py`

tkeffer:
Looks good. What's with the dickin' around of the whitespace? Did some tool do that?

gjr80:
My editor has a setting to automatically delete trailing whitespace on save. I turned it on after a few files a while ago had so much whitespace it bloated the file size.

tkeffer:
What I (try to) do in that situation is do a commit with just the reformatting, then a commit with the recoding. That way, a reader doesn't have to wade through a bunch of non-relevant stuff to find the code changes.

gjr80:
Ok, will make sure I do it properly next time.

matthewwall:
i have seen this kind of whitespace thrashing before.  we are getting a lot of it in another project i'm working on where developers are using different versions of pycharm, or where they have their pycharm customized differently from everyone else.

some editors will automatically indent for you, but then they forget to remove the whitespace.  other editors (latest pycharm, i think) will flag that extraneous whitespace or just automatically remove it.

sometimes it is the indents, sometimes it is trailing whitespace (often leftover from removing a comment, or from copy-pasting from a shell or other environment into the editor - different editors handle copy-paste in different ways)

anyway, tom's suggestion about doing a single commit with formatting changes is a good one.  in practice, this means you should first open the file you intend to work on in your editor, run its 'lint' (or equivalent), save the file, then `git diff` to see if it tidied up anything.  if so, do the tidy-up commit.  then do your hacking.

tkeffer:
These days, I'm using pycharm almost exclusively. It doesn't touch the whitespace unless you ask to reformat the file (`Ctrl-Alt-L`). 

If we all use the same formatting style, it should work well. Right now, mine is the default. Look in the top level project directory, file `.idea/codeStyles/codeStyleConfig.xml`. Mine reads

```xml
<component name="ProjectCodeStyleConfiguration">
  <state>
    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
  </state>
</component>
```

it turn, the `Default` code style (look in `~/.PyCharm2018.2/config/codestyles/Default.xml`) is

```xml
<code_scheme name="Default" version="173">
  <Python>
    <option name="HANG_CLOSING_BRACKETS" value="true" />
  </Python>
  <codeStyleSettings language="Python">
    <option name="RIGHT_MARGIN" value="120" />
  </codeStyleSettings>
</code_scheme>
```

Matthew would probably prefer something smaller for `RIGHT_MARGIN`, but I figure what's the point of having a modern wide screen display if you can't use it!

PR #352, opened 2018-11-01
Implement $almanac.sidereal_time tag

gjr80:


gjr80:
Within pyephem/epehem `sidereal_time()` is a method of `class Observer` and is called without parameters (refer PyEphem [other Observer methods](https://rhodesmill.org/pyephem/quick.html#other-observer-methods)).

Key change is removing `_get_observer()` method from `class AlmanacBinder` and making it a function that is passed either an `Almanac` or `AlmanacBinder` object.

I have tested the modified `almanac.py` using the doctest (?) examples in `almanac.py` and all continue to work correctly.

tkeffer:
Looks good, although I wonder if this

```python
        # Check to see if it is an observer attribute
        elif attr in ['sidereal_time']:
            # These functions require the time of the observation
            observer = _get_observer(self, self.time_djd)
            # Call the function
            return math.degrees(getattr(observer, attr)())
```
could be simplified to this
```python
        # Check to see if it is an observer attribute
        elif attr =='sidereal_time':
            # These functions require the time of the observation
            observer = _get_observer(self, self.time_djd)
            # Call the function
            return math.degrees(observer.sidereal_time())
```

gjr80:
Yes it could, and I thought about doing that but decided to leave it in a similar format to the other attributes in case there were more to be added. I will leave as is for the moment and have a look as to whether there are any more attributes that may be worthwhile, I think we have everything covered now (`next_rising`, `next_setting` and friends being the main ones and they are covered in `AlmanacBinder`).

tkeffer:
OK. Three things:
1. Add a doctest;
2. Add `sidereal_time` to the documentation;
3. Make a note in the `changes.txt` log.

gjr80:
> OK. Three things:
> 1. Add a doctest;
> 2. Add `sidereal_time` to the documentation;
> 3. Make a note in the `changes.txt` log.

Done, done and done. Could not help myself and put a couple of notes in re RA and sidereal time returning decmial  degrees rather than the customary 0-24 hours.

On a related note I noticed `$almanac` now has a `separation` tag and the tag returns an `ephem.Angle` object. An `ephem.Angle` object stores a floating radian value but formats as a string in `HH:MM:SS.yyy` format. Nothing wrong with this (in fact I like the string formatting) but it is different to all of the other angles that weeWX returns from pyephem. Not saying one thing or another needs to be done but I think there is some inconsistency that needs to be looked at. Also, given the (expected) low user base I don't think it is a critical  or time sensitive issue.

tkeffer:
I did a quick implementation of `separation` last January at a user's request. 

What would be a more consistent thing to return? Decimal degrees?

gjr80:
> What would be a more consistent thing to return? Decimal degrees in a `ValueHelper`?

I think decimal degrees (if the purpose of the data was for calculations I would say radians) as that is what the other ephem/almanac angles use. Not so sure about a `ValueHelper` though as none of the other ephem/almanac angles are returned as `ValueHelpers`. Chances are though that any change to the current would break the original user's user of `separation`. I would like a simple means of displaying `ra` and friends in 24 hour notation but that is a formatting issue not a units issue. Rather than dragging this closed PR on I will give this some thought and come up with a separate issue/PR as required but either way aiming to get a consistent approach in 3.9.0.

tkeffer:
Thanks, Gary! You know more than I do about the likely use cases for the astronomer / enthusiast.

PR #354, opened 2018-11-04
Added support for WMR89

fcauwe:
Added support for WMR89

### DIFF
--- a/bin/weewx/drivers/wmr89.py
+++ b/bin/weewx/drivers/wmr89.py
@@ -247,7 +247,7 @@ class WMR89(weewx.drivers.AbstractDevice):
                    _record = self._sensors_to_fields(_record, self.sensor_map)
                  
                  if _record is not None:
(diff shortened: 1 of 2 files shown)


fcauwe:
Hi,
Is there something unclear with this merge? Should the code be optimized?
François

tkeffer:
Sorry for not getting back to you.

There are two standards for putting something in the core distribution: it has to be easy to support, and the device should be in widespread use. 

Unfortunately, WMR89 doesn't meet either of these criteria: in the past, Oregon Scientific stations have been among the most difficult to support, and the WMR89 is also not very common.

One other issue: there is another [driver for the WMR89](https://github.com/matthewwall/weewx-wmr89), authored by Matthew Wall.

Perhaps the two of you could come up with a single, definitive version, then put a link in the [Wiki](https://github.com/weewx/weewx/wiki#drivers) to it?

PR #355, opened 2018-11-07
Add RP5 RESTful API support

sapegin-o1eg:
Is it possible to add more RESTful services out of the box? I have made a module for publishing reports to weather site rp5 (https://github.com/sapegin-o1eg/weewx-rp5). It would be grateful if the module will be in the WeeWX.

### DIFF
--- a/bin/weewx/restx.py
+++ b/bin/weewx/restx.py
@@ -1760,6 +1760,104 @@ class AWEKASThread(RESTThread):
         return ''
 
 
+# ============================================================================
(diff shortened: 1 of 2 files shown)


tkeffer:
Thanks for your contribution, but there is not enough interest in rp5 to warrant putting it in the code base --- there is not a single reference to it in weewx-user, nor in weewx-development.

I suggest putting a link to your uploader in the [Wiki](https://github.com/weewx/weewx/wiki)

PR #364, opened 2019-01-27
Create PlotlyJSONGenerator ReportGenerator for plotly.js

kevinoid:
`PlotlyJSONGenerator` is a `ReportGenerator` which creates JSON files for [plotly.js](https://plot.ly/javascript/), a free and open source declarative interactive charting library built on [d3.js](https://d3js.org/) and [stack.gl](http://stack.gl/).  It can be used by skins to create interactive graphs/charts.

It is based on `ImageGenerator` and shares a significant amount of code.  These classes could be refactored to remove much of this duplication.  It currently uses the `ImageGenerator` configuration from `skin.conf`, since it serves the same purpose and supports nearly all of the properties, although that could be changed.

I hosted a demo using a modified version of the Seasons skin at https://kevinlocke.name/misc/weewx-demo/seasons-plotly/index.html (and an unmodified version generated from the same data at https://kevinlocke.name/misc/weewx-demo/seasons/index.html for reference).

If this is something which you might be interested in merging into weewx in some form, I would be interested in working towards that goal.  Let me know what you think and how you would like to proceed.

Best,
Kevin

### DIFF
--- a/bin/weewx/plotlygenerator.py
+++ b/bin/weewx/plotlygenerator.py
@@ -1,0 +1,984 @@
+#
+#    Copyright (c) 2009-2015 Tom Keffer <tkeffer@gmail.com>
+#    Copyright (c) 2017-2019 Kevin Locke <kevin@kevinlocke.name>
+#
+#    See the file LICENSE.txt for your full rights.
+#
+"""Generate JSON data files for plotly.js plots for up to an effective date.
+Based on weewx.imagegenerator.ImageGenerator."""
+
+from __future__ import with_statement
+
+import cmath
+from datetime import datetime
+import errno
+import itertools
+import json
+import locale
+import math
+import os.path
+import syslog
+import time
+
+import weeplot.genplot
+import weeplot.utilities
+import weeutil.weeutil
+import weewx.reportengine
+import weewx.units
+from weeutil.weeutil import to_bool, to_int, to_float
+from weewx.units import ValueTuple
+
+# =============================================================================
+#                    Class PlotlyJSONGenerator
+# =============================================================================
+
+class PlotlyJSONGenerator(weewx.reportengine.ReportGenerator):
+    """Class for managing the plotly.js JSON generator."""
+
+    # Map weeWX marker type to most similar plotly marker type
+    _plotly_marker_types = {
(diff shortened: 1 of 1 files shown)


matthewwall:
i am very anxious about adding code to weewx that depends on d3 or plotly or any other graphing tool.  so my initial reaction is that this should be implemented as an extension, and not included in weewx itself.

however, to make that happen we would probably want to do some refactoring.  part of that refactoring would be reduce/eliminate code duplication, so that the PlotlyGenerator derives from ImageGenerator, or has access to the same utilities that ImageGenerator uses.

part of that refactoring would be the work gary has done to provide JSON output generically.  hopefully the PlotlyGenerator could use the API that gary has created in order to emit the JSON (or csv?) required by plotly (or D3 for that matter).

tkeffer:
General comments:

I think this is way too specialized to be included in the code base. It should really be an extension.

Having said that, the right way to do this is by using a combination of Gary's JSON extensions, templates, and search list extensions. I'm not convinced that a specialized report generator is necessary.

In any case, as you note, it shares a lot of code with imagegenerator. If that's the case, it could be a subclass of `ImageGenerator`.

kevinoid:
Thanks for the discussion.  Those are good thoughts.  Maintaining it as an extension makes sense to me.

Assuming Gary's JSON work will provide a way to generate time-bounded JSON data files as part of a skin, it would be possible to implement the conversion from generic JSON to plotly.js JSON format using in-browser JavaScript.  For the plots to resemble the images, the configuration would have to be served and parsed in-browser as well.  I'm not keen on reimplementing `PlotlyJSONGenerator` in JavaScript right now, but future extensions and skins could use that approach.  For the current Python `PlotlyJSONGenerator`, I am not sure how it would be integrated with Gary's JSON work.

Having `PlotlyJSONGenerator` extend `ImageGenerator` (after it is refactored to be extensible) or a shared abstract base class sounds like a great idea to me.  It obviously isn't required for the extension to work, but it would certainly cut down on the amount of code in the extension and the resulting maintenance burden.

Shall we close this PR in favor of maintaining `PlotlyJSONGenerator` as an extension, or is there any reason to keep it open?

PR #365, opened 2019-01-31
Removed what looked like typo

chrisrycx:


matthewwall:
yup, that is leftovers from a sloppy merge.

PR #382, opened 2019-02-28
Implement minsum, minsumtime and sum_le aggregates

gjr80:
Use case for `sum_le` - counting days with no rain or for none weather applications days where power generation is less than a given value.
Use case for `minsum` and `minsumtime` - non-weather applications minimum daily power generation and the date on which it occurred.


tkeffer:
Looks good, although a test would be welcome

gjr80:
Thought I was doing a good job by picking up on the docs. Will get onto the
test when I'm back home next week.

gjr80:
Clouded in this weekend so find my self catching up. Just got my head around the test suite again, how do you want these aggregates tested? I see none of the other (left field) aggregates are tested at the moment, do you want a new table of aggregates added to `StandardTest/index.html` to cover off those that are not currently tested? In for a penny, in for a pound.

tkeffer:
Most are in there. For example `sum_ge`, `max_ge`, `max_le`, `maxsum`, and `maxsumtime` are tested in the template `byyear.txt.tmpl`.

It's not much of a test, but it's something!

PR #383, opened 2019-03-02
Add tests for untested aggregates

gjr80:
Refer PR #382.

Existing `maxmin`, `maxmintime`, `minmax` and `minmaxtime` aggregates are not tested in the current test suite. New aggregates `minsum`, `minsumtime` and `sum_le` need to be added to the test suite.

`minsum` and `minsumtime` tests are nonsense (eg `$month.outTemp.minsum`) but with the periodic simulated data used in the test suite the more sensible (but not much more useful) `$month.rain.minsum` returns 0.0 which whilst deterministic could also result from incorrect operation of `minsum`.

### DIFF
--- a/bin/weewx/test/expected/StandardTest/index.html
+++ b/bin/weewx/test/expected/StandardTest/index.html
@@ -665,6 +665,38 @@
       </tr>
   </table>
     
+  <h1>Test for aggregates</h1>
+  <table>
+      <tr>
+        <td align="right">$month.outTemp.maxmin</td>
+        <td align="left">40.0&#176;F</td>
+      </tr>
+      <tr>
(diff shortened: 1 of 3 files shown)


matthewwall:
you can do stochastic tests in unit tests.  set the random number seed before you invoke any random numbers.  that will make the test repeatable.

gjr80:
Seeding the random number generator will not help with using `minsum` against rain as the issue is the periodicity and magnitude limits of the rain 'generator' function used in the simulator rather than its randomness. Plus I just get nervous where the result of query is 0; is that because there actually was nothing found or because something was found but there was an error and nothing was returned.

Or perhaps I have totally missed your point. I struggle at times with my (infrequent) use of the test suite, though this time as far more successful and productive than the last!

PR #388, opened 2019-03-11
Add Weather Display import module

gjr80:


tkeffer:
I merged your changes into the `python3` branch. Look it over carefully and see if it's what you want.

I had to add a number of Python 3 fixes, as well as a few small errors I found in my limited testing of wee_import.

It wouldn't hurt to run it through whatever test suites you might have. It would be nice to do it under Python 3, but Python 2 would be OK.

gjr80:
Installed the latest `python3` branch (latest commit https://github.com/weewx/weewx/commit/ac93649e9504846eabc5620ad15d7ab5b32fe1d3) on a VM and ran through each of the `wee_import` sources with my test data. All worked fine (did not check imported data for accuracy though) except for WD which threw this:
```
gary@stretch6:/var/tmp$ sudo /home/weewx/bin/wee_import --import-config=/var/tmp/wd.conf --dry-run
Starting wee_import...
Traceback (most recent call last):
  File "/home/weewx/bin/wee_import", line 861, in <module>
    main()
  File "/home/weewx/bin/wee_import", line 810, in main
    wlog)
  File "/home/weewx/bin/weeimport/weeimport.py", line 337, in sourceFactory
    log)
  File "/home/weewx/bin/weeimport/wdimport.py", line 380, in __init__
    _temp.append([p, int(filter(str.isdigit, fn)[:-4]), int(filter(str.isdigit, fn)[-4:])])
TypeError: 'filter' object is not subscriptable
```
Will have a look at that today.

The other issue I noted is that `wee_import` was logging to `syslog` when I have this install setup to log to `weewx.log` (which is working fine for WeeWX). `wee_import` does have some logging functions that are different to the other utilities so maybe there is something going on there that Python 3 does not like or does differently. Again will look at today.

tkeffer:
In Python 3, `filter` is a generator function. It no longer returns a `list`. You can get the old behavior by changing the line to:

```python
    _temp.append([p, int(list(filter(str.isdigit, fn))[:-4]), int(list(filter(str.isdigit, fn))[-4:])])
```

Better yet, restructure the statement to use list comprehension. The filter statements become something like... (NOT TESTED)
```python
[x for x in fn if str.isdigit(x)]
```

gjr80:
> Better yet, restructure the statement to use list comprehension.

Used a list comprehension and join, it's about as good as I will get as not feeling very pythonic today.

Not a syslog expert but adding the `ident` parameter to `.openlog` fixed the syslog issue, can't see the difference to python2 so I don't know why I didn't see the same problem with python2. I always run my installs with logging to `weewx.log`. No matter seems to work correctly now.

PR #396, opened 2019-04-13
Fix to vector average wind speed calculation

timtsm:
The vector average wind speed (vecavg) I see displayed is often higher than the average wind speed, but the triangle inequality implies that the vector sum should always be less than or equal to the straight sum. It looks like weewx is dividing the vector sum components (xsum and ysum) by the time the wind vector was nonzero (dirsumtime), not by the time there is wind data (sumtime, including times when the wind is not blowing). This means that, eg, wind at <0,0> for one minute followed by <20,0> for one minute has a computed vector average of 20 (and a straight average of 10, which is what the vector average should have been too in this case) while wind at <0.1, 0> then <20,0> has a vector average of 10.05 (and straight average of 10.05). This fix just changes weewx to use sumtime instead of dirsumtime for that calculation.


tkeffer:
Sharp eyes, and right you are! Thanks so much, Tim! 

Because this is a bug fix, and not a new feature, I made the commit against the `master` branch. See commit https://github.com/weewx/weewx/commit/e48eeb2f8927319542fdc0628983eae54eb6d92e

PR #399, opened 2019-04-17
fixed windrun calculation to be for current period only

### LINKED ISSUE #250: `StdWXCalculate` calculates `windrun` as a 'day' value rather than a 'per period' value
`StdWXCalculate` can optionally (it is disabled by default) augment each archive record with a calculated `windrun` field. At present the calculated value is the cumulative `windrun` for the day. This means that most `windrun` aggregates (based on either archive or daily summaries) are meaningless eg `$day.windrun.sum` would be expected to give the total `windrun` for the day so far, but it currently returns the sum of the day `windrun` so far for each of today's archive records. `$hour.windrun.sum` would be expected to give the total `windrun` seen so far in the current hour, but it results in the sum of the day `windrun` so far for each of the archive records in the current hour.

`windrun` should operate in a similar way to `rain` where the value is the 'amount since the last value' rather than being day cumulative.

`windrun` is currently only provided for archive records, consideration should be given to including windrun in loop packets; however, this will present additional challenges for stations that emit partial packets.

windcrusader:
I've made some changes to wxservices.py so that calc_windrun only calculates the run for the current archive period. This I believe fixes  #250. The method is somewhat simpler now as it doesn't have to query the db to find out previous windrun values, which I think is the new intent.

Being new to this whole process I'm not quite sure how to test it works. I flicked the code onto my test system and it didn't throw any errors, however, I can see that windrun is not actually in my db schema and I'm not actually sure how to get it there?

Comments and suggestions welcomed.

### DIFF
--- a/bin/weewx/wxservices.py
+++ b/bin/weewx/wxservices.py
@@ -375,32 +375,18 @@ class WXCalculate(object):
             pass
 
     def calc_windrun(self, data, data_type):
-        """Calculate the wind run since the beginning of the day.  Convert to
-        US if necessary since this service operates in US unit system."""
+        """Calculate the wind run for the archive interval. """
         # calculate windrun only for archive packets
         if data_type == 'loop':
             return
-        ets = data['dateTime']
-        sts = weeutil.weeutil.startOfDay(ets)
-        run = 0.0
-        try:
-            dbmanager = self.db_binder.get_manager(self.binding)
-            for row in dbmanager.genSql("SELECT `interval`,windSpeed,usUnits"
-                                        " FROM %s"
-                                        " WHERE dateTime>? AND dateTime<=?" %
-                                        dbmanager.table_name, (sts, ets)):
-                if row and None not in row:
-                    vals_us = weewx.units.to_US({'interval' : row[0],
-                                                 'windSpeed' : row[1],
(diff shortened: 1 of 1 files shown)


tkeffer:
You're close!

Generally, if the necessary data to calculate something are  missing, we don't include the results in the record at all. OTOH, if the data are there, but `None`, then we include `None`. So, you want something like:

```python
if 'windSpeed' in data:
    data['windRun'] = None if data['windSpeed'] is None else data['windSpeed'] * data['interval'] / 60.0
```

The next part is more complicated. The type `windrun` is not in the default `wview` schema, so you cannot use a test that relies on the database. The test suite could be modified so that it not only tests values in the database, but also in the "current" record. Let me do that.

windcrusader:
Thanks for the feedback. I've corrected it per your suggestion.

Not necessarily for this thread, but I'd be keen to learn more about the testing approach if you have a wiki page or something you can point me in the direction of.

tkeffer:
You had a mistake in your logic. Instead of

    if 'windSpeed' is None

it should be 

    if data['windSpeed'] is None

I've fixed it, and the changes have been merged into `master`. Thanks, Brad!

tkeffer:
No Wiki page on testing. You just have to wade through the test suites!

PR #400, opened 2019-04-19
Add timestamp and max_delta to "current" tag

PeterQuinn925:
I gave it a shot. I hope I've understood it well enough.


tkeffer:
Looks great!

PR #403, opened 2019-04-30
wunderfixer socket timeout change to 20 seconds (from 10)

UberEclectic:
Tom, I have been having a barrage of issues with wunderfixer (and restx.py) giving up too quickly on WU.
WU is apparently having major capacity issues.  I reached out to them to see if they're throttling me, and the answer was no.

I found that most of the time a 20 second timeout gets around the WU problem, at least for wunderfixer.
It's a very minor change.

As an aside I can now glean that the reason wunderfixer has any work to do at all is because even the restx.py code defaults to 10 seconds, so that explains why records go missing when WU is overloaded.  I'll leave that one for you to decide on whether to change the timeout over there.

Oh and apologies for all the other "noise" in this pull request.  That's all from before and pertaining to the fact that I just moved the watchdog stuff into a different branch to keep it on the side for now.
BTW, I did make the watchdog more generic by putting the default to notify-only, so anyone could use it for any weather station just to know when weewx hasn't logged any new records in > 10 minutes (default interval, but configurable).  Works even when weewx itself has crashed, not just the station. =D


tkeffer:
I took a slightly different approach, adding an option `--timeout` to the utility. Kept the default at 10, but users can increase this if they think it will help.

A simple commit, rather than yet another commit on top of all those unrelated commits. https://github.com/weewx/weewx/commit/81d4c509f61d3969f1a09688456dcfb6ce082df8

UberEclectic:
Cool, thanks Tom!

Do you think it would make sense to add a parameter to the weewx.conf to govern the timeout used by restx.py as well?  It's hard-coded to 10 seconds, also.

tkeffer:
Unfortunately, socket timeout is a global resource, so you can't set it individually for each uploader. It is set for the whole application. 

Changing the timeout for `weewxd` is already possible. See option [`socket_timeout`](http://weewx.com/docs/usersguide.htm#General).

UberEclectic:
Thanks, Tom!
Doc says socket_timeout default is 20 seconds.
I'm reading here that restx.py is using 10 seconds:

def __init__(self, queue, protocol_name,
                 essentials={},
                 manager_dict=None,
                 post_interval=None, max_backlog=sys.maxint, stale=None,
                 log_success=True, log_failure=True,
                 timeout=10, max_tries=3, retry_wait=5, retry_login=3600,
                 softwaretype="weewx-%s" % weewx.__version__,
                 skip_upload=False):


And later on in the usage section...

          timeout: How long to wait for the server to respond before giving up.
          Default is 10 seconds.

Although, there is a call...

super(AmbientThread, self).__init__(queue,
                                            protocol_name=protocol_name,
                                            essentials=essentials,
                                            manager_dict=manager_dict,
                                            post_interval=post_interval,
                                            max_backlog=max_backlog,
                                            stale=stale,
                                            log_success=log_success,
                                            log_failure=log_failure,
                                            **timeout=timeout**,
                                            max_tries=max_tries,
                                            retry_wait=retry_wait,
                                            retry_login=retry_login,
                                            softwaretype=softwaretype,
                                            skip_upload=skip_upload)

Is that where timeout is being inherited from upstream and ultimately the weewx.conf internal default of 20 if not explicitly set?

Issue #353, opened 2018-11-01
Refactor logging into a single facility


tkeffer:
We use many different logging strategies across the code base. They should be refactored into a single one, then used  consistently.

matthewwall:
here are some use cases to help guide the evaluation of options:

1) log level.  these levels are most commonly used in weewx:

* error - something failed that user should know about
* info - feedback about settings and/or operation
* debug - detailed feedback

additional levels such as fatal (fail that requires shutdown) or warn (a mild error?) are probably not necessary to distinguish?

2) do not repeat if not necessary.

3) lots of logging when setting up, but change to log-only-exceptions after configuration is tuned.  default behavior should be to provide good feedback, but it should be easy to switch to log-exceptions.  example is the `log_success` and `log_failure` pattern for uploaders.  use case is low-write scenarios (e.g., on SD/SATA-DOM applications) and low-disk-space scenarios (e.g., tiny /var/log partition on a router)

4) driver diagnostic logging (e.g., every string of bytes to/from device).  some drivers need multiple levels of diagnostics, e.g., serial-comms vs parsing-of-serial-data vs loop/archive.  the pattern of debug_xxx withing the driver's config stanza has been used in many drivers to provide these distinctions.

5) uploader diagnostic logging.  uploaders need simple log entry to indicate that upload succeeded/failed.  also need ability to do detailed logging, e.g., actual data sent in a packet/url.

6) consistency - use consistent patterns in the log message formats to make it easy to write patterns for logwatch or elastic search.  for example, put entire log message into a single line whenever possible, but if must use multi-line then do it in a way that can be post-processed easily.

7) platform-agnostic.  we should consider using the python `logging` module instead of using syslog.  

8) logging shim.  whether we use `logging` or `syslog`, we will probably need a thin shim to (1) configure the log and (2) to reduce the amount of code that developers need to write in order to make log entries.  the `logdbg`, `loginf`, `logerr` pattern results in compact, easy-to-manage log management, but probably could be done more pythonically.  doing `syslog.syslog(syslog.LOG_ERR, msg)` is too verbose imho.  the python pattern of `logger.info(msg)` has worked well for me in other projects, and it comes with the benefit of getting module/function labels in log messages if you configure the logger correctly.

tkeffer:
In this issue, I had intended only to take on your use case 8. I've made a start in commit https://github.com/weewx/weewx/commit/a930e65c0024b3738084b5dc10a6f422b790a0a3.

[...]

PR #406, opened 2019-05-10
Added Python 2 vs. 3 conditional for logging method

UberEclectic:
This makes the same driver usable with WeeWx 3.9 in addition to 4.0.
I expect at some point this will be removed, but it's helpful right now for testing backward compatibility of Python 3 code running on Python 2.

I chose the Python version as the conditional, because it was way less code than checking the WeeWX version in the import / dev section, early in the driver.

### DIFF
--- a/bin/weewx/drivers/wmr300.py
+++ b/bin/weewx/drivers/wmr300.py
@@ -779,16 +779,16 @@ examples:
 from __future__ import with_statement
 from __future__ import absolute_import
 from __future__ import print_function
+import sys
 import time
 import usb
 
 import weewx.drivers
 import weewx.wxformulas
 from weeutil.weeutil import timestamp_to_string
-from weeutil.log import logdbg, loginf, logerr, logcrt
 
 DRIVER_NAME = 'WMR300'
-DRIVER_VERSION = '0.20'
+DRIVER_VERSION = '0.21'
 
 DEBUG_COMM = 0
 DEBUG_PACKET = 0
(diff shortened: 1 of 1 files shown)


tkeffer:
Because `wmr300.py` is included in the standard distribution, there is no reason to make it usable under WeeWX Version 3 and Version 4. The `development` branch is pure Version 4.

We are transitioning to a standard logging API (see issue #353). That's why `loginfo()` and other functions are being imported from `weeutil.log`.

UberEclectic:
This was done to allow testing of the v4 driver on a v3 installation under Python 2.
I used it to validate a key objective of v4 being based on Python 3, but with backwards compatibility with Python 2.
Can you think of a better way to make these v4 drivers backward compatible with v3?
Perhaps a backport of the new logging mechanism from v4 to v3?
That way most v3 code would run "as is" with the old logging mechanism, but we could try out the newer v4 drivers by dropping them in easily and not needing to modify them to work with v3.

tkeffer:
> Can you think of a better way to make these v4 drivers backward compatible with v3?

There are lots of changes in V4 that are not compatible with V3. It would be impractical to have every driver work with both versions, and no reason to do so.

I know it can be a pain to maintain separate `master` and `development` branches, but I'm not sure what it would mean to have two branches sharing the same code base.

As for testing, the test suites are also maintained separately.

The key is regular "catch up" merges from `master` to `development`.  If these are done regularly (and they have been), then software development mostly proceeds forward in time, always with `development` as the latest-and-greatest.

PR #407, opened 2019-05-14
Corrected wunderfixer use of restx.AmbientThread class

UberEclectic:
Corrected problem with wunderfixer use of restx.AmbientThread class being out of sync with actual restx.AmbientThread parameter requirements.

1.4.1 05/14/2019
Made WunderStation class consistent with restx.AmbientThread parameters

PR #409, opened 2019-05-16
Python2 backward compatibility fixes for wmr300.py; wunderfixer fix to match restx.AmbientThread class

UberEclectic:
Cleaner pull to replace #407.


tkeffer:
Much better and cleaner. Thanks, Leon.

PR #416, opened 2019-05-23
4.0.x wunderfixer Change to new API KEY method of querying historical data

UberEclectic:
Tested on python2 and python3.
Can't help WU's sporadic 404 and 503 errors, but at least the 403 is gone.

### DIFF
--- a/bin/weewx/drivers/wmr300.py
+++ b/bin/weewx/drivers/wmr300.py
@@ -1978,8 +1978,8 @@ if __name__ == '__main__':
     if options.get_history:
         ts = time.time() - 3600 # get last hour of data
         for pkt in stn.genStartupRecords(ts):
-            print(to_sorted_string(pkt))
+            print(to_sorted_string(pkt).encode('utf-8'))
 
     if options.get_current:
         for packet in stn.genLoopPackets():
-            print(to_sorted_string(packet))
+            print(to_sorted_string(packet).encode('utf-8'))
--- a/bin/weewx/restx.py
+++ b/bin/weewx/restx.py
@@ -771,6 +771,7 @@ class AmbientThread(RESTThread):
                  post_indoor_observations=False,
                  protocol_name="Unknown-Ambient",
                  essentials={},
+                 apikey=None,
                  post_interval=None, max_backlog=six.MAXSIZE, stale=None,
                  log_success=True, log_failure=True,
                  timeout=10, max_tries=3, retry_wait=5, retry_login=3600,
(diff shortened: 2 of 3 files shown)


UberEclectic:
Using the new API actually simplified the post-processing of the WU data by quite a lot!
JSON is lovely like that!  =D

Docs on the new API are over here...  This is for people like us who have a PWS, as opposed to other people who have to pay for API usage:

https://docs.google.com/document/d/1eKCnKXI9xnoMGRRzOL1xPCBihNV2rOet08qpE_gArAY/edit
The API Key is generated in the "Member Settings" section of wunderground.com ("My Profile" > "Member Settings" > "API Keys").

This means the WU section of the weewx.conf needs to have an additional line added, a la:

```
    station = KMIXYZ123
    password = 'DontDareTell!'
    apiKey = 1234567890123456789012345678901
```

I went with "apiKey" instead of "apikey" mainly to avoid adding another obfuscation string to wee_debug.  I saw that "apiKey" was already there, so I went with that.

The old wunderfixer is completely broken nowadays with constant 503 errors and IBM has no plans to fix it.   So...  This new version based on the API key is the only workable solution.

The only "hack" I did was in passing of the options.apiKey, which I did because I didn't want to also have to touch restx.py to update the restx.AmbientThread parameters again, because that could likely break lots of other code.

I figure when you're ready for more general re-writes with the WU API KEY in mind, you can roll the apiKey into restx.py and I can update wunderfixer accordingly.

UberEclectic:
Hold this up.  I didn't realize adding apiKey to weewx.conf would cause the engine.py to blow up.
I figured it would ignore keywords that it didn't actively need to parse.  :-/

tkeffer:
This PR is unnecessary. Just do a single PR against the `master` branch, then we'll do a "catch up" commit against the `development` branch.

UberEclectic:
There are some unresolved issues, anyway.
However, the last time I looked at the diffs there were a bunch that were in conflict, because of prior PR delta's you had not accepted.
Hope that is not still the case.

UberEclectic:
Yeah any subsequent PR in my development vs. yours is still a mess, due to unaccepted intermediate changes.

Also. I have maintained 3.9.x vs. 4.x differenced versions of wunderfixer.
If I push the 3.9.x version to master, how will you get the 4.x version if not in a separate pull?

tkeffer:
> If I push the 3.9.x version to master, how will you get the 4.x version if not in a separate pull?

That's part of the "catch up" process. Most likely, the catch up will resolve in conflicts, which will have to be resolved.

Just focus on v3.9.x. Let me worry about the conflicts.

PR #417, opened 2019-05-23
3.9.x wunderfixer Change to new API KEY method of querying historical data

UberEclectic:
**BEFORE MERGING, we need some other volunteer to test this, since I have moved on to 4.0.**

Can't help WU's sporadic 404 and 503 errors, but at least the 403 is gone.


UberEclectic:
Tested and working.

gjr80:
Whilst the forecasting extension is not part of the WeeWX core it is commonly in use (well until WU implemented it's recent API changes) and it uses the config option `api_key` for the WU API key. I suggest that if we are going to specify a WU API key under `[StdRESTful]` we use the same `api_key` config option name or we risk having the same API key listed with two slightly differing names.

UberEclectic:
I can do this.

Regards,
\Leon
--
Leon Shaner :: Dearborn, Michigan (iPhone)

tkeffer:
Update?

UberEclectic:
WU have partly fixed the new API query for stations east of UTC, but no fixes yet for stations west of UTC.
For stations east of UTC the "black hole" date is now yesterday (vs. day-6 ago as per the prior behavior).

I reached out to IBM yesterday with the above info, but no word back.

For now, I am holding onto an updated version of wunderfixer that uses api_key as the weewx.conf property, *and* adds a new option for whether to normalize archive records to match the WU 5-minute intervals (defaulting to True/enabled).

I have opted NOT to include the date compensation code that I wrote for the wunderdates API debug tool, but given how long it is taking IBM to fix these bugs, I could re-think that if there is a consensus that we should make such an accommodation.  The date compensation code is only invoked if there is a date mismatch, so it isn't a terrible thing to have just as safe-guard, I suppose.  :-/  Thoughts?

UberEclectic:
I just realized that the "yesterday" black hole for stations east of UTC isn't real.
It's because my local time is "today" and stations of east of UTC are on "tomorrow" and I am choosing which API based on whether the requested date is "today" or historical.
So when I use the "today" API for those stations, it returns the "tomorrow" data relative to my local time.
To handle this corner case properly, I would need to compute the stations local date, to decide which API to use.
I don't see such a change as necessary, since the real wunderfixer will be running in the same timezone as the station.

Net net, really the one remaining WU API bug in this context is for stations west of UTC, when making a query after UTC has already rolled past midnight into the next day.

tkeffer:
Not really following this, but, FWIW, two things:

1. The rest of weewx assumes that the weewx server is running on the same time zone as the weather station. 

2. I don't like the idea of coming up with a workaround for a WU flaw to fix a workaround for what is another WU flaw. At some point, it becomes their problem.

-tk

tkeffer:
Three comments
1. This PR mixes two very different things: a patch for wunderfixer, and the concept of "watchdog" scripts.
2. Wunderfixer has already been ported to the new WU API. See commit https://github.com/weewx/weewx/commit/32c35ce380b766054b05a7b64d1d82ab113a5d73
3. I've never liked the idea of watchdog scripts, preferring instead to find engineering solutions. I can see you found an unusual situation where this may be difficult. Nevertheless, it's a corner case (specific hardware experiencing a specific hang). I am reluctant to include it in the general distribution, lest it give the impression that this is the way we do things in weewx-land. Why not just include it in the Wiki?

PR #429, opened 2019-06-30
Fix the connection error in wee_import.

bockpl:
Fix the connection error in wee_import.
Unable to open Weather Underground station .... **** HTTP Error 403: Forbidden

### DIFF
--- a/bin/weeimport/wuimport.py
+++ b/bin/weeimport/wuimport.py
@@ -193,7 +193,9 @@ class WUSource(weeimport.Source):
                                                      date_tt[0])
         # hit the WU site, wrap in a try..except so we can catch any errors
         try:
-            _wudata = urllib2.urlopen(_url)
+	    headers = { 'User-Agent' : 'Mozilla/5.0' }
+            req = urllib2.Request(_url, None, headers)
+            _wudata = urllib2.urlopen(req)
         except urllib2.URLError as e:
             self.wlog.printlog(syslog.LOG_ERR,
                           "Unable to open Weather Underground station %s" % self.station_id)


rezin8:
I'm WAY behind on this, but found this PR when searching for a solution to an issue. The reason for the header info is because WU has started blocking requests from CURL or WGET. Throw a Mozilla header or anything else in there and it will get around the issue. I've tested his fix and it actually solves the 403 issue I was getting.

tkeffer:
Still don't understand. WeeWX has never used a Mozilla header, so it's no different than curl or wget, yet it has posted to WU for over 10 years. What's changed? Can you show us a log? Preferably with debug=1

rezin8:
It's not WeeWX that's the source of problem, it's Weather Undergound. This is just a workaround. Let me see if I can run a debug and show it happening.

tkeffer:
> It's not WeeWX that's the source of problem, it's Weather Undergound. This is just a workaround. Let me see if I can run a debug and show it happening.

I get that. I just haven't seen any evidence that this is a problem, not that the solution solves it. 

Thanks in advance for trying to capture the behavior in a log.

rezin8:
The debug=1 doesn't actually show the issue. Here's a snippet from the syslog file. I killed the weewx process, ran the importer without the "fix", and it shows the 403, then ran the importer with the header "fix" and it shows that it ran. 

There's a couple other things out there showing that Akamai is blocking pulls like this, so you have to throw the fake header in there. It doesn't necessarily have to be a Mozilla header, it could be Safari, or anything else to make Akamai think it's coming from a browser. I'm trying to figure out how I could see the actual behavior since it's not showing up in /var/log/syslog... maybe a tcpdump... 

[syslog.txt](https://github.com/weewx/weewx/files/4141998/syslog.txt)

rezin8:
Let's try this syslog file again...
[syslog2.txt](https://github.com/weewx/weewx/files/4142014/syslog2.txt)

gjr80:
Hang on, the WU components of wee_import were radically changed in 4.0. Biggest change was use of the new WU API. I suspect this PR is no longer relevant, at the very least it should be looked at in the context of 4.9 not 3.9.2.

rezin8:
Ah, you may have a point. I just installed directly from apt like a week ago, so no idea why it installed 3.9.2 if the current release is 4.9...

gjr80:
Sorry, 4.0 not 4.9, darn phone. 4.0 is in the process of being released and master was updated a couple of days ago.

tkeffer:
@rezin8 Would you be willing to try the beta of V4.0? It's really quite easy:

1. Download from weewx.com: http://weewx.com/downloads/development_versions/ into a convenient spot.
2. Unpack the tar file. For example, if the file is `weewx-4.0.0b11.tar.gz`:
 
        tar xvf weewx-4.0.0b11.tar.gz

3. Change directory (`cd`) into the resultant subdirectory:

        cd weewx-4.0.0b11.tar.gz

4. Run `wee_import` from the directory, using your old `weewx.conf` and import files:

        ./bin/wee_import --config=/etc/weewx/weewx.conf --import-config=/path/to/import.conf

gjr80:
> @rezin8 Would you be willing to try the beta of V4.0? It's really quite easy:

Don't forget the hidden step 3.5; now that `wee_import` uses the new WU API to obtain WU station data you need to provide a WU API key for `wee_import` to use. That is specified in the import config file, from memory that is the only change to the 3.9.2 WU import config file. The 4.0 example WU import config file should have sufficient prompts for the info required. Failing that the 4.0 Utilities Guide has details on the WU import config file.

rezin8:
Now we're cooking with oil. That seems to work. Granted, I just did a dry run, but the new API definitely works better. 
[syslog3.txt](https://github.com/weewx/weewx/files/4142749/syslog3.txt)

rezin8:
Ok, I started playing around with the WU import and tried pulling data from last year. For December, it would pull 10 records, then say that it couldn't load the config. For October, the same.

Looking at the error, I went to WU (which this station has been connected to for 5 years), and for some reason, they don't have my data from that time. In fact, my data is very sparse going back to 2015. 

So, as it sits, it looks like if WU has the data, the 4.0 import will work, but if they don't have the data, it's going to fail with a confusing message about not being able to load the config. I would recommend expanding the error message to say something like "If the Station ID is correct, WU may not have data for the date you are requesting. Please check your station history." or something along those lines. 

Thanks to both of you for jumping back on this long closed issue! Now, I guess I go back to trying to get the CSV import to work! 

[syslog4.txt](https://github.com/weewx/weewx/files/4142759/syslog4.txt)

tkeffer:
Gary, can you check on the error message?

gjr80:
Having had a quick look now it appears this error (`IOError`) is raised in response to a 204 (no content] response code when hitting the WU API for a days data. `wee_import` defines its own `WeeImportIOError` error when `wee_import` is unable to obtain source data and I think that it is that error that should have been raised instead of `IOError`, in the `wee_import’` context the `IOError` was intended for when a non-source file can’t be read, hence the config file reference. Could change the message passed by the ‘WU’ `IOError`, but it is still going to be pretty general (or long) because there are numerous reasons why WU did not return a response.

As for why the 204 error, don’t know at the moment. WU has always exhibited unpredictable behaviour; for example, when attempting to get a days data sometimes all records are returned, sometime not all records are returned. This was the case under the old daily history query (not part of the API I believe) and related behaviour has been found to exist under the new WU API when obtaining the current days data. 

I need to get some time in front of the PC and play around with the code and WU API to look at this further.

gjr80:
Sat down in front of the PC to play with b11 `wee_import`. Found the following:

- Using my WU API key and a station ID of KOHPOWEL9 the following command resulted in the 'Unable to load config file.' error (as explained previously this indicates a 204 response error indicating the WU provided no response).

`/home/weewx/bin/wee_import --import-config=/var/tmp/wu.conf --from=2019-10-01 --to=2019-10-31 --dry-run`

- Using my WU API key and my station ID the same command imported data successfully (albeit a dry run)

- Using my WU API key and Tom's station ID the same command imported data successfully (albeit a dry run)

I went back to the WU web page and looked at station KOHPOWEL9. If I try to look at the daily data for 1 October 2019 I am told 'No data available'. If I look at 2 October 2019 I am again told 'no data available'. Likewise for the other handful of October 2019 dates I looked at. November 2019 was the same but 1 December 2019 did provide data. I note that previous logs indicate that some October 2019 data does exist, but at present is seems WU is not returning it for some reason.

So what does it mean?

- For one reason or another the WU API is not returning data meaning `wee_import` cannot import what it is not presented. Anecdotally this seems to be typical WU behaviour. Will changing the request header have any effect, unknown as I don't have time to test that just now. I see no evidence to suggest it would make a difference.
- `wee_import` fails hard if it encounters a WU period for which it can find no data. This is probably not an issue for a single day import but is an issue for an extended period import that happens to include a day for which there is no data. 
- A WU API key gives you access to not only your stations data but any stations data (this was the case as well for queries under the old `WXDailyHistory.asp` page). This is useful to know and greatly assists in troubleshooting WU imports.

So what do we (I) need to do? 
- I need to change the WU import module to raise a `WeeImportIOError` exception and not an `IOError` exception. Not a showstopper but the present error message is confusing/just plain wrong. As part of this I will also look at improving the error message (just a simple change to `WeeImportIOError` itself will remove the 'Unable to load config file.' message and replace with the more appropriate 'Unable to load source file.' (Note the second error message would remain untouched).
- I need to rework the WU import module to skip days with no data. This won't fix any WU shortcomings but it will stop a missing days data terminating an extended import. Not a showstopper (its always been there and no complaints yet) but it does need to be fixed sooner rather than later. The workaround is to import shorter periods or one day at a time.

I will raise an issue to track these.

gjr80:
Commit 177bd19a80885f25304f38ad2aba4338e9bce84b fixes the raising of incorrect exception on a 204 response from WU as well as revising the resulting error messages. I believe the error message is brief and points the user to the three most commonly likely causes; station id, date or data.

Issue #499 raised to track the hard fail when a day of no data is encountered.

rezin8:
Thanks for the help guys! Even though I wasn't able to pull my data from WU, I was able to finally work the CSV's I had into a workable format to get them in, so I now have everything going back to 2015 in the database. Thanks again. http://gregledet.net/weather/

PR #440, opened 2019-08-31
Fix weeimport with calc_missing to write required history before calc

sshambar:
wee_import writes in tranches, but when calc_missing is enabled,
some calculated values require history that is not available
until the tranche is commited.

This results in ranges of records with calculated pressure or ET
values of NULL.

This patch adds a new function timespanLimit() to check if records
in the tranche need to be written to the archive before calcMissing
is called for the next imported record.

The new function performs some basic sanity checks to see if
pressure or ET (both of which require historical values) need to
be calculated before triggering an early tranche write...

### DIFF
--- a/bin/weeimport/weeimport.py
+++ b/bin/weeimport/weeimport.py
@@ -940,6 +940,44 @@ class Source(object):
             self.wxcalculate.do_calculations(record, 'archive')
         return record
 
+    def timespanLimit(self, tranche, new_rec):
+        """Since calcMissing uses historical values, this checks if
+        values in tranche need to be written to the database before
+        it's called"""
+        if not (self.calc_missing and len(tranche)):
+            return False
+        # if ET needs to be calculated, all history is needed in database
+        if ('ET' not in new_rec or new_rec['ET'] is None) and \
+           ('radiation' in new_rec and new_rec['radiation'] is not None):
+            return True
+        if 'pressure' in new_rec and new_rec['pressure'] is not None:
+            return False
+        first_ts = tranche[0]['dateTime']
+        last_ts = new_rec['dateTime']
+        # pressure requires temp 12 hours in the past...
+        return last_ts >= first_ts + (12 * 3600 - 1800)
+
+    def saveTranche(self, archive, nrecs, tranche, unique_set):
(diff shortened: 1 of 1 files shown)


gjr80:
I see the problem this PR is attempting to fix and agree that it needs to be fixed. I am concerned though that as the code stands if the tranche does need to be saved before missing obs are calculated then the trance is being saved each time a record is imported, in other words we are losing the the speed benefit gained by saving in tranches (you could easily achieve the same effect by doing the check up front and forcing the tranche size to 1 if needed).

We should be able to refactor the PR code such that if historical obs are required in order to calculate missing obs then we save the tranche once it is full, then calculate the missing obs and finally save the tranche that now includes the missing obs. Will be interesting to see how this stacks up time wise. I don't have the original timing stats emails from when we developed `wee_import` (might be available in a long buried GitHub issue though) but from memory but tranche size made a huge difference in processing times. I guess it will depend on whether saving a tranche twice is faster than using tranche size = 1.

I am presently in the final throws of updating `wee_import` to utilise the new WU API for WU based imports. No need to update this PR, I will add in the code to the development branch `wee_import`.

sshambar:
Sounds like a plan.

The current code attempts to only reduce the tranche to 1 if needed for ET, and otherwise reduces it ~12 hours for pressure; but yes ET calculation will cause significant slowdowns.

I considered refactoring to perform in-place calculations, but I didn't want to include that much "non-import logic" in the high level functions...especially since the required historical values may come from either the tranche or the database, and resolving that correctly required more knowledge of the code than I have at the moment :)

I'll leave it in your capable hands... would you like me to close the pull request?

gjr80:
Hmm, going to have to give this some more thought as there is a fundamental flaw in our process. The `addRecord()` called to save a tranche of records to archive uses and SQL `INSERT` statement to insert the record(s) into the archive table. This will work fine the first time but if we try to 'add' a record with the same timestamp (field `dateTime`) again it will be rejected due to a duplicate primary key error. 

There is a method available to update a record but this method is structured to update one field at a time. I suspect this approach will be very slow.

Will raise an issue to track this with reference back to this PR. Will close this PR.

PR #441, opened 2019-09-04
First cut of python logging in wee_import

gjr80:
Further to our emails I have removed the `--log` option completely. Since the new logging system prepends `wee_import` to all log entries, `--dry-run` has been changed to log as per a non-dry run with the exception of the entries relating to actually saving records (since this does not occur in a dry run).

The Utilities Guide will need updates re removal of `--log` and various screen captures. Will do this once this PR is finalised.

### DIFF
--- a/bin/wee_import
+++ b/bin/wee_import
@@ -815,7 +815,7 @@ def main():
 
         # advise the user we are starting up
         print("Starting wee_import...")
-        log.log(logging.INFO, "Starting wee_import...")
+        log.info("Starting wee_import...")
 
         # If we got this far we must want to import something so get a Source
         # object from our factory and try to import. Be prepared to catch any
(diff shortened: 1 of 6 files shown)


tkeffer:
My only question is about whether the module name will be correct in the logs.

Typically, each module should start with

```python
import logging
log = logging.getLogger(__name__)
```
This means that each module will get its own logger, with a unique name --- the module name.

However, you have only one place with this snippet of code: the module `weeimport.weeimport`, so your program will only have one logger and its name will be `weeimport.weeimport`.  This logger is then passed in to the constructor of an instance of `WeeImportLog`, which is then shared with all modules. So, all modules will show the same name in the log, namely `weeimport.weeimport`.

Unless I've gotten something wrong, which is entirely possible!

What to do about it? I think you're going to have to go back to a more canonical form, where each module has its own logger, whose name is the name of the module.

You might be able to use functions that do what you want. Each function would take the logger as one of its arguments.

gjr80:
So each module will need to be fed or get a config dict as well. 

Another approach would be to do away with the WeeImportLog class altogether, though I expect this will mean more conditionals in the mainstream code. Appreciate that the logging module makes it easy to log to console and log simultaneously though in the case of wee_import we don’t always want to log the same thing to the console and the log.

I will have another look this afternoon.

tkeffer:
Each model needs the boilerplate preamble at the top. No config dict necessary (at least for logging purposes).

```python
import logging
log = logging.getLogger(__name__)
```

Then just use 

```python
log.info("this is an informational message")
```

This would mean doing away with `WeeImportLog` class.

As for what to do about things that should go to the console (presently done with calls to `self.wlog.printlog()`), I can see two choices.
1. Use simple print statements. This is what the other utilities do.
2. Keep `WeeImportLog`, but refactor it so it just handles printing to the console, then continue what you're doing. Renaming it would probably be a good idea.

gjr80:
Perhaps I need to see if I have a proper understanding of the python logging module. In a nutshell:

- log = logging.getLogger(__name__) return a logger instance, in each module these will be different since __name__ is different
- calls to weeutil.logger.setup sets up the logger instances. The config applied is applied to the logger in each module hence the config dict is only required once early on.
- logging from the WeeWX API calls made by utilities such as `wee_import` will still appear under process name 'wee-import' as even though there are getLogger calls in each of the API modules the logger config calls in `wee_import` have set the logger config.

I think my confusion was coming from me thinking there were multiple logger instances being created that each had to be individually configured.

gjr80:
Refactored `wee-import` logging as per your advice. Class `WeeImportLog` has been removed.
PR updated.

tkeffer:
> I think my confusion was coming from me thinking there were multiple logger instances being created that each had to be individually configured.

There are multiple loggers created --- one per module --- BUT, they obey a hierarchy. Unless told otherwise, they inherit properties from the root logger. So, if you set a property on the root logger, say its handler, all child loggers will inherit it. The call to `weeutil.logger.setup()` is setting up the properties of the root logger.

The power of the `logging` package is that child loggers can override what they inherit from the root logger, either through `weewx.conf`, or programmatically. For example, if you want the `wdimport` module to use a different formatting in its log, here's how you would do it

File **wdimport.py**:
```python
# As before
import logging
log = logging.getLogger(__name__)
# Create special formatter
formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
# Tell the logger to use it
log.setFormatter(formatter)
```

Now it will use a special format in the log. All other properties are inherited from the root logger.

You can accomplish the same thing through `weewx.conf`:

```ini
[Logging]
    [[loggers]]
        [[[weeimport.wdimport]]]
          handlers = wdhandler
    
    # Definitions of possible logging destinations
    [[handlers]]
        [[[wdhandler]]]
            formatter = wdformatter
    
    [[formatters]]
        [[[wdformatter]]]
            format = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
```

The `logging` documentation is not great, but it's worth going through it.

tkeffer:
Oh, one last suggestion. All those calls to
```python
log.log(logging.INFO, "Informational message")
```
can be more simply expressed as
```python      
log.info("Informational message")
```
Same with DEBUG, etc.

gjr80:
replaced log calls with info/debug/error calls

PR #442, opened 2019-09-06
Fix engine to exit correctly on Terminate signal

sshambar:
Re-send SIGTERM to self after exiting main loop so process exits
with expected exit code.

SIGTERM was being handled, and an exception raised (which results
in a exit code of 1).  Process managers like systemd will see this
as an error code, and mark the process as failed.  If the process
exits like it would from a SIGTERM signal, then the process would
appear to have terminated normally (typically exit status of 128 + signal#)


tkeffer:
Good call that this needs a fix.

But, why not just call sys.exit()?

sshambar:
Exiting with 0x80+signal# is what glibc does, I'm not 100% positive other libc's do the same (Solaris, ulibc, HPUX etc :) -- so I figured a signal to the default handler will always do the Right Thing(tm).

tkeffer:
Thanks! This will most likely appear in V4.0.

PR #465, opened 2019-11-03
Made "make rpm-package" to support Fedora properly

HQJaTu:
Expansion of `make rpm-package` to support Fedora / RHEL propely with correct macros and destinations.
The intention is to keep existing Suse packaging while allowing other systems to be supported.


matthewwall:
i have not forgotten this merge request.  i'm not happy about maintaining one unit file for redhat and a different one for suse (i thought the whole point of systemd was to unify all of that?).  but the bigger issue is maintaining a separate spec file for each of the redhat releases.  i'm probably going to do one for rhel7 and another for rhel8, then maintain a yum repository at weewx.com.

HQJaTu:
Yeah. There are multiple things I don't like in life, but unfortunately for me, nobody cares. :-)

As you can see from my commit, it's not about systemd, it's about different names for RPM macros. Suse-people define them with different names than RedHat/Fedora-people.

Stormwind99:
When using `make rpm-package`, I received rpm build errors about shebangs indicating python instead of python2 or python3 (which Fedora rpms apparently prohibit).

By switching all python shebangs to python3, I was able to build the rpm and install it (but a few warnings/non-fatal errors did occur during install):
https://github.com/Stormwind99/weewx/commit/e1dbdbb0b605a73aef1eb704f637a1d65053ffe2

matthewwall:
the intent for weewx4 is to have two different rpm files for redhat systems, one for rhel7 and one for rhel8.  the rhel7 one will use 'python', the rhel8 will use 'python3'

matthewwall:
i really do not want to support a different systemd unit file for each operating system (or operating system version).

there are two things preventing deployment of a systemd unit file in packages that target systems that have been completely taken over by systemd (i.e., python3-weewx.deb and weewx-el8.deb):

1) the documentation consistently refers to init.d when describing how to start/stop weewx as a service.  the docs should be reworked so that they are not specific to init.d or systemd or any other syntax.  (most extensions have the same issue, so whatever we adopt for the docs will propagate to the extension readmes)

2) the weewx4 packages use `/etc/default/weewx` to specify which python interpreter should be used.  any systemd unit file must respect that.  this probably just means invoking `/usr/bin/weewxd` instead of `/usr/share/weewx/weewxd`, but it needs to be tested.

tkeffer:
I'm taking it that this PR is dead...? Closing out. We can re-open if I'm wrong.

PR #467, opened 2019-11-07
Do not directly instantiate Cursor, use the Connection cursor method.

bellrichm:
With change I am seeing my memory usage staying stable.
I also did a quick browse of the Sqlite Connection class. It does look to me that it does cleanup old Cursors.

### DIFF
--- a/bin/weedb/sqlite.py
+++ b/bin/weedb/sqlite.py
@@ -90,7 +90,7 @@ def _get_filepath(SQLITE_ROOT, database_name, **argv):
     # not defined:
     root_dir = SQLITE_ROOT or argv.get('root', '')
     return os.path.join(root_dir, database_name)
-    
+
 class Connection(weedb.Connection):
     """A wrapper around a sqlite3 connection object."""
 
@@ -127,7 +127,7 @@ class Connection(weedb.Connection):
     @guard
     def cursor(self):
         """Return a cursor object."""
-        return Cursor(self.connection)
+        return self.connection.cursor(Cursor)
 
     @guard
     def execute(self, sql_string, sql_tuple=()):
@@ -203,9 +203,6 @@ class Cursor(sqlite3.Cursor):
 
     # The sqlite3 cursor object is very full featured. We need only turn
     # the sqlite exceptions into weedb exceptions.
-    def __init__(self, *args, **kwargs):
-        sqlite3.Cursor.__init__(self, *args, **kwargs)
-
     @guard
     def execute(self, *args, **kwargs):
         return sqlite3.Cursor.execute(self, *args, **kwargs)


tkeffer:
Thanks Rich. That was some impressive sleuthing, and you made it easy with a PR!

I incorporated it into the `development` branch, rather than `master`, as we have no plans for a new 3.9.x release. 

I'm a little surprised that the function `sqlite3.Connection.cursor()` will take a cursor constructor. The [documentation](https://docs.python.org/2.7/library/sqlite3.html#sqlite3.Connection.cursor) makes it sound like it will only take a factory function. But, it seems to work.

tkeffer:
I've been running your PR for about 4 days now. VmRSS seems flat, but VmSize is rising very slightly. It seems to help, but then, my memory leak was not very big. 

image

bellrichm:
Interesting. Unfortunately I did not keep the graphs. I'll swap out the fix and post what I am seeing on my Pi. In the meantime here are a couple of tests I ran with my test class. All testing is with python 2.

Running on a pi, using the 'factory'.
Before looping 100000 times: mem_size: 10.472656 mem_rss: 5.902344 mem_share: 3.855469
After loop:: mem_size: 10.597656 mem_rss: 6.835938 mem_share: 4.667969
Garbage collected 0 objects
After garbage collection:: mem_size: 10.597656 mem_rss: 6.835938 mem_share: 4.667969

Running on a pi, instantiating directly.
Before looping 100000 times: mem_size: 10.472656 mem_rss: 5.906250 mem_share: 3.859375
After loop:: mem_size: 15.535156 mem_rss: 11.921875 mem_share: 4.730469
Garbage collected 0 objects
After garbage collection:: mem_size: 15.535156 mem_rss: 11.921875 mem_share: 4.730469

Running on ubuntu, using the 'factory'.
Before looping 100000 times: mem_size: 31.597656 mem_rss: 7.429688 mem_share: 4.308594
After loop:: mem_size: 31.781250 mem_rss: 8.328125 mem_share: 5.007812
Garbage collected 0 objects
After garbage collection:: mem_size: 31.781250 mem_rss: 8.328125 mem_share: 5.007812

Running on ubuntu, instantiating directly.
Before looping 100000 times: mem_size: 31.597656 mem_rss: 7.421875 mem_share: 4.300781
After loop:: mem_size: 41.183594 mem_rss: 17.746094 mem_share: 4.925781
Garbage collected 0 objects
After garbage collection:: mem_size: 41.183594 mem_rss: 17.746094 mem_share: 4.925781

bellrichm:
Because I have been trying to stabilize my environment, I don't have a nice history to show.
For this conversation, ignore the first drop in memory. That is when I disabled a skin that seems to also be leaking. The steady state in the middle is with the updated sqlite.py.  Around 9 AM today I backed out the update. That is when the steady increase starts. From past experience, in about a week the Pi will crash.
Out of curiosity, if I may ask, what if any, extra services are you running? The memory increase I have is primarily due to one service.

dayMemoryUsage

PR #472, opened 2019-12-02
wmr200: Discard packets with 'interval' of zero

cmanton:
Packets with pc adjusted time that occur after the
startup timestamp but before the minimal interval
expected (60 seconds) produce an interval of 0.

Issue: 375


tkeffer:
Wow. That's a subtle one!

cmanton:
Hopefully that will do the trick

PR #474, opened 2019-12-20
Robustness enhancements to cc3000 driver including py2/py3 support

chaunceygardiner:
cc3000.py:
- changes to work on both py2 and py3 (tested on both)
- change timeout to 1s (more not needed)
- retry commands when reading cmd echo times out (the signal of failure)
- use timeout of 20s for memory clear (which takes about 12s)
- various changes as assumptions of commands not currently correct (tried on 2 units)
- use weewx.wxformulas.calculate_rain for rain delta
- use weewx.crc16

engine.py:
- Failure to get time from driver is not treated as an error.
- Print stack trace when driver fails to load.

rsyncupload.py:
- Add timeout option.
- Change to allow single file to be rsynced.

util/logwatch/scripts/services/weewx:
- Fix as copied files text has changed in 4.x.
- Changes to organize very detailed cc3000 output in logwatch reports.

### DIFF
--- a/bin/weeutil/rsyncupload.py
+++ b/bin/weeutil/rsyncupload.py
@@ -25,7 +25,8 @@ class RsyncUpload(object):
 
     def __init__(self, local_root, remote_root,
                  server, user=None, delete=False, port=None,
-                 ssh_options=None, compress=False, log_success=True):
+                 ssh_options=None, compress=False, log_success=True,
+                 timeout=None):
         """Initialize an instance of RsyncUpload.
         
         After initializing, call method run() to perform the upload.
@@ -46,6 +47,7 @@ class RsyncUpload(object):
         self.ssh_options = ssh_options
         self.compress = compress
         self.log_success = log_success
+        self.timeout = timeout
 
     def run(self):
         """Perform the actual upload."""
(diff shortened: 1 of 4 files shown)


tkeffer:
Well done! It's nice to have such a close, careful look at the cc3000 driver!

I also agree that there's no reason to penalize a bad clock read.

My only comment is that, when reading bytes, I prefer to leave them as bytes, rather than decode and convert to unicode. Indeed, in one case, the method `CC3000.read()`, you decoded, but then immediately had to encode in order to check the CRC. Better to leave bytes as bytes.

Same when writing. If possible, form the output as a byte string, rather than as a unicode string (which would then have to be encoded).

But, that's small details. 

Thanks for your work, John.

But, that's small details.

chaunceygardiner:
Hi Thomas,

I appreciate you taking the PR.

As for your comment about leaving bytes as bytes, I noticed that’s what you did in the vantage driver and it makes sense to be consistent.  I’ll do that in a follow up in the next few days.

As for WeeWx 4, I’ve converted (locally) a bunch of code to py2/py3 (rtgd, forecast, purpleair) and have two of my three instances of WeeWx happily running WeeWx 4 under py3 for a couple of days now.  Thanks for all of your hard work!

PR #475, opened 2019-12-23
Development

chaunceygardiner:
cc3000 driver cleanup and fixes.                                                                                                                                          
(Tested on Py2/Py3.)                                                                                                                                                      
- Greatly improved startup when searching for missing archived records.                                                                                                   
  The reworked gen_records doesn't request all records on the device.                                                                                                     
  Rather, it (over) extimates how many non-archive records might be                                                                                                       
  in the returned records (based on the timespan) and adds that number                                                                                                    
  to the request.  As such, there is always a search for records, but                                                                                                     
  the search is fast.                                                                                                                                                     
- Only convert to decode/encode in read/write if running under Py3.                                                                                                       
- Fixed various issues to get driver working for all wee_device commands.                                                                                                 
- Add history-since (the option was there, but the implementation was missing).                                                                                           
- Removed unused retry_wait setting.                                                                                                                                      
- Removed stripping of unprintable chars as this has not been observed and                                                                                                
  will be handled by the CRC.                                                                                                                                             
- Update doc with tested driver version.

### DIFF
--- a/bin/weeutil/rsyncupload.py
+++ b/bin/weeutil/rsyncupload.py
@@ -25,7 +25,8 @@ class RsyncUpload(object):
 
     def __init__(self, local_root, remote_root,
                  server, user=None, delete=False, port=None,
-                 ssh_options=None, compress=False, log_success=True):
+                 ssh_options=None, compress=False, log_success=True,
+                 timeout=None):
         """Initialize an instance of RsyncUpload.
         
         After initializing, call method run() to perform the upload.
@@ -46,6 +47,7 @@ class RsyncUpload(object):
         self.ssh_options = ssh_options
         self.compress = compress
         self.log_success = log_success
+        self.timeout = timeout
 
     def run(self):
         """Perform the actual upload."""
@@ -56,7 +58,8 @@ class RsyncUpload(object):
         # that as a request to copy all the directory's *contents*,
         # whereas if it doesn't, it copies the entire directory.
         # We want the former, so make it end with a slash.
-        if self.local_root.endswith(os.sep):
+        # Note: Don't add the slash if local_root isn't a directory
+        if self.local_root.endswith(os.sep) or not os.path.isdir(self.local_root):
             rsynclocalspec = self.local_root
         else:
             rsynclocalspec = self.local_root + os.sep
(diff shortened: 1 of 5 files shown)


tkeffer:
I don't understand why the check against Python version is necessary in line 705. Why not just leave the results as `bytes`? Why decode and treat as Unicode?

chaunceygardiner:
Hi Thomas,

Thank you for merging the PR.

Based on your comment from my previous PR, I spent quite a few hours on two attempts at keeping the reads from the device as bytes and always passing bytes to the write.  Although the code was working, I wasn’t happy with the code from an aesthetic standpoint.

I haven’t yet found a clean way to do this.  I had encode/decode at too many places.  User entered text (e.g., set DST) needed to get encoded.  Presenting it back to the user needed to be decoded.  In general, anything shown to the user (e.g., wee_import —info) needs to be decoded.  Time was another one.  strftime returns Unicode and needs to be decoded to include in the TIME= command.  I initially didn’t decode for the log file and changed logwatch to optionally allow for the byes indicators.  But I just didn’t like the way the log file read.  So I decoded everything in the log.  I also found the sensor map odd.  The key is unicode, with the value bytes.  It was tough to remember when I needed bytes and when not.

I’ll keep thinking about this.  What are your thoughts about the bytes in the log file from a readability standpoint and also the log output being different between Py2 and 3?  How about presenting bytes to the user?  Also, when Py2 support is dropped, do you still envision all of the encodes/decodes staying?

Perhaps I just need to become more comfortable with all the encodes/decodes.

Cheers,
John

tkeffer:
To answer your questions:

1. Don't like bytes in the log file unless they are meant to represent byte data coming off the device, particularly binary data. Then they make sense.
2. It follows then that log output should be identical under Python 2 and 3. 
3. Generally not a good idea to present bytes to the user --- too confusing. There might be exceptions where the information is intended to be used for debugging the raw data coming off the device. This is a similar situation to the logs.
4. I've been trying to structure my code so that it reads like Python 3 code. This is particularly true with the `import` section. That way, 5 years down the line when we drop Python 2 support, you're just removing the Python 2 stuff. Nothing else changes.

PR #480, opened 2019-12-28
CC3000: Add get/reset max and min values commands, get debug_serial working and update comments.

chaunceygardiner:
Add get/reset min and max values.  Make those and some others available in wee_device and when running cc3000.py as a main program.

Get debug_serial = 1 working in PY3.  Had to special case PY2/3 in _format_bytes as iterating over bytes in PY3 yields ints (and calling byte2int on an int throws an exception).

Updated comments.


tkeffer:
Thanks, John

FYI, there's a function [`y_or_n()`](https://github.com/weewx/weewx/blob/0ecd4fed8325b1ee2afdc44894df52fcd754373f/bin/weeutil/weeutil.py#L1403) available for managing prompts.

chaunceygardiner:
> Thanks, John
> 
> FYI, there's a function [`y_or_n()`](https://github.com/weewx/weewx/blob/0ecd4fed8325b1ee2afdc44894df52fcd754373f/bin/weeutil/weeutil.py#L1403) available for managing prompts.

PR 481 sent.

PR #484, opened 2019-12-29
Reimplement gen_records(), refactor command(), doc changes, logwatch changes.

chaunceygardiner:
Reimplement gen_records which is used to download archive records from the station.  This is greatly simplified by having it used the command() function (like all other commands).

Refactor command() function as the retry duplicated code (I was lazy).  It also allows for more than one retry (so far, this has never been needed).

Lots of doc changes.

The logwatch script is now complete for all commands that are retried.

Bug Fix: nminutes->options.nminutes


tkeffer:
logwatch is becoming a hard-to-maintain monster, written in Perl, no less. I wonder if there is some automated way we could scan the python code, searching for log statements, and generate the script?

chaunceygardiner:
I agree it is a mess and I made it a bigger mess.

My thoughts.

To generate the script, one would need to know which statements to count, which to list and which to ignore.  Also, what labels to assign.  Also, in the general case, one will not know the actual text of the log record written, so a regex seems like a hard requirement.

It would be feasible to add a “#LOGWATCH ...” comment before the log statement with a regex, an optional count(<label>) and/or list(<label>) — or an ignore.  Variable names could be generated from the label so that like labels will be grouped together in the group or list.

The actual log statement could be ignored.  And no need to annotate more than one of the same type of log statement.  And no need to have these comments proximate to the log statements, but it would be easier to keep track of them by having them proximate.

PR #485, opened 2019-12-29
Add Custom API connector for easier integration with custom platforms

paoliniluis:
I created a new custom API connector based on the WU connector so now anyone can easily push data into their own custom platforms.

### DIFF
--- a/bin/weewx/restx.py
+++ b/bin/weewx/restx.py
@@ -705,7 +705,7 @@ class StdCustomApi(StdRESTful):
             self.loop_thread = AmbientLoopThread(
                 self.loop_queue,
                 _manager_dict,
-                protocol_name="Wunderground-RF",
+                protocol_name="CustomAPI",
                 essentials=_essentials_dict,
                 **_ambient_dict)
             self.loop_thread.start()
--- a/bin/weewx/restx.py
+++ b/bin/weewx/restx.py
@@ -648,6 +648,86 @@ class StdWunderground(StdRESTful):
         """Puts new archive records in the archive queue"""
         self.archive_queue.put(event.record)
 
+class StdCustomApi(StdRESTful):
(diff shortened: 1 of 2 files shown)


tkeffer:
Thanks, Luis, for your contribution, but I guess I don't understand what problem this is trying to solve.

1. It is not really a "custom" API. It is the Ambient protocol, with a specialized URL.
2. Is there much of a need for this? Can you give us a use case?
3. Assuming there is a need, why not just package it as a 3rd party extension? 
4. Finally, it is exactly the same as the WU API, except it allows the user to specify an URL to post to. So, if there really is a need for something like this, why not just use the WU API, except allow the URL to be specified as an option?

paoliniluis:
Hi Tom! thanks for your super quick review, let me address your questions:
1) exactly, I used the word "custom" as I wanted to express that you could hit a custom endpoint with this
2) I came across this need as I was thinking about the way I could use WeeWX as the source of data of an AWS Lambda to create alternative weather platforms (as WU is becoming a true headache for many people after IBM acquisition). I really don't know if many more people have the same issue, but I wanted to create a simple and open connector so anyone could hit an API with the data that the platform provides without much tinkering in the code
3) That can be an option, I just wanted to include it as a standard so you could have the option as soon as you install WeeWX by default, so anyone that installs the platform can start straight away pushing data without installing an extension.
4) I thought about including it as something different so users don't get confused by WU and a custom endpoint.

Thanks for reading and open to any changes!

tkeffer:
My apologies, but this does not sound very compelling. The use case is hypothetical and, even then, there is a great workaround: just include an extension.

There are many WeeWX uploaders that do this, with their own custom endpoint. For example, look at [WeeRT](https://github.com/tkeffer/weert-js).

PR #490, opened 2020-01-15
Add barometerRate, et. al. to weewx.units

chaunceygardiner:
Add altimeterRate, barometricRate and pressureRate in new group_pressureRate.

Support inHg/h, mbar/h, mmHg/h and hPa/h units and conversions between them.

I’ve followed the group_rainrate group for naming conventions.

Added tests.

Pressure rate of change is fairly common to report.  Thanks for considering adding this.


tkeffer:
Seems reasonable. However, the documentation should also be updated. In particular, the appendix *Units* in the Customizing Guide. 

Don't worry about updating the appendix *Archive Types*. I think I'm going to eliminate it, in favor of a better description of wind types.

tkeffer:
Oh, one other thing: the types `altimeterRate` and `pressureRate` make sense, as they build on the types `altimeter` and `pressure`. But, why `barometricRate`?

chaunceygardiner:
OK, I’ve updated the customization guide and changed barometricRate to barometerRate.

Funny you asked why I chose barometricRate.  I used that because that’s a term I noticed being used when researching the topic, but I like the consistency of using barometerRate.

tkeffer:
One inconsistency: you have `group_pressure_rate`, and `group_pressurerate`

PR #494, opened 2020-01-29
Bug fix for clean install w/ cc3000 driver and Python 3

chaunceygardiner:
genStartupRecords was not expecting a since_ts value of None.  That happens if the db has just been created and Py3 doesn’t like it.

TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'


tkeffer:
How come your fork is so many commits ahead of weewx/development?

chaunceygardiner:
Hi Tom,

My fork is identical to weewx/development.  (I just did a compare to make sure.)

I pull and merge from weewx/development every time it changes and push to my fork.  I believe these count as commits ahead of weewx/development.

Perhaps there is a better way or I’m just doing it wrong.  Please advise.  I’d rather not have it say I’m 54 (at present) commits ahead.

PR #497, opened 2020-01-31
Support new WeeWX installation with large amount of CC3000 records in station.

chaunceygardiner:
This change tackles reading a large number of records from the CC3000 (something likely to happen on a clean WeeWX installation).  Alternating between reading records from the station and inserting them into the database causes checksum errors, etc.  This change reads the records into memory first and then parses and yields them for insertion into the database.

Tested with a clean WeeWX install (i.e., no database) and 11,889 records in the station.

### DIFF
--- a/bin/weewx/drivers/cc3000.py
+++ b/bin/weewx/drivers/cc3000.py
@@ -1155,11 +1155,16 @@ class CC3000(object):
     def gen_records_since_ts(self, header, sensor_map, since_ts):
         if since_ts is None:
             since_ts = 0.0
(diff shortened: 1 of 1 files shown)


tkeffer:
Interesting strategy.

chaunceygardiner:
It certainly wasn’t the first, second or third thing I tried.  It finally came to me to try it because I could dump all records with wee_device and never hit an error.  I guess the station just doesn’t like waiting when the DOWNLOAD command is invoked.  The station has only 2MB of flash memory so reading into memory doesn’t seem unreasonable to me.

PR #498, opened 2020-02-01
Don’t cache “now” before potentially long running catchup.

chaunceygardiner:
I’m not sure if this is a problem with other stations, but the CC3000 will return records that are created after a download of records is initiated.  From the CC3000 doc on the DOWNLOAD command:

“It should be noted, the CC-3000 will record even while outputting. This means that should a logging interval occur during the download this record will also be output.”

As such, with an accurate clock, records are ignored if they were created self.archive_delay seconds after the catchup was initiated.


tkeffer:
Tricky one, that. Thanks for spotting this.

PR #501, opened 2020-02-03
Add/fix pm1_0, pm2_5 and pm10_0 observation types.

chaunceygardiner:
Perhaps I’m not understanding something, but it looks like the air quality observation types in units.py should match the extended schema.

It also looks like most of the rest the extended schema is missing from units.py.  I’m assuming that is intentional.

### DIFF
--- a/bin/wee_import
+++ b/bin/wee_import
@@ -837,8 +837,8 @@ def main():
             log.info("**** Nothing done.")
             exit(1)
         except weeimport.weeimport.WeeImportIOError as e:
-            print("**** Unable to load source file.")
-            log.info("**** Unable to load source file.")
+            print("**** Unable to load source data.")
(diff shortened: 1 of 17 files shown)


tkeffer:
Simple oversight. They are not the only ones that got forgotten. There's also `nh3`, `no2`, `o3`, and others.

rezin8:
Man, I should have looked at the 4.0 code before I started working on integrating my new PM2.5 into 3.9... I've been working on this for 2 days, and I have 4.0b11 sitting there from when I was working on another issue with Tom... Grrr......

tkeffer:
There are lots of goodies in V4.0!

Check the [change log](https://raw.githubusercontent.com/weewx/weewx/master/docs/changes.txt). You may find some other pleasant surprises.

PR #505, opened 2020-02-06
Added Inside Humidity to Seasons skin (if sensor provides it)

terrymacdonald:
Added the Inside Humidity to the Seasons skin. All Inside Humidity
entries are wrapped in a has_data if statement so that they only
up if the sensor is generating the underlying data.

I have made this change on my own build, and thought it would be useful for others whose weather station has an inside humidity sensor.

### DIFF
--- a/skins/Seasons/current.inc
+++ b/skins/Seasons/current.inc
@@ -73,6 +73,12 @@
         <td class="data">$current.inTemp</td>
       </tr>
 #end if
+#if $day.inHumidity.has_data
+      <tr>
+        <td class="label">$obs.label.inHumidity</td>
+        <td class="data">$current.inHumidity</td>
+      </tr>
+#end if
 #if $day.extraTemp1.has_data
       <tr>
         <td class="label">$obs.label.extraTemp1</td>
--- a/skins/Seasons/hilo.inc
+++ b/skins/Seasons/hilo.inc
@@ -221,6 +221,21 @@
       </tr>
       #end if
 
+      #if $day.inHumidity.has_data
+      <tr>
+        <td class="label">$obs.label.inHumidity</td>
+        #for $archive in $archive_data
+        <td class="data new_row hilo_$archive[0]">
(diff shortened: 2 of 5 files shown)


tkeffer:
Not a bad idea.

But, let's not load up Seasons with every sensor!

terrymacdonald:
Understood 😀. I thought it was common enough on most weather stations that
it was worth adding.

How would you feel about an automatic page refresh for Seasons as another
pull request? Maybe have it use the archive time period somehow was the
page refresh time period so you can always be sure the page has the latest
information on it

Thanks
Terry

matthewwall:
1) not everyone *wants* the page to automatically refresh

2) how often should the page refresh?  different hardware has different refresh rates.  you could use the weewx archive interval as the refresh period (simple use of cheetah for that)

3) page refresh is a really crude way to get 'live' data.  better to use a modern javascript approach instead of brute-force reloading the entire page

4) there are a few approaches to doing the live data - mqtt with websockets, weert, individual ajax.

in my opinion it would be better to spend some time generalizing an ajax/websockets approach than to futz with page reloads

terrymacdonald:
Hi Matthew.

That is very true. This was going to be a simple optional stopgap solution,
selectable through a select or toggle of some kind on the title bar to give
people the ability to automatically refresh if they choose.

I completely agree moving to an asynchronous style of UI using an API as
the source of the data should be the ultimate goal, but there is quite a
lot of difference in the work between those two approaches.

I think people would appreciate a refresh button at this stage rather than
wait until the API is done. After all, it's pretty much what we all do
manually each time we click to the browser tab now.

Thanks
Terry

tkeffer:
We've tried to avoid opinionated features, loaded with policy. An automatic "refresh" has been discussed before, and rejected, for these reasons. 

It's easy enough for the user to add him/herself!

terrymacdonald:
Fair enough. I'll keep it in my fork.

Thanks for all your hard work in putting this system together. Really do
appreciate it!

Thanks
Terry

PR #527, opened 2020-04-24
Remove executable permissions from content

tyll:
Installing files without a shebang as executable results in a warning
with rpmbuild. This is at least triggered for the license.txt file for
the Seasons skin. Remove the executable permissions from other content
files for uniformity.


tkeffer:
Thanks, Till! I never noticed that.

PR #528, opened 2020-04-28
Add driver for the Gill Metpak Pro weather station

DylanGore:
Added a new driver to support the [Gill MetPak Pro](http://www.gillinstruments.com/products/anemometer/metpak-pro-weather-station.html)

### DIFF
--- a/bin/weewx/drivers/metpak_pro.py
+++ b/bin/weewx/drivers/metpak_pro.py
@@ -1,0 +1,385 @@
+#
+# Copyright (c) 2020 John Ronan <jronan@tssg.org>
+# Copyright (c) 2020 Dylan Gore <dgore@tssg.org>
+#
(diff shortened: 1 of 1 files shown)


tkeffer:
I appreciate the work you put into this. It looks like you did a good job.

But, there is not enough use of the Gill weather station to warrant including it as a supported station, at least not yet. It doesn't show up at all in weewx-user, nor on the user's map.

It would be best to package it as an extension, then include it in the [Wiki](https://github.com/weewx/weewx/wiki). See the section [*Extensions*](http://www.weewx.com/docs/customizing.htm#extensions) in the Customizing Guide for details on how to do this.

jpronans:
Fair enough. If its ok with you, we'll keep it as a fork for the moment until we Iron out all the bugs
We plan to deploy several of these and have them feeding the User map etc.

PR #531, opened 2020-05-03
Add field wind_samples to Vantage driver archive records

gjr80:
Refer [this weewx-user thread](https://groups.google.com/d/topic/weewx-user/k9NXJPytVyk/discussion).


tkeffer:
Had to make minor change: the dash value for `wind_samples` is zero, not `0xff`.

Thanks, Gary

PR #536, opened 2020-05-05
fileparse: Use math.ceil()

tyll:
Use math.ceil() instead of implementing it manually. This makes it more
clear what is happening without having to recognise the pattern for it.

### DIFF
--- a/examples/fileparse/bin/user/fileparse.py
+++ b/examples/fileparse/bin/user/fileparse.py
@@ -15,7 +15,7 @@
 # See http://www.gnu.org/licenses/
 
 
-# This driver will read data from a file.  Each line of the file is a 
+# This driver will read data from a file.  Each line of the file is a
 # name=value pair, for example:
 #
 # temperature=50
@@ -49,6 +49,7 @@
 
 from __future__ import with_statement
 import logging
+import math
 import time
 
 import weewx.drivers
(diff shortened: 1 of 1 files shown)


tkeffer:
Except that's not what the code does. It is intended to *round* to the nearest integer. For example,

```python
print(math.ceil(1.2))
```

gives `2`, when what we want is `1`.

tyll:
Awesome, now I understand. So it should be

```python
int(round(1.2))
```

Does this work for you?

I will update this, then. I noticed that other drivers use the same code as fileparse, so I can adjust them, too, if you do not object.

tkeffer:
The `int()` is unnecessary. If the number of decimal points is not specified, then the result of `round()` is always an integer. So, I suppose we could replace all those

    int(time.time() + 0.5)

with

    round(time.time())

but... why?

gjr80:
> If the number of decimal points is not specified, then the result of round() is always an integer.

Do you mean an integer or python type `'int'`? `round(1.2)` returns a python type `'int'` under python3 but a type `'float'` under python 2.7

tkeffer:
You're right.

Tyll, you're really not going to simplify any more than it is already. 

There are bigger problems out there! :-)

tyll:
> You're right.
> 
> Tyll, you're really not going to simplify any more than it is already.

The simplification is that it gives it an explicit name about what is happening. I was not aware that `int()` always rounds down. Since as far as I see, the documentation about how to write a driver is the fileparse plugin, I stumbled upon this code and I tried to understand what is happening. My using `int(round(time.time()))` instead of `int(time.time() + 0.5` there is no/less ambiguity about what is meant to happen. Since I learned this now, closing this works for me but it will not help the next person stumbling upon this.

PR #538, opened 2020-05-07
Fixed TypeError caused by ord() function when on Python 3

radiome7:
After dist-upgrade debian stretch -> buster I had an issue with weewx and WMR918 weather station:

May  7 22:24:13 host weewx[1721] CRITICAL __main__:     ****    File "/usr/share/weewx/weewx/drivers/wmr9x8.py", line 243, in genLoopPackets
May  7 22:24:13 host weewx[1721] CRITICAL __main__:     ****      buf.extend(list(map(ord, self.port.read(preBufferSize - len(buf)))))
May  7 22:24:13 host weewx[1721] CRITICAL __main__:     ****  TypeError: ord() expected string of length 1, but int found
May  7 22:24:13 host weewx[1721] CRITICAL __main__:     ****  Exiting.

After removing map() and ord() functions everything works as expected again.

### DIFF
--- a/bin/weewx/drivers/wmr9x8.py
+++ b/bin/weewx/drivers/wmr9x8.py
@@ -240,7 +240,7 @@ class WMR9x8(weewx.drivers.AbstractDevice):
         wm918max = max(list(wm918_packet_type_size_map.items()), key=operator.itemgetter(1))[1]
         preBufferSize = max(wmr9x8max, wm918max)
         while True:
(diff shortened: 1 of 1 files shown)


tkeffer:
Thanks so much, but this was fixed in commit https://github.com/weewx/weewx/commit/9844b6b085aed244a88b999ffb3e8bdf2008df20

BTW, your solution would not work under Python 2.

PR #547, opened 2020-05-11
Auto reboot CC3000 if station stops sending observations.

chaunceygardiner:
One of my two CC3000s stopped sending observation values today.  I had to reset the device to return it to normal operation.  This PR automatically reboots the device when this occurs.

This appears to be extremely rare, but must have been observed at least once before as existing code detected and logged it.


tkeffer:
This looks like a reasonable patch, but could you restructure your PR so it does not include 57 commits? One or two should do.

For example, you can get your `master` up to date, then create a branch with your changes, then PR that branch.

chaunceygardiner:
Thanks.  Will close and resend.

Issue #491, opened 2020-01-15
Allows StdWXCalculate to calculate rain from total rain


tkeffer:
The goal here is to make "total rain" a first class observation type. User would be able to get `rain` from, say, `dayRain`.

With the introduction of the aggregate type `diff` in V4.0, it should also be possible to use `dayRain` in plots.

NB: It is way too late to squeeze this in V4.0. It will have to wait until V4.1.

matthewwall:
more generally:

* get a delta value from a total value (requires state - need the last value - and wraparound detection)
* get a total value from a delta value (requires the definition of a timespan)

you definitely want these at the reporting level, where you display cumulative values with incremental values, for example.

but are they necessary at the calculation level, where you insert the result into a packet, and thus possibly into the data storage?

tkeffer:
I'm thinking of the former, more than the latter.

I am seeing drivers calculate rain from rain totals so often, that it struck me it's a lot like having drivers calculate `dewpoint`. Getting `rain` from a rain total makes it a derived quantity.

I should also add that, if for some reason your driver does not emit `rain`, and you really don't want it in your database, you either leave it out of `[[Calculations]]`, or set its option to `None`. Or, not include it in the database schema.

The new xtypes facility also has a `diff` aggregate, which would allow you to use `rain` in the reporting layer even if all you have is rain totals in the database.

gjr80:
Just trying to understand what is required here, if I understand correctly it is as simple as (1) creating a 'rain' xtype [let's call it `RainPourer` - not as good as `PressureCooker` but on par with `RainRater` :)] that calculates `rain` from a cumulative rain field and (2) adding `rain` to `StdWXCalculate` via the new `RainPourer` xtype. State would be stored in the `RainPourer` xtype.

A couple of questions. 
- Is the intent for the chosen cumulative rain field (1) to be user specified (2) chosen from a list of likely suspects (eg `dayRain`, `monthRain`, `yearRain` etc) or (3) a single hardwired field ?
- the reference to using `dayRain` in plots using `diff`, understand this is a done deal and already implemented, ie nothing more to do ?

tkeffer:
You've got the right idea. 

1. The user should be able to pick the source total (`dayRain`, etc.). This is because different devices emit different totals, so we have to be flexible.
2. I probably should not have mentioned `diff`. It was just a thought that given a field like `totalRain`, we can do plots of `rain`. But, that doesn't take away from the need to treat `rain` as a derived quantity. 

[...]

PR #553, opened 2020-05-13
FileParse now handles cumulative rain

gjr80:
Appreciate that the fileparse driver is included as a basic driver example but I know of two cases now where users have wanted to read rain data and issues have arisen since fileparse expects per-period rain data. This can be problematic if the fileparse data file and fileparse driver polling end up out of sync (data will be missed).

This PR modifies the fileparse driver to accept cumulative rain data, this is much more tolerant of the data file generation and fileparse driver polling getting out of sync.

If there is no appetite to change the fileparse driver then by all means reject the PR.

### DIFF
--- a/examples/fileparse/bin/user/fileparse.py
+++ b/examples/fileparse/bin/user/fileparse.py
@@ -35,6 +35,14 @@
 #     path = /var/tmp/wxdata     # location of data file
 #     driver = user.fileparse
 #
+# If WeeWX field rain is to be derived from a cumulative total, add the option
+# cumulative_rain under [FileParse] as follows:
+#
+# [FileParse]
+#     ...
+#     cumulative_rain = true     # true if rain is derived from a cumulative
+#                                # value, otherwise false or omit setting
+#
 # If the variables in the file have names different from those in the database
 # schema, then create a mapping section called label_map.  This will map the
 # variables in the file to variables in the database columns.  For example:
@@ -52,9 +60,12 @@ import logging
 import time
 
 import weewx.drivers
+import weewx.wxformulas
+
+from weeutil.weeutil import tobool
 
 DRIVER_NAME = 'FileParse'
-DRIVER_VERSION = "0.7"
+DRIVER_VERSION = "0.8"
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tkeffer:
I think I'd rather allow `fileparse` to read cumulative rain, and then have `StdWXCalculate` be able to derive `rain` out of it. That would be a more general solution. See issue https://github.com/weewx/weewx/issues/491.

Do you want to try and take that on?

gjr80:
Darn, I knew I saw something recently about calculating rain from cumulative and when I found `calculate_rain` I thought that must have been it, of course it was #491. Sure, put me down for it, probably a good practical case to get my head around the new StdWXCalculate. Just don’t push out 4.1 in the next couple of weeks, busy next week.

staze:
Don't want to reopen this, but curious where this stands. Does FileParse currently handle cumulative rain, or does it still require delta?

gjr80:
> Don't want to reopen this, but curious where this stands. Does FileParse currently handle cumulative rain, or does it still require delta?

No, FileParse still needs delta rain (well strictly speaking FileParse doesn't care, it just passes on what it's given, WeeWX requires delta rain). It's not forgotten though, the approach that is being taken is to treat rain as just another derived observation (like apparent temperature, missing pressures etc) and give the WeeWX `StdWXCalculate` service the ability to calculate WeeWX field `rain` from a cumulative rain value. This way not only will the existing FileParse (indirectly) support cumulative rain but you give a similar ability to future drivers, plus with some careful coding we can also allow other fields to have similar 'derived delta from cumulative' support, eg lightning strike counts.

Whilst some working code has been produced it is very much still a work in progress that I hope can be included in 4.2.0. Issue #491 refers. A few other things taking my time at present.

staze:
> > Don't want to reopen this, but curious where this stands. Does FileParse currently handle cumulative rain, or does it still require delta?
> 
> No, FileParse still needs delta rain (well strictly speaking FileParse doesn't care, it just passes on what it's given, WeeWX requires delta rain). It's not forgotten though, the approach that is being taken is to treat rain as just another derived observation (like apparent temperature, missing pressures etc) and give the WeeWX `StdWXCalculate` service the ability to calculate WeeWX field `rain` from a cumulative rain value. This way not only will the existing FileParse (indirectly) support cumulative rain but you give a similar ability to future drivers, plus with some careful coding we can also allow other fields to have similar 'derived delta from cumulative' support, eg lightning strike counts.
> 
> Whilst some working code has been produced it is very much still a work in progress that I hope can be included in 4.2.0. Issue #491 refers. A few other things taking my time at present.

Thanks! I had seen the 491 stuff, but wasn't sure what the status was. Looks like you last left off asking for assistance. 

Hopefully 4.2 does include. It's not rainy at all in the lower willamette valley right now, so Fileparse is working fine. Hopefully it's all available by the time things start getting wet again. =D

Thanks for the quick response!

PR #570, opened 2020-06-13
Added "lux" group, and added lightning units

jbellanca:
1. Added luminosity observation type to the group dictionary, with new group "group_lux".
2. Added lightning observation types to the group dictionary, with various groups (distance, count, volt) (strike_count, disturber_count, noise_count, energy, and distance)
3. Added "group_lux" for luminosity, including label "lux"


tkeffer:
See comments in PR https://github.com/weewx/weewx/pull/571.

This PR is a little closer to what we need, but not quite there.

PR #571, opened 2020-06-13
Updated defaults for luminosity proposed pull #570

jbellanca:
Updated defaults for luminosity proposed pull #570


tkeffer:
A little more complicated than this.

The WeeWX unit groups cover a range of possible units (e.g., °C or °F) for a generic type of measurement (e.g., temperature). The unit `lux` measures illumination, so, logically, it should be part of a `group_illumination`. 

However, within WeeWX, we have traditionally grouped it with `group_radiation`, with a constant conversion factor. This could definitely be improved, but would require more thought than simply adding a `group_lux`.

Issue #574, opened 2020-06-16
TE923 driver reports less rain than the hardware console


twitham1:
The Meade manual claimed 2 values for rain gauge resolution, neither of which match the values observed on the console.  0.03 inches would report too high and 0.6578 mm would report too low; the latter is used by the current driver.

In 2012 I observed 36 bucket tips for every 1.0 inches displayed.  So I adjusted my wview at the time:

    // decode rain counter:
    sensors->rain = (float)(buf[31] * 0x100 + buf[30]);
    // TE923 manual says resolution == 0.03 inches, not 1 millimeter (0.039").
    // But console is not divisible by 0.03 after a few tips.  Each
    // bucket tip is actually about 1/36" or 0.27777777" (0.705556 mm).
    sensors->rain /= 36;

In 2014, [wview 5.21.0 fixed its driver](https://groups.google.com/forum/#!msg/wview/OouIf0hfUi0/tUEKVefEJn4J) to use 1/36" per rain count.

In 2015, issue #67 reported 10.66/9.9 correction needed in weewx, very near my current .705556/.6578 correction for weewx.

In 2020, [another new user reports rain too low](https://groups.google.com/forum/#!msg/weewx-user/JXN2p_V7dc8/WfnTIpBGAAAJ)

I will submit a pull request next, and ask TE923 owners on weewx-user if they have any evidence that the current multiplier is more accurate than 1/36 inch.  If it is decided a fix should be made, this could be mentioned in the release notes so anyone using a correction could remove it or adjust it at that upgrade.

twitham1:
Request for feedback: https://groups.google.com/d/msg/weewx-user/evqi12m5MLw/b6mddwgTAQAJ

PR #575, opened 2020-06-16
TE923: Fix issue #574 with 1/36 inch rain counter.

twitham1:
See also issue #574 for more details and examples of why I believe rain counter should report 1/36 inch to match hardware console.  Fixing the driver at the source seems better than all users having to add a correction.  Unless, of course, more users can show that the current value is working as it should.  Will ask them on weewx-users.

### DIFF
--- a/bin/weewx/drivers/te923.py
+++ b/bin/weewx/drivers/te923.py
@@ -46,14 +46,18 @@ From the Meade TE9233W manual (TE923W-M_IM(ENG)_BK_010511.pdf):
   Wind/Gust sampling interval: 11 seconds
   UV transmitting interval: 300 seconds
   Rain counter resolution: 0.03 in (0.6578 mm)
+    (but console shows instead: 1/36 in (0.705556 mm))
   Battery status of each sensor is checked every hour
 
 This implementation polls the station for data.  Use the polling_interval to
 control the frequency of polling.  Default is 10 seconds.
 
-The manual says that a single bucket tip is 0.03 inches.  In reality, a single
-bucket tip is between 0.02 and 0.03 in (0.508 to 0.762 mm).  This driver uses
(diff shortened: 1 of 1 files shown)


bin/weewx/drivers/te923.py
```diff
@@ -562,7 +566,7 @@ class TE923Configurator(weewx.drivers.AbstractConfigurator):
     }
     
     city_dict = {
-        0: ["ADD", 3, 0, 9, 0o1, "N", 38, 44, "E", "Addis Ababa, Ethiopia"],
+        0: ["ADD", 3, 0, 9, 1, "N", 38, 44, "E", "Addis Ababa, Ethiopia"],
```
twitham1, line 569: This is unrelated cleanup of automated 01 -> 0o1 octal change, better stated as simply 1.  Zero functionality change here.

tkeffer:
Let's give it a try. I also made an entry in the change log. https://github.com/weewx/weewx/commit/4beed401cd9ce8882f53833dc7267af56461d5be

twitham1:
Thank you Tom!

It turns out my new comment at line 2035 was not needed and can be removed or changed to document that the units reported there are millimeters.

That wee_device --minmax option reports the values that the console displays in the memory button.  I count 167 bucket tips in my database for the month of May and the console shows 4.64 inches, exactly matching the 1/36" records I have been keeping.  1/36*167=4.6388888, matching the hardware console!  This will be correctly reported as 118 millimeters by --minmax with the new multiplier.

tkeffer:
I wonder why on earth they chose 1/36" as a bucket size...?

twitham1:
I wonder if the geometry of the hardware just happened to generate 36 tips per inch when it was calibrated?  If so, then why on earth did they call this 0.03 inch or 0.6578 mm in the manual?  This documentation bug has caused many wrong measurements over the years.

PR #581, opened 2020-07-09
Added units and labels for lightning distance / strikes

claudobahn:
I didn't assign any units to lightning_energy, because looking at the datasheet of one of the popular ICs used for detection (as3935), their energy number does not represent any physical energy and is purely for relative sizing.


tkeffer:
Thanks for fixing this overlooked problem.

Could you please include an entry in the change log as part of this PR?

claudobahn:
@tkeffer I've squashed in an update to hopefully address your request. I then took it a little further by also adding the lightning types to the `customizing.htm` and providing default labels for at least the core two values (count / distance). While doing that, I also added a string format and label for the `count` unit.

I did not give lightning_disturber_count and lightning_noise_count the same treatment, because these appear to be mostly internal diagnostic readings present in the as3935 IC.

Let me know if you'd like any other changes.

claudobahn:
@tkeffer I'm wondering if somewhere something like this is needed:
`weewx.accum.extract_dict['lightning_strike_count'] = weewx.accum.Accum.extract_sum`

Basically, what I'm trying to make sure of is that if individual packets feeding into the loop have counts of 1, 5, or whatever, the LOOP packet accumulation sums up all the counts when creating the archive record.

If you could shed light on what the correct way of doing that would be?

claudobahn:
Ok, I think I've figured it out. Pull request inbound.

PR #582, opened 2020-07-09
Updated accumulator to extract lightning strike counts as a sum from LOOP packets

claudobahn:
I also synchronized `accum.md` w/ the current default configuration. If that's undesired, let me know and I'll pull it out.

This Pull Request changes the accumulator to use summing for lightning strike counts, since the loop packets report counts, the archive record needs to sum up all those counts for the archival period.

@matthewwall looping you in, in case you have an opinion, since you had created a driver for the AS3935 at one point.


tkeffer:
Nice detail to take care of! Thanks again.

PR #592, opened 2020-08-11
Add some control to logging before configuration file is processed

bellrichm:
Background:
I’ve got both the Simulator and MQTTSubscribe drivers using the Seasons skin running on an iPad under Pythonista and also under a-Shell. But, in both cases the ‘default’ logging doesn’t work. 
Under Pythonista, sys.platform returns ’ios’.  Using the pattern to support ‘darwin’, logger.py could be updated to support ‘ios’. But under a-Shell, sys.platform returns ‘darwin’ and access to /var/log is denied. So, I’m thinking a more general way to do the ‘initial’ (before the configuration file is read) logging setup might be useful.

A few ideas I’ve thought of are:
- Add a command line option (—skip-initial-logging?) to skip the ‘initial’ logging. This would consist of if statements around the logging calls and the obvious loss of information.
- Add a command line option (—initial-logging-to-console?) to have the ‘initial’ logging go to the console. When setting up the logger, the necessary information to log to the console would be passed in instead of an empty dictionary. This would result in default logging being in two places. First to the console then to either syslog or a rotating file.
- Add a command line option (—log-to-console?) to make logging to the console the default. If there is no [Logging] stanza in the configuration file, all logging will go the console.  This might have the added benefit that all necessary information (log information, loop packets, and archive records) to help people is in one  place, the console. This pull request is an attempt at implementing this idea.

If supporting running on the iPad is too ‘one off’’, it is easy enough to hack logger.py to remove the SyslogHandler and RotatingFileHandler if/when logger.py is updated.

Note, all of the development for this pull request was done on my iPad. WeeWX continues to amaze me.

### DIFF
--- a/bin/weeutil/logger.py
+++ b/bin/weeutil/logger.py
@@ -21,14 +21,14 @@ import weewx
 #  {value}: these are plugged in by the function setup().
 #  %(value)s: these are plugged in by the Python logging module.
 #
-LOGGING_STR = """[Logging]
+BASE_LOGGING_STR = """[Logging]
     version = 1
     disable_existing_loggers = False
 
     # Root logger
     [[root]]
       level = {log_level}
-      handlers = syslog,
+      handlers = {handlers},
 
     # Additional loggers would go in the following section. This is useful for tailoring logging
     # for individual modules.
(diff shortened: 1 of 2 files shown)


tkeffer:
My apologies, but I'm trying to understand what problem we are trying to solve. You say "initial" logging doesn't work. By this, I assume you mean logging from when the program starts up, until the first time `weeutil.logger()` is called. If so, this is a brief period, well before any drivers or services get instantiated.

Or, do you mean something else? Perhaps the "default logging", rather than "initial logging", configuration?

That aside, in general, I prefer putting options in the configuration file over the command line because, that way, all the state that goes into configuring and running the program goes in one place: you don't have to ask a user, "What command line options did you use?"

bellrichm:
Tom,
No apologies are necessary. This is definitely an edge case and I am fine with closing it.
The ‘problem’ is the logger setup before the config file is read/processed , around line 80 in weewxd. In one of my environments sys.platform returns ‘darwin’. But, I don’t have write access to /var/log...  And currently there is no way to change the default logger for ‘darwin’. 
Seriously, this is such an edge case we should probably just close this.
Thanks again for WeeWX. 
rich

PR #597, opened 2020-08-29
Bugfix : Wrong behaviour of to_float under Python2

moonpyk:
The to_float function was not interpreting unicode strings properly under Python 2. The function now uses the six.string_types method like to_int does

### DIFF
--- a/bin/weeutil/weeutil.py
+++ b/bin/weeutil/weeutil.py
@@ -1245,7 +1245,7 @@ def to_float(x):
     >>> print(to_float(None))
     None
     """
-    if isinstance(x, str) and x.lower() == 'none':
+    if isinstance(x, six.string_types) and x.lower() == 'none':
         x = None
     return float(x) if x is not None else None
 


tkeffer:
I guess I'm not seeing this. Under Python 3, `six.string_types` is a synonym for `str`. 

Can you give me an example of where the present code was failing?

moonpyk:
Oh deary me I misstyped, the function doesn't behave correctly on python 2 ! 

Try to configure your Season skin with line_gap_fraction = None. under WeeWX 4.1 on python 2, the function will fail with message "TypeError: float() argument must be a string or a number" under Python3 it works correctly like before. 

Or simpler test case : `
to_float(u'None') -> Error
`
If the string is coming from configobj under python2 we're gonna have an error.

Cheers,

tkeffer:
Thanks, Clément, for catching this. I added some regression tests.

moonpyk:
You're welcome, have a wonderful day :)

Issue #601, opened 2020-09-05
Modernize heat index


tkeffer:
Heat index was originally based on a chart from Steadman (1979), but there are now more modern ways of computing it.  See the formula offered by the NWS

https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml

PeterQuinn925:
This does look easy. Just to make sure I'm not missing anything - It's updating the heatindexF and heatindexC functions in wxfomrulas.py to use the equations in the link rather than what's there now. Right?

tkeffer:
That's it!

The link describes a two-step process, which I think is important to follow. If we're going to do it right, let's do it right!

When you're done, make sure you double check the tests for heat index (in `bin/weewx/tests/test_wxformulas.py`). Because the algorithm changed, the results will probably change. Make sure you're getting the correct result.

PeterQuinn925:
Ok. Done it. I don't know how to test the tests, but I've updated the code and the test results.

I have no intuitive understanding of heat index. I don't live in an area where it's a meaningful stat. I followed the recipe and tested it and I think it's right.

tkeffer:
Fixed in commit https://github.com/weewx/weewx/commit/4a0a9c1b488da5ac952735ddddfc25097aec02f4

tkeffer:
I'm a little suspicious of the results. Is it proper for heat index to be less than the sensible temperature? Take today. Outside temperature is a bit over 40, humidity about 80%. Here is today's plot of wind chill and heat index.

![image](https://user-images.githubusercontent.com/3321259/97062397-9026e580-154f-11eb-8ae7-564dcb42f79d.png)

The Wind Chill line is effectively the outside temperature (`outTemp`), so heat index is well below `outTemp` for long periods of time.

If you look at the [NOAA heat index calculator](https://www.wpc.ncep.noaa.gov/html/heatindex.shtml), it shows the same results.  But, the [heat index chart](https://www.weather.gov/safety/heat-index) does not go below 80F.

PeterQuinn925:
I guess I didn't do it right.

tkeffer:
No, you did good! The implementation is correct. Remember, NOAA's own heat index calculator gives the same results.

I'm just wondering if its appropriate for temperatures below 80°F.

duane-nepa:
I don't think it's appropriate for temperatures below 80F. Looking for a reference other than the NOAA graph and memory.

duane-nepa:
From Part 1.
[...]

PR #602, opened 2020-09-06
Modernize heat index #601

PeterQuinn925:


tkeffer:
I interpreted the [instructions](https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml) a little differently. They say

> In practice, the simple formula is computed first and the result averaged with the temperature. If this heat index value is 80 degrees F or higher, the full regression equation along with any adjustment as described above is applied.

I take this to mean that `HI` should be calculated first, then averaged with `T`. If the resultant value is `<80`, then you use `HI`. Otherwise, you use the full regression algorithm.

What do you think?

tkeffer:
Fixed in commit https://github.com/weewx/weewx/commit/4a0a9c1b488da5ac952735ddddfc25097aec02f4

PR #610, opened 2020-09-27
add ability to select the y-axis to label

brentfraser:


tkeffer:
Nicely done!

One last small task: put in a one or two line note in the change log (`docs/changes.txt`), mentioning the new feature, and referencing the issue and PR numbers. Follow the example of other changes. Put it under the V4.2.0 changes.

bobskisan:
I've tried to implement the scale on both sides of the graphs which works when I manually run wee_reports but the next time it updates automatically it reverts back to just the left side. I have been putting y_label_side = both in the Seasons skin.conf
    [[day_images]]
        x_label_format = %H:%M
        bottom_label_format = %x %X
        time_length = 97200 # 27 hours
        y_label_side = both
Is this the correct place to put it. If not where should it go? I have tried various other places with the same result.

Thanks
Bobski

tkeffer:
That would be a first.

That's the correct place to set the option.

Delete all your plots. Then set `debug=1`, then restart weewxd, then post the log from startup through the first reporting cycle. Once that's done, double check the resulting plots and see how they turned out.

bobskisan:
Somewhere along the process I have done a reboot and it works well after the reboot. Not sure what was happening.

Thanks
Bobski

PR #621, opened 2020-11-25
cc3000 udev rule doesn't work on Raspbian.

chaunceygardiner:
Raspberry Pi OS (Buster) needs the subsystem to be specified on this rule (peculiarly, the unmodified rules work fine on Debian Buster on my NUCs).


tkeffer:
I assume this version works on your NUC as well?

chaunceygardiner:
Yes, this version all works on the NUC.  Thanks for taking the change.

PR #626, opened 2020-12-26
expand directives in [[Calculations]] to allow bindings to be optionally

g-eddy:
specified (none specified = all bindings) e.g. mydata = software,archive

--as discussed in forum

### DIFF
--- a/bin/weewx/drivers/simulator.py
+++ b/bin/weewx/drivers/simulator.py
@@ -141,7 +141,10 @@ class Simulator(weewx.drivers.AbstractDevice):
     def trim_observations(self, stn_dict):
         """Calculate only the specified observations, or all if none specified"""
         if 'observations' in stn_dict and stn_dict['observations'] is not None:
-            desired = [x.strip() for x in stn_dict['observations'].split(',')]
+            desired = stn_dict['observations']
+            if isinstance(desired, str):
+                # convert comma-separated string to list
+                desired = [x.strip() for x in desired.split(',')]
             for obs in list(self.observations):
                 if obs not in desired:
                     del self.observations[obs]
--- a/bin/weewx/wxservices.py
+++ b/bin/weewx/wxservices.py
@@ -23,8 +23,18 @@ class StdWXCalculate(weewx.engine.StdService):
         """Initialize an instance of StdWXXTypes"""
         super(StdWXCalculate, self).__init__(engine, config_dict)
 
-        # Get the dictionary containing the calculations to be done. Default to no calculations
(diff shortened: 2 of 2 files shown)


tkeffer:
A few comments:

1. In general, you want to do one fix or change per PR. The fix for the simulator was dealt with in commit https://github.com/weewx/weewx/commit/d9a4daac4d63498126426c4b0e772654c1c505b1, so this PR cannot be accepted without affecting that.
2. Options parsed by `ConfigObj` are strings, unless they contain one or more commas, in which case they are a `list`. So,

         pressure = prefer_hardware

    will be parsed as a string, while

        pressure = prefer_hardware,
        pressure = prefer_hardware, loop

    will be parsed as lists. To deal with this problem, there is the function `weeutil.weeutil.option_as_list()`, which *always* returns a list. This way options can be treated consistently.
3. This means entries in `self.calc_dict` are either one- or two-element lists.
4. Make the decision on whether to apply the calculation in `new_loop_packet()` and `new_archive_record()`. If an observation type has a one-element list, apply the default, which is to do the calculation. If it's a two-element list, look at the second element to decide whether to apply the calculation.
5. This leaves `do_calculations()` almost untouched. You need only change this

        directive = self.calc_dict[obs]

    to this

        directive = self.calc_dict[obs][0]

6. Would suggest lower-casing the second element, so either `LOOP` or `loop` is accepted.

tkeffer:
Just thought of two other things:

7. The PR should be applied against the `development` branch. In general, new features belong in `development`, fixes in `master`.
8. As you suggested, it will need an entry in the User's Guide.

g-eddy:
but do_calculations() is responsible for walking the list of calculations, and the decision is made per-calculation.
for new_loop_packet() and new_archive_record() to make the per-calculation decision, walking the list will have to be moved from do_calculations() into both of them.
or, instead of new_loop_packet() passing ‘loop’ into do_calculations() and decision made in there, new_loop_packet() could apply a filter to the list of calculations and pass the filtered list to do_calculations

are you wanting me to refactor my code and resubmit?
(i guess i can undo a commit and roll forward again with your changes)
____________
Graham Eddy

tkeffer:
You're right. So, the decision must be made in `do_calculations()`. Still, I'd prefer to use lists, not strings. For one thing, you could get fooled by an observation name like `archive_temperature`.

Because you need to change branches, and because you'll need to untangle the simulator changes from the `wxservices` changes, it's probably best that we close this PR, and submit a new one.

On your end, you'll have to either undo the `master` commits from your copy, or do a new fork.

PR #627, opened 2020-12-26
Windchill bugfix (for unit system METRICWX)

edi-x:
Windchill is not calculated right with unit system METRICWX because the wind speed uses the unit mps but the current calculation treats the wind speed as kph.


tkeffer:
Thanks for spotting this. I modified your PR slightly so it was backwards compatible.
commit https://github.com/weewx/weewx/commit/064c97424d8b27d8eadaedc6eec1455d83d5a0a4

PR #630, opened 2020-12-28
alter StdWXCalculate to add optional loop,archive bindings to directives

g-eddy:
* hopefully i got branch, commits, etc correct this time. i replaced the fork and worked with that
* decision on which calculations to make have been moved to new_loop_packet() and new_archive_record(), by splitting the calculations list into two, worked out at configuration time
* lists instead of comma-separated strings used internally
* i haven't changed Manual as that is part of 4.3.0beta whereas i am thinking in terms of 4.2.0 as base
* i noticed during testing that ET returned a None value which was faithfully inserted into the packet. this is a minor defect outside scope of this change

### DIFF
--- a/bin/weewx/wxservices.py
+++ b/bin/weewx/wxservices.py
@@ -23,12 +23,43 @@ class StdWXCalculate(weewx.engine.StdService):
         """Initialize an instance of StdWXXTypes"""
         super(StdWXCalculate, self).__init__(engine, config_dict)
 
-        # Get the dictionary containing the calculations to be done. Default to no calculations
-        self.calc_dict = config_dict.get('StdWXCalculate', {}).get('Calculations', {})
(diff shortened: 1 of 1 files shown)


tkeffer:
You're on the right track. I've modified your PR, trying to simplify it, and make it a bit more "WeeWX-like." I've also added documentation to the User's Guide.

We cannot use f-strings, because they were introduced in Python 3.6, and we are targeting 2.7 and later, so I changed back to traditional formatting. 

Take a look and see if the changes are to your liking. Branch https://github.com/weewx/weewx/tree/g-eddy-development

Thanks! This will be a solid addition.

g-eddy:
fine with me. i’ll pick up the style and utilities (and that github thingy) over time. cheers
____________
Graham Eddy

PR #631, opened 2020-12-31
Daily average counts

edi-x:
Added aggregation types 
- avg_ge (number of days with average greater or equal to a given value) and 
- avg_le (number of days with average less or equal to a given value)


tkeffer:
What if `sumtime` is zero?

A tag to test, please, in `test_skins/index.html.tmpl`, along with expected values.

Documentation in the customizing guide.

edi-x:
@tkeffer , thanks for your feedback. I have implemented the requested changes. Unfortunately I was not able to run the test environment on my Windows machine because time.tzset() is not supported there und I had also problems with the logging environment. So I could not check, if the test results are OK. Is my assumption correct that a linux/unix environment is required to run these tests?

tkeffer:
The tests did not work because they used `sum_le` instead of `avg_ge` and `avg_le`. I corrected that. Also, the correct answer for `avg_le` is 80, not 81. The reason is that the very first record in the test database is actually in 2009 (by design), so you need to exclude that one when calculating results for 2010.

I also added expected results for the metric test.

Thanks again.

PR #632, opened 2021-01-03
Add line_gap_fraction to daytemp_with_avg example

PeterQuinn925:
This caught me. Trying to save someone else of the same problem.


tkeffer:
The example works now with commit https://github.com/weewx/weewx/commit/94ae1e6c4a4ebd1a53a1cf17dda095d9d88e4855. I don't think your annotations are necessary anymore.

Thanks for the PR, though!

PR #637, opened 2021-01-13
log configuration errors

bellrichm:
When running as a daemon/service and there configuration errors, WeeWX ‘just dies’ with nothing logged. This might happen when a person has WeeWX running, adds a service, and has an error in weewx.conf.
This pull request is more to explain the problem. Feel free to close and implement a better solution.


tkeffer:
this looks great. Thanks, Rich!

PR #643, opened 2021-01-24
make graph order configurable on Seasons skin

mrbalky:
I prefer a different order to the graphs on the default Seasons skin. This makes that order configurable in the `skins.conf` file. The default in `skins.conf` preserves the original order from `index.html.tmpl`.

One note is that this assumes that if the `$day` graph `has_data`, the `$week`, `$month` and `$year` graphs will also have data.

### DIFF
--- a/skins/Seasons/index.html.tmpl
+++ b/skins/Seasons/index.html.tmpl
@@ -6,41 +6,56 @@
 #encoding UTF-8
 ##
 
-#set global graphOrder = $getVar('Extras.Graphs.order', 'barometer,outsideTemp,feelsLikeTemp,outsideHumidity,wind,windDirection,windVector,rain,uvIndex,radiation,insideTemp,insideHumidity,extraTemp,signalQuality').split(',')
+#set global plotOrder = $getVar('Extras.Plots.order', 'barometer,outsideTemp,feelsLikeTemp,outsideHumidity,wind,windDirection,windVector,rain,uvIndex,radiation,insideTemp,insideHumidity,extraTemp,signalQuality').replace(' ', '').split(',')
 
-#def dataSettings($field, $image, $label)
-#return {"field": field, "image": image, "alt": label}
+#def plotSettings($hasData, $image, $label)
+#return {"hasData": hasData, "image": image, "alt": label}
 #end def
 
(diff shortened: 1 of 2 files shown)


tkeffer:
Clever, but I wonder if it won't create a number of support issues. Unless you understand the details of how Cheetah works (and you clearly do), it's not obvious how this works. 

Let me think about this a bit.

matthewwall:
consider adding a fallback value for the graph order in the .tmpl file - just in case nothing is defined in weewx.conf

mrbalky:
I'm working on a bit of a refactor to remove the assumption that if the day window has data all windows will have data. It'll include a default graph order as requested.

What is your specific worry about support? Perhaps I can address that directly in the refactor as well.

skins/Seasons/index.html.tmpl
```diff
@@ -44,116 +82,16 @@
...
-            #if $day.inHumidity.has_data
-            <img src="dayhumin.png"    alt="$obs.label.inHumidity" />
-            #end if
-            #if $day.extraTemp1.has_data or $day.extraTemp2.has_data or $day.extraTemp3.has_data
-            <img src="daytemp.png"      alt="$obs.label.extraTemp1" />
-            #end if
-            #if $day.rxCheckPercent.has_data
-            <img src="dayrx.png"        alt="$obs.label.rxCheckPercent"/>
-            #end if
+            $graphs('day', 'day')
```
mrbalky, line 85: This is strange, but the first 'day' is the '$day' object, and the second is the 'day' image prefix (as in "daytempdew.png").

tkeffer:
Nothing specific, but I know from experience that if people don't understand what they are looking at, they start asking questions. Sometimes, very dumb questions.

The original templates are wordy, but they are very simple to understand, and they look just like regular HTML, which many people are familiar with.

This new approach is heavily parameterized, which makes it easy to reorder, but it doesn't look at all like HTML. That leads to confusion.

mrbalky:
Ah OK. I guess that makes sense. Well, I added a bunch of comments so maybe that will help.

tkeffer:
I hate to discourage PRs, but after reflection, I'm going to say no. It's very clever, but it replaces the familiar HTML with unfamiliar Cheetah / Python code.

Issue #644, opened 2021-01-26
Improved handling of diacritics in `strict_ascii` encoded reports


gjr80:
When generating `strict_ascii` encoded reports the Cheetah Generator removes any characters from the report output that cannot be converted to an ASCII equivalent. This means that accented characters in, for example,  station locations in the NOAA format reports are removed and can result in unexpected results. This is a limitation of the existing Cheetah Generator where (for `strict_ascii` encoding) the Cheetah unicode output is simply encoded as ASCII and any errors ignored, ie `unicode_string.encode('ascii', 'ignore')`

A better approach would be to replace any accented characters with a non-accented equivalent. Whilst some detail is lost the intended result can usually be inferred from the replacement text.

[This weewx-user post](https://groups.google.com/g/weewx-user/c/I0FnvFdBlqk/m/bAgrDnNLAAAJ) refers.

tkeffer:
Fixed in commit https://github.com/weewx/weewx/commit/adaf4d920b67e2f587cdf27e330eca763b20e85c

PR #645, opened 2021-01-26
better handling of accented characters in strict_ascii encoded reports

gjr80:
Converts accented characters in `strict_ascii` encoded report output to non-accented equivalents. I believe there are better/more complete solutions but they involve another dependency. This solution is simple, utlilises an existing library and will handle most common cases. 

Tested against the current development branch under python 2.7 and python 3.7.

Refer issue #644.

### DIFF
--- a/bin/weewx/cheetahgenerator.py
+++ b/bin/weewx/cheetahgenerator.py
@@ -61,6 +61,7 @@ import datetime
 import logging
 import os.path
 import time
+import unicodedata
 
 import Cheetah.Filters
 import Cheetah.Template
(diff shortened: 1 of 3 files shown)


tkeffer:
Why not just change the encoding from `strict_ascii` to `utf8` for the NOAA reports? I don't see any formatting problems. The only place where UTF-8 would be used is in the station name, and that's in a left-justified line.

Lest your work be wasted, we could add another encoding option. Call it `normalized_ascii`. Someone might find it useful.

gjr80:
For some reason I had it in my mind that utf-8 caused some formatting issues with the NOAA format reports. If there isn't don't worry about this PR, not wasted time as it kept me from that test case :)

tkeffer:
No longer needed. Fixed in commit https://github.com/weewx/weewx/commit/adaf4d920b67e2f587cdf27e330eca763b20e85c.

Plus, added encoding `normalized_ascii` in commit https://github.com/weewx/weewx/commit/2d501adb80d968be091bb3722a75e8bb2250206f

PR #647, opened 2021-02-06
debian pkg installation doc clarification

vinceskahan:
This PR does a little minor tweaking of debian.htm to try to make it a little more obvious that there are a few steps in the 'configure apt' section that need to be done.   We saw again this morning on weewx-user that new linux users aren't quite getting the concept of 'add the key, add the repo' before doing the 'apt-get upgrade, apt-get install' steps.


vinceskahan:
I shouldn't be doing PR from my clone master, so I'll close this one and resubmit it off a feature branch.  Sorry.

PR #655, opened 2021-03-18
file hash additionally to file time for skip upload decision for FTP upload

roe-dl:
This refers to [thread "FTP Upload"](https://groups.google.com/g/weewx-development/c/p2_owWENsZc/m/gXykBH_nAQAJ).

Some files are created every archive cycle, but they do not change in contents. So it would not be necessary to upload them again. Calculating and saving a hash value of the file contents makes it possible to upload the file only if it is really changed. 

I checked that with my installation for one day, and the average upload duration is about 1/3 it was before.

### DIFF
--- a/bin/weeutil/ftpupload.py
+++ b/bin/weeutil/ftpupload.py
@@ -17,6 +17,12 @@ import time
 
 from six.moves import cPickle
 
+try:
+    import hashlib
+    has_hashlib=True
+except ImportError:
+    has_hashlib=False
+    
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


tkeffer:
Thanks for the PR. It looks very useful.

I should be able to take a closer look in the next few days.

tkeffer:
Looks good. Did you test it under Python 2 as well as 3?

roe-dl:
3 only. 

I have no installation with Python 2 in the moment, but I can set it up and do it on the weekend.

roe-dl:
I tested it with Python 2.7.17 unter Ubuntu 18.04 today. Worked as expected. No error messages. 

I changed the timestamp of some files that were not changed otherwise by the `touch` command to check. As expected they were not uploaded again, as the contents was the same.

So I consider that test successful

PR #659, opened 2021-04-03
Allow time period parameter for wee_database --reweight again (corrected parameter validation)

edi-x:
wee_database --reweight could not be used with the implemented parameter --date, --from and --to due to a wrong parameter validation. The mentioned parameter are now allowed again.


tkeffer:
Thanks! That's a useful addition.

PR #660, opened 2021-04-03
weewx-multi script corrections

edi-x:
I have updated the script to work correctly when you follow the instructions under https://github.com/weewx/weewx/wiki/weewx-multi.

I have to say that I am not an expert for these linux scripts. I tried to follow the instructions under https://github.com/weewx/weewx/wiki/weewx-multi to use two different drivers, but the services did not start. After my changes here everything seems to work correct on my Debian system. My changes where partially inspired by the changes committed by Matthew Wall (b11b4f8f7a71eb34bf40c98335de24cb54676be4) for the related Debian script weewx.debian.

I don't know how my changes could affect installations at other Linux distributions. However I want to share my changes here and ask for a review.


matthewwall:
superceded by commit 732c1849

we might want to change to WEEWX_BIN at some point (all of the other scripts in init.d use WEEWX_BIN, but multi uses WEEWX_BINDIR), but not just yet.  the intent is to replace the singleton init scripts with a multi version so that every weewx installation is a weewx-multi, with a default install degenerating to the singleton case.

however, we also need to install systemd unit files on systems with systemd, and those units need to be multi-aware.  i prefer to do the changeover all at once in some future weewx release.

Issue #556, opened 2020-05-23
String encoding issue when using Python 3 for WS1 driver


windcrusader:
https://github.com/weewx/weewx/blob/85525a230f538dff80713f200da668754a41a532/bin/weewx/drivers/ws1.py#L252

I tried to upgrade my system to Python 3 today and I can across what appears to be a simple encoding error in the WS1 driver when using Python 3. In Python 3 when the serial data is pulled in from the device it encodes it as a byte string e.g b"xxxxx". The checks for a valid string in the driver don't account for this. 

I managed to fix this in my version by changing the referenced line to:
`buf = self.get_readings().decode("utf-8")`

However, I feel that perhaps this is not a general fix, as it may break the driver when it is used in Python 2. I'm sure there is a better way to handle the byte strings that will be robust across both versions. Can someone please advise, then I'll submit a pull request with the fix.

windcrusader:
Actually I did a brief check on a terminal session of Python 2.17.16 and seems that calling .decode() won't have any negative consequences; at least I think so. 

If this is agreed, I'll make the pull request to fix the issue.

tkeffer:
Thanks, Brad, for finding this and, even better, figuring out a solution.

However, I would prefer changing the comparison at line [143](https://github.com/weewx/weewx/blob/85525a230f538dff80713f200da668754a41a532/bin/weewx/drivers/ws1.py#L143), rather than do a decode. The reason is that the data coming off the station really are bytearrays, not strings. So, the line becomes

            if buf[0:2] != b'!!':

There may be other lines with similar problems --- I didn't check. I can also see that the TCP version has similar problems.

See the Wiki article [*Supporting Python 3*](https://github.com/weewx/weewx/wiki/Python-3-strategy) for a general discussion.

Thanks again!

Issue #372, opened 2019-02-11
Running ADS-WS1 getting error 'unexpected buffer length'


jezkerwin:
I'm running a Argent Data ADS-WS1 weather station on a raspberry Pi.

When looking at /var/log/messages, I see the following errors and then after a while the engine shuts down.
Occasionally a successful reading happens, but for the most post it's these messages.

`Feb 11 21:11:04 weewx[6897]: ws1: Failed attempt 1 of 5 to get readings: Unexpected buffer length 51
Feb 11 21:11:31 weewx[6897]: ws1: Failed attempt 2 of 5 to get readings: Unexpected buffer length 594
Feb 11 21:12:11 weewx[6897]: ws1: Failed attempt 1 of 5 to get readings: Unexpected buffer length 730
Feb 11 21:12:21 weewx[6897]: ws1: Failed attempt 2 of 5 to get readings: Unexpected buffer length 102
`

Does anyone know what this means, and how to fix?
Thanks.

jezkerwin:
Here is an excerpt from /var/log/syslog

```
Feb 11 21:18:09 . weewx[7001]: restx: AWEKAS: Posting not enabled.
Feb 11 21:18:09 . weewx[7001]: engine: Finished loading service weewx.restx.StdAWEKAS
Feb 11 21:18:09 . weewx[7001]: engine: Loading service weewx.engine.StdPrint
Feb 11 21:18:09 . weewx[7001]: engine: Finished loading service weewx.engine.StdPrint
Feb 11 21:18:09 . weewx[7001]: engine: Loading service weewx.engine.StdReport
Feb 11 21:18:09 . weewx[7001]: engine: Finished loading service weewx.engine.StdReport
Feb 11 21:18:09 . weewx[7001]: engine: Starting up weewx version 3.9.1
Feb 11 21:18:09 . weewx[7001]: engine: Station does not support reading the time
Feb 11 21:18:09 . weewx[7001]: engine: Starting main packet loop.
Feb 11 21:18:09 . weewx[7001]: manager: Daily summary version is 2.0
[...]

PR #664, opened 2021-04-15
WS1 driver TCP mode issues

mjuniper:
Fix encoding errors, simplify data fetching, address unexpected buffer length errors.

I am using Python 3.

I've been using the WS1 driver in serial mode with no issues for a while. But when I switched to TCP mode it did not work at all.

I have my WS1 connected to a raspberry pi zero and exposed to the network with ser2net with this config:

`2000:raw:600:/dev/ttyUSB0:2400`

And I have weewx running on a pi 4.

When I enabled TCP mode, I ran into a couple of issues:

1. String encoding errors similar to #556 
2. Unexpected buffer length errors (like #372)

Regarding 2, in my testing here, the WS1 never seems to send more than 6 bytes at a time via ser2net. So when we do [this](https://github.com/weewx/weewx/blob/master/bin/weewx/drivers/ws1.py#L363) and expect to get the amount of data we asked for, we will be disappointed. Same thing [here](https://github.com/weewx/weewx/blob/master/bin/weewx/drivers/ws1.py#L394).

So what I would see is it would go along for a while getting data then about every minute or so, it would hit [this](https://github.com/weewx/weewx/blob/master/bin/weewx/drivers/ws1.py#L142) and that would cause it to wait 10 seconds before trying again.

This leads me to wonder, is nobody else using TCP mode or is there something funky about my setup here?

In any case, I rewrote some of the TCP mode stuff to be much simpler (at least to my eye). It works flawlessly here - it's been running for several hours. But I am worried that this might break for other people.

Feedback is welcome - I am not a Python developer, I write javascript for a living :)

Also, I notice that the [docs](http://weewx.com/docs/usersguide.htm#[WS1]) are out of date. They do not mention TCP mode but they do mention polling interval. I notice that polling interval was taken out of the WS1 driver some time ago. But for my purposes, using TCP mode, it would be useful. I am wondering, why was that taken out?

### DIFF
--- a/bin/weewx/drivers/ws1.py
+++ b/bin/weewx/drivers/ws1.py
@@ -307,6 +307,7 @@ class StationSocket(object):
         import socket
 
         log.debug("Connecting to %s:%d." % (self.conn_info[0], self.conn_info[1]))
+
         for conn_attempt in range(self.max_tries):
             try:
                 if conn_attempt > 1:
@@ -515,8 +516,8 @@ if __name__ == '__main__':
         exit(0)
 
     if options.debug:
-        weewx.debug = 1
-        DEBUG_READ = 1
+        weewx.debug = 2
+        DEBUG_READ = 2
 
     weeutil.logger.setup('ws1', {})
 
--- a/bin/weewx/drivers/ws1.py
+++ b/bin/weewx/drivers/ws1.py
@@ -328,34 +328,37 @@ class StationSocket(object):
                       % (self.conn_info[0], self.conn_info[1], ex))
             raise weewx.WeeWxIOError(ex)
 
-    def get_data(self, bytes=8, encoding="utf-8"):
+    def get_data(self, bytes=8):
         import socket
         try:
-            return self.net_socket.recv(bytes, socket.MSG_WAITALL).decode(encoding)
-        except (socket.error, socket.timeout) as ex:
+            data = self.net_socket.recv(bytes, socket.MSG_WAITALL)
+        except Exception as ex:
             raise weewx.WeeWxIOError(ex)
+        else:
+            if len(data) == 0:
+                raise weewx.WeeWxIOError("No data recieved")
+
+            return data
 
     def find_record_start(self):
         if DEBUG_READ >= 2:
             log.debug("Attempting to find record start..")
 
-        buf = ''
+        buf = bytes("", "utf-8")
         while True:
             data = self.get_data()
-            if len(data) == 0:
-                raise weewx.WeeWxIOError("No data recieved")
 
             if DEBUG_READ >= 2:
                 log.debug("(searching...) buf: %s" % buf)
             # split on line breaks and take everything after the line break
             data = data.splitlines()[-1]
-            if "!!" in data:
+            if b"!!" in data:
                 # if it contains !!, take everything after the last occurance of !! (we sometimes see a whole bunch of !)
-                buf = data.rpartition("!!")[-1]
+                buf = data.rpartition(b"!!")[-1]
                 if len(buf) > 0:
                     # if there is anything left, add the !! back on and break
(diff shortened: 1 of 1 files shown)


tkeffer:
Thanks, Mike, for taking the time to submit such a useful PR.

A few comments and questions:
1. Did you try it under Python 2.7?
2. Line 409, I think you meant `buf`, not `buff`
3. Line 398. You encode the buffer, converting it to unicode. But, method `ValidateString()` requires a byte string.
4. Lines 276-277: variables `ip_addr` and `ip_port` are not used.
5. If the driver is run directly, it will now only work with sockets. Perhaps the parser could be given an option of whether to use a serial port, or TCP port?
6. Bump the version number. Say, to v0.5?
7. There are quite a few similarities between this driver and the ultimeter driver. Take a look at the [latest version of the latter](https://github.com/weewx/weewx/blob/master/bin/weewx/drivers/ultimeter.py) to see how it solved some of the decode problems.


Thanks again! It's super useful having this! And, you're writing Python code like a native.

mjuniper:
* [ ] Did you try it under Python 2.7?
* [ ] Line 409, I think you meant `buf`, not `buff`
* [ ] Line 398. You encode the buffer, converting it to unicode. But, method `ValidateString()` requires a byte string.
* [ ] Lines 276-277: variables `ip_addr` and `ip_port` are not used.
* [ ] If the driver is run directly, it will now only work with sockets. Perhaps the parser could be given an option of whether to use a serial port, or TCP port?
* [ ] Bump the version number. Say, to v0.5?
* [ ] There are quite a few similarities between this driver and the ultimeter driver. Take a look at the [latest version of the latter](https://github.com/weewx/weewx/blob/master/bin/weewx/drivers/ultimeter.py) to see how it solved some of the decode problems.

mjuniper:
Thanks Tom. I'll get on this stuff. I do have one area of uncertainty right off the bat:

>     3. Line 398. You encode the buffer, converting it to unicode. But, method `ValidateString()` requires a byte string.

I actually thought that line was converting it to `bytes`!

When I poke at this in the terminal I see this:

>>> type("foo")
<class 'str'>
>>> type(b"foo")
<class 'bytes'>
>>> type("foo".encode())
<class 'bytes'>
>>> type(bytes("foo", "utf-8"))
<class 'bytes'>

What should I be doing there?

I may have more questions as I go through this.

tkeffer:
You're right. It does convert back to bytes. I missed that you were using `encode`, not `decode`. 

But, somehow, it doesn't feel right. I was hoping you could clean it up! Data coming off a serial port will be in bytes, and should be manipulated as such. Especially encoded data like this.

The ultimeter driver had the same problem. I converted everything to use bytearrays and bytes. It also helps to document the expected parameter types for each function.

No problem if you don't have the time.

mjuniper:
> Did you try it under Python 2.7?

I'm not exactly sure how I would do that. I tried just invoking it with `PYTHONPATH=/usr/share/weewx python /usr/share/weewx/weewx/drivers/ws1.py` but that did not work - I assume because I don't have the necessary dependencies for Python 2. Any suggestions?

tkeffer:
Good question. Since 2.7 was deprecated, the Python Foundation not only no longer supports it, but they have also been removing all traces from depositories and PyPi. They've even modified old tools (e.g., `get-pip`) so they no longer work. 

I can think of only two approaches that might still work:
1. Set up a VM running an older version of Debian (or Ubuntu, or Mint, or ...).
2. Use [`pyenv`](https://github.com/pyenv/pyenv). It's invaluable for managing multiple versions of Python. However, I don't expect the average user to use it. Frankly, it's been so long since I've set up my `pyenv`, I'm not sure that it's still possible.

I will see what I can do on my end. I don't have a WS1. However, it's a simple protocol, so I might be able to emulate it using mock.

Issue #248, opened 2017-03-29
how to merge newskin into weewx master


matthewwall:
what else does the newskin branch need before we merge it into master?

this is what it does as of 29mar17 (see the newskin branch):

* Standard is replaced by an entirely new skin
* the mobile and smartphone skins have been split out into separate skins
* defaults for all skins can be specified in [StdReports]

the branch includes the changes for taking advantage of the defaults in StdReports.  for example, when you set units using wee_config, those units apply to *all* reports.  then you can customize per-report if desired.

i see at least these issues:

1) how to incorporate the new standard skin.  should it be installed in every upgrade?  if so, then using what name (so it does not clobber StandardReport)?  or should users get the new skins only if they do a clean install?

2) what other features/functionality are necessary in the new standard skin before it is ready for release?

3) what to call the new standard skin?

tkeffer:
Looks like the merge conflicts were never resolved. If I run it, I get:

~~~
Traceback (most recent call last):
  File "./bin/wee_reports", line 15, in <module>
    import weewx.station
  File "/home/tkeffer/git/weewx/bin/weewx/__init__.py", line 9
    <<<<<<< HEAD
     ^
SyntaxError: invalid syntax
~~~
Easy enough to fix. Pressing on ...

I get the error

~~~
Mar 29 18:12:08 nuc wee_reports[10232]: cheetahgenerator: Generate failed with exception '<class 'NameMapper.NotFound'>'
Mar 29 18:12:08 nuc wee_reports[10232]: cheetahgenerator: **** Ignoring template /home/weewx/skins/Standard/index.html.tmpl
Mar 29 18:12:08 nuc wee_reports[10232]: cheetahgenerator: **** Reason: cannot find 'extraTemp1' while searching for 'obs.label.extraTemp1'
Mar 29 18:12:08 nuc wee_reports[10232]: ****  Traceback (most recent call last):
Mar 29 18:12:08 nuc wee_reports[10232]: ****    File "/home/tkeffer/git/weewx/bin/weewx/cheetahgenerator.py", line 329, in generate
Mar 29 18:12:08 nuc wee_reports[10232]: ****      print >> _file, compiled_template
Mar 29 18:12:08 nuc wee_reports[10232]: ****    File "/usr/lib/python2.7/dist-packages/Cheetah/Template.py", line 1005, in __str__
Mar 29 18:12:08 nuc wee_reports[10232]: ****      rc = getattr(self, mainMethName)()
[...]

PR #665, opened 2021-04-17
Make Internationalization/Localization Easier

roe-dl:
This refers to the thread about [localization](https://groups.google.com/g/weewx-development/c/MOS1PGFUzzM) and the thread about [include files](https://groups.google.com/g/weewx-user/c/H_YfJsNT_Jk)

I thought about how to make localization less difficult. I made a suggestion in the localization thread mentioned above. In order to show the possibilities I created this PR and edited the Seasons skin a little bit.

If you set "lang = en" in the skin section of weewx.conf you will see the skin in English language. If you set "lang = de" you will see the skin in german language. No further changes are necessary within the configuration files.

The localization is not complete in this example. It is proof of concept, only. If you like it, I can further work on it.

The solution makes it possible for users to easily provide localization files for their language that can be included in WeeWX. No knowledge of HTML is needed for that. They only need to copy the file `en.conf` and to translate the labels within the file.

Advantages:
* Change localization by changing one single entry in weewx.conf, only.
* No need to change template files of the skin for localization.
* Skin authors can ship their skin with multiple localizations included.

What is missing by now:
* Localization is not complete for Seasons skin
* page dependent labels to prevent "if - else if - else if - else" cascades in template files
* may be dividing the localization file into a global part for observation type labels and a skin-specific part for skin-specific labels

What do you think?


tkeffer:
Very nice! Thanks for getting the ball rolling.

There are some other approaches, which I have been thinking about. 

1. First, we can use something similar to the existing Python utility [`gettext`](https://docs.python.org/3/library/gettext.html), the standard way of localizing Python programs. In this case, we are not localizing a Python program, but rather a Cheetah template, but we can apply the same idea. It would look something like
     ```
    <p>$gettext("Outside temperature is") $current.outTemp</p>
     ```
    where `$gettext()` is actually a search list extension. The target language would be set in `weewx.conf`. The result would be HTML files in the local language. 

2. The other way to approach the problem is at runtime. This would allow switching languages in the final HTML. We do this by leveraging one of the existing Javascript localization libraries, such as [Globaliza](https://github.com/globalizejs/globalize), or [polyglot.js](https://github.com/airbnb/polyglot.js).

matthewwall:
thank you, thank you, thank you for working on this!  here are some usage scenarios:

a) single report in a single language.  skin developer ships skin X with N different languages.  user specifies 'es' to get that skin in spanish, 'fr' for french, etc.  there is only ever one language emitted by the report generator, even though the skin developer has included multiple translations.

b) multiple reports from single skin, each report in different language.  skin developer includes multiple translations.  report generator creates report in each language that user specifies, not just a single language.

also, it is important to distinguish localization (l10n) (basically number/date formatting) and internationalization (i18n) (basically language).  just because i live in the us does not mean that i want typical us i10n when i specify us-en for l18n.

roe-dl:
First I have to admit I meant internationalization (i18n). I should change the title.

Some Thoughts.

@matthewwall 
It was case a) that I had in mind when I wrote the code.

To get multiple languages at the same time I thought about having multiple report sections in StdReport pointing to the same skin directory but to different languages and different target directories. That is a little bit like your case b)  but it is not the same.

@tkeffer 
My work was based on the target to embed the solution into the structure WeeWX is already using. I would like to consider the alternatives without a clear opinion about them:

**gettext**

advantages:
* well known library
* templates are more readable (at least for English speaking users)

disadvantages
* more difficult to edit, especially for unexperienced users. Need to learn how to use the tools to create those binary files used by gettext library
* does not correspond to the methods WeeWX uses for configuration so far (which is configObj)

**Javascript solution**

Basically the files are created every 5 minutes. The values only update when the file is re-created (except for special solutions like Belchertown skin). 

I guess it should be possible to create internationalized files without the need of JavaScript. Not all files created are HTML. Not all files created are displayed in browsers.

tkeffer:
Sorry. I should have been more clear. I'm not saying that we should literally use the Python `gettext` facility, but rather that we should take inspiration from it, and do lookups based on natural language keys. 

Maybe it's just me, but I find

    $gettext("Current Conditions")

more readable than

    $obs.label.current_conditions_title

It also lends itself well to a default, should the localization fail to provide a value.

roe-dl:
I looked into the documentation of configObj, and I guess keys including whitespace are possible, and thus `$gettext("Current Conditions")` is, too. I will check how to include it.

roe-dl:
As I worked on it today there is another commit reflecting the discussion. I included a `$gettext()` tag. It takes one or two arguments:
* key to look up in the localization/internationalization file
* optional: name of the subsection

Alle the keys used by `$gettext` are located in the `[Texts]` section of the localization/internationalization file.

As some skins contain of a lot of files, it may be difficult to find a specific key in the file. Therefore it is possible to have subsections in the `[Texts]` that are named like the sections in skin.conf that define a template. There is a special tag `$page` that I introduced which holds the section name.

examples:

* `$gettext("Current Conditions")` looks for ['Texts']['Current Conditions'] 

* Assuming we are creating statistics.html defined in section `[CheetahGenerator][[ToDate]][[[statistics]]]` then `$gettext("Title",$page)` looks for ['Texts']['statistics']['Title'].

I changed the "Current Conditions" label and the "High/Low" label to show how it works. And I added `$gettext("Sun",$page)` and `$gettext("Moon",$page)` in celestial.inc to show how subsections work.

roe-dl:
I wonder whether we should call the tag `$text`, only, instead of `$gettext` as the double t in the middle is more likely for typos. And `$text` would point to the section `[Texts]` in a similar manner as `$obs.label` does to `[Labels]`.

tkeffer:
In the Java / C / C++ world, people usually use an underscore as an alias. Something like (NOT TESTED):

```
#set _=$gettext

_("Current temperature") $current.outTemp
```

roe-dl:
I will try that whether it works with Cheetah.

Regarding the observeration type labels, nothing is changed from the user's point of view. They use `$obs.label.outTemp` as before. There is no need to change existing skins for that. It works automatically. They only need to provide the localization file.

roe-dl:
It is possible to use `$_` as an alias for `$gettext`. So with that last commit `$gettext`, `$text`, and `$_` can be used alternatively. If the text does not include whitespace, a dot can be used instead of parentheses. It is a matter of taste sometimes, and the skin author can use, what they find more logically.

Examples:
* `$gettext("Current Conditions")` is the same as `$text("Current Conditions")` is the same as `$_("Current Conditions")`
* `$gettext("Day")` is the same as `$_("Day")` is the same as `$_.Day`
* `$obs.label.outTemp` uses the language setting automatically. No changes to the skin needed.

tkeffer:
This looks very encouraging.

A few comments.

1. I know I suggested `$_()`, but, I'm taking it back. We should follow the [Zen of Python](python.org/dev/peps/pep-0020/) rule "There should be one-- and preferably only one --obvious way to do it." That means, use `$gettext()` only in all skins that come with WeeWx. 
2. Along the same line, do we need `lang_file` _and_ `lang`? Why not just `lang`, and assume that it will be in a subdirectory `lang`?
3. Assuming `lang`, it should be extracted from the merged `skin_dict`, not `config_dict`. This way, it can obey the hierarchical structure that other options use. 
4. In general, the argument to `$gettext()` should be the default value. So, for example, use
    ```
      $gettext("About this weather station")
    ```
    not
    ```
      $gettext("About")
    ```
    There are many cases like this throughout the skin. That means, in theory, the entire `en.conf` file can go away.
5. Needs some documentation in `docs/customizing.htm`. Do you best, and I'll fix up any English issues.    

This looks great, Karen, and is a sorely needed feature.

-tk

tkeffer:
Regarding my point 4 above, one other comment: there is no reason why a skin could not be written in German. Then make the English users translate it! :-)

roe-dl:
I had one comment to point 4.

Consider Altitude. It is the same word if you mean the altitude of some location and the altitude of some celestial body in the sky. This is true for both English and german. But we do not know, how it is in other languages. There can be as well different words. So I divided the key into "GeoAltitude" for the Altitude of the location and "AstroAltitude" for the Altitude of sun and moon. So it is possible to translate it to different words. This is the reason why it is not always possible to use the real Englisch text as key.

tkeffer:
Good point, and there will be cases where the lookup key cannot be used as the default. 

The original `gettext()` solves this problem by using "contexts", and the special function `pgettext()`.
```
pgettext("location", "Altitude")
pgettext("celestial", "Altitude")
```
But, your approach is simpler.

roe-dl:
regarding point 1: Shall I remove `$_()` all together or in the skin only?

tkeffer:
Remove `$_()` entirely. 

We need to set clear examples in the skins that are included in WeeWX. Nothing tricky.

roe-dl:
Point 2+3 are done as suggested. 

One additional question to point 1: I personally would prefer `$text()` over `$gettext()` because it is shorter and more to the point for my understanding. Would that be OK for you or would you prefer `$gettext()`? As you are "the boss" I would do what you say here.

tkeffer:
I think I'd prefer `$gettext()`, then alias it to something shorter in the skin if you must. `$text()` wouldn't mean anything to anyone.

One other thing: can you do your PR relative to the `development` branch? That's where new features go.

tkeffer:
I think the only thing we are missing is the documentation...

roe-dl:
I think I will do it the next few days. When do you need it?

tkeffer:
We are a volunteer organization. Things happen when they happen. So, not in a hurry.

roe-dl:
Is it true that there can be a section `[Defaults]` (with *single* brackets) in weewx.conf? I only know it with double brackets in the `[StdReport]` section.

roe-dl:
Live data of the Seasons skin using `$gettext` (temporary links):
* [german METRIC](https://www.woellsdorf-wetter.de/Seasons/de/index.html)
* [English METRICWX](https://www.woellsdorf-wetter.de/Seasons/en/index.html)
* [English US](https://www.woellsdorf-wetter.de/Seasons/en-US/index.html)

The settings in `weewx.conf` are:
```
    [[SeasonsReportExt1]]
        skin = Seasons
        lang = de
        target_unit = METRIC
        HTML_ROOT = .../Seasons/de

    [[SeasonsReportExt2]]
        skin = Seasons
        lang = en
        target_unit = METRICWX
        HTML_ROOT = .../Seasons/en

    [[SeasonsReportExt3]]
        skin = Seasons
        lang = en
        target_unit = US
        HTML_ROOT = .../Seasons/en-US
```

tkeffer:
Hang on. Let's stick with internationalization and not wander into changing units. There are all kinds of potential conflicts that can arise if the present unit group system is specified, along with your `target_unit`. Good PRs focus on just one topic.

I think we are also going to need some documentation for the Customizing Guide. At a minimum, how to localize a skin, but it would also be nice to discuss writing an I18N skin.

Lot of work, eh? Welcome to my world! :-)

tkeffer:
> Is it true that there can be a section `[Defaults]` (with _single_ brackets) in weewx.conf? I only know it with double brackets in the `[StdReport]` section.

There was an experimental version that used `[Defaults]`, but it was not around very long. Now it's just `[StdReports]/[[Defaults]]` See issue #248 if you're interested.

roe-dl:
The unit thing refers to the last paragraph of @matthewwall's [comment](https://github.com/weewx/weewx/pull/665#issuecomment-821825249).

And I personally was annoyed when I set up WeeWX and had to add all the individual unit groups to the `[[[Units]]][[[[Groups]]]]` section. So I wished to have a simple possibility to refer to one of the predefined unit systems. I considered that part of the localization.

Regarding documentation: I started with the user's guide, because it was the less to do there. But I have in mind that I have to update the customization guide, too.

tkeffer:
I get your point and, in retrospect, I wish I had originally designed weewx around just a simple set of choices, US, or Metric, then allowed overrides for individual observations --- no unit groups. Unit groups inject another level of indirection, which makes things complex. 

But, it works, and is very flexible. I worry that your approach adds yet another level of overrides onto an [already complex system of overrides](http://www.weewx.com/docs/customizing.htm#How_options_work). And, it is too broad. Indeed, you felt the need to add a new unit system, `METRICDE`, whose changes can easily be accommodated already by the present unit group architecture.

Finally, the installation process allows users to select between `US` and `METRICWX`, then automatically sets up appropriate selections for the unit groups. So, the simple approach is already covered.

roe-dl:
Yes, I know, backwards compatibility can be a burden.

roe-dl:
May you want to comment on the changes of the user's and the customization guide? Is there something important missing?

roe-dl:
> [...] makes things complex.
> 
> Finally, the installation process allows users to select between `US` and `METRICWX`, then automatically sets up appropriate selections for the unit groups. So, the simple approach is already covered.

May I say some thoughts to that? 

I remember some warning in the installation process against changing the unit system to some other than `US`. I personally did not change that, and I did not change the `[StdReport][[Defaults]]` section because I did not want to loose the original English definitions. When I did so, I thought about setting up an English version of my website, too. So, at least two different languages are used in parallel.

Some extensions define new unit groups and units. The `target_unit` option in the skin configuration regards them automatically. The `[StdReport][[Defaults]]` as well as the `[[Units]][[[Groups]]]` section of skin configuration do not.

I guess most users can use one of the predefined unit systems and have to change one or two units against them only. So the `target_unit` option together with some (not many) lines in the `[[Units]][[[Groups]]]` section are enough. If you want to reduce complexity, you may want to declare `[StdReport][[Defaults]]` deprecated and remove it in some later version.

Additionally you could get the initial defaults in `_build_skin_dict()` from `units.py` and remove `defaults.py` entirely. For my opinion that would make maintaining the sources easier, as unit definitions would concentrate in `units.py` only.

> Indeed, you felt the need to add a new unit system, `METRICDE`, whose changes can easily be accommodated already by the present unit group architecture.

Your are right. I removed that.

tkeffer:
The [warning about changing unit systems](http://www.weewx.com/docs/usersguide.htm#StdConvert) is for changing the system used in the _database_, not reports. Indeed, it stresses that reporting units are completely independent from the database units. Perhaps some of the confusion is coming from your choice of the name `target_unit`, which is the name used for selecting the database units. A better choice might have been `reporting_unit` or something like that.

> Some extensions define new unit groups and units. The target_unit option in the skin configuration regards them automatically. The [StdReport][[Defaults]] as well as the [[Units]][[[Groups]]] section of skin configuration do not.

I don't see the difference. Your `target_unit` approach requires that a unit be assigned to the new observation type for each of the unit systems. The unit group approach requires that a unit group be assigned. 

It would certainly be possible to remove `defaults.py` in favor of adding them into `units.py`. Don't see the advantage, but it is certainly possible. 

Anyway, I assume you are done? I will take a look.

tkeffer:
I hope you don't mind my comments: you are proposing one of the biggest PRs we have had in a while, and one that directly faces the user. So, we need to get it right. 

A few specific comments regarding `gettext.py`:
1. Most important: rather than search lots of different dictionaries for a key, a better approach is to honor a hierarchical approach and use one of the utilities `accumulateLeaves()` or `search_up()`.
2. None of these imports are actually used:
    ```
    from weewx.units import ValueTuple,ValueHelper
    import configobj
    import weeutil.weeutil
    import weeutil.config
    import os
    import os.path
    ```
3. Because `gettext.py` is not an extension and will be part of weewx itself, there is no reason to check for V3 style logging. You can assume V4 logging.
4. I really don't follow the logic of `Gettext.__init__()`. A few comments would be in order.
5. Variable names like `d`, `ii`, `jj`, etc., are not helpful. Something more descriptive, please.
6. There is no point in using `__getattr__`, then testing for static names like `nav`, `page_header`, etc. Why not just make them properties of `FilesBinder`?

That's all for now.

roe-dl:
> Perhaps some of the confusion is coming from your choice of the name `target_unit`, which is the name used for selecting the database units. A better choice might have been `reporting_unit` or something like that.

I found two option names already used in weewx.conf. One was `target_unit` and one was `unit_system`. The latter is used by the MQTT services. Considering your comment I will remove `target_unit` and let `unit_system` there. 

Your other comments I will check later. In the moment I check the compatibility with Python 2.7.

roe-dl:
Some answers:

1. I am not familiar with those functions. I will learn about them and then try to use them.
2. I commented them out. Nothing bad happened. So I can remove them.
3. Done.
4. I put things that can be done at init time into `__init__()`. And I put things that change with every call into `get_extension_list()`. The language code is needed for `<html lang="...">` tag. So I retrieve it from the dict. Because, in theory, users could set a whole file name in the `lang` option, path if any is removed there. The next statements check for the existence of `[Labels][[Generic]]` and `[Texts]` section in skin_dict and remember them for further use. I added some comments.
5. I will change the names.
6. I will have to look into it. Because you commented it, I guess, it can be improved.

roe-dl:
2. Removed.
5. Done.

Python 2.7 was working without errors.

roe-dl:
Regarding `__init__()`. May I ask a question? 

At the moment I check if `[Labels][[Generic]]` and `[Texts]` section are present and set two variables that represent those sections.

If I consider your comments, I think, alternatively I could do a deep copy of `self.generator.skin_dict` and add the sections there if they do not exist. Then I only have one variable `self.skin_dict`. But I need a deep copy because otherwise I would change the original `self.generator.skin_dict` which I certainly should not do.

I am not sure whether that is better than the previous solution.

tkeffer:
If you look in the function `StdReportEngine._build_skin_dict()`, you'll see it already does a deep copy, then merges `skin.conf` and `[StdReports]//[[Defaults]]`. So everything is already in there. You just need to merge in your text configuration file `de.conf` on top of the mix.

Incidentally, let's assume that the language text files are always under the subdirectory `lang`, and are called `de.conf`, `en.conf`, etc. So, given a language, you always know where to find its configuration file. The stuff around lines 80 in `gettext.py` goes away.

BTW, I've sent you an invite to Slack. It's a much better medium for a lot of these meta-discussions.

roe-dl:
When it comes to `gettext.py` the language file (`en.conf`, `de.conf` etc.) is already merged into skin_dict. No files are read in `gettext.py`. It only provides the `$gettext` tag.

But I could make sure within `_build_skin_dict()` already, that the sections `[Labels][[Generic]]` and `[Texts]` are present (empty if there are no definitions). Then the checks in `gettext.py` could be removed. If you want me to do that that way, I will do that.

Regarding the `lang` tag: From my experience users always do what they should not do. I had a boss long years ago, and he often said: "What is possible that will always happen some day." He taught me to include all possible checks. But I can remove it, if you want.

roe-dl:
> BTW, I've sent you an invite to Slack. It's a much better medium for a lot of these meta-discussions.

I am no sure, I found that invitation.

roe-dl:
You teach me into better code. :-)

Made some improvements to the code. Found a way to remove two code sections in `Gettext.__init__()`. Did some tidying up.

tkeffer:
> Regarding the lang tag: From my experience users always do what they should not do. I had a boss long years ago, and he often said: "What is possible that will always happen some day." He taught me to include all possible checks. But I can remove it, if you want.

Checks are good. Multiple modes for an option are not. Unless I'm misunderstanding what you are doing, it looks like the user can specify either

    lang = de

or

    lang = lang/de.conf

I am saying we should offer only the former. Python Zen: _There should be one-- and preferably only one --obvious way to do it._


> When I wrote the code for the lang tag, I looked how the skin option is implemented in WeeWX. There is SKIN_ROOT, and the directory of the skin is in there. But the user is free to put an absolute path there and put the skin files anywhere on his/her computer. So I thought I follow the rules if the directory for the language files is relative to the skin directory, but the user is free to set an absolute path. So you know how I came to that.

Then offer a separate option `LANG_ROOT`, relative to `SKIN_ROOT`, which defaults to `./lang`.

roe-dl:
> Checks are good. Multiple modes for an option are not. Unless I'm misunderstanding what you are doing, it looks like the user can specify either ...
> I am saying we should offer only the former. Python Zen: _There should be one-- and preferably only one --obvious way to do it._

It's not me. It is simply how `os.path.join()` works. I cannot prevent the user from putting an absolute path there without a lot of code.

It is the same with the `skin` option. There a single name should be, a directory relative to `SKIN_ROOT`, but nothing prevents the user from putting an absolute path there.

> Then offer a separate option `LANG_ROOT`, relative to `SKIN_ROOT`, which defaults to `./lang`.

Do we really need that? I think it is not necessary.

tkeffer:
You've lost me. No matter. Just a simple `lang` option please, with one, simple, mode.

roe-dl:
OK, no problem.

roe-dl:
I removed all the code in `Gettext.__init__()` that you did not understand. Thank you for pointing me to that. You was right. It was too complicated.

roe-dl:
For my opinion point 6 is left only. All other points are done. 

Problably I cannot do point 6 today because of other work to be done. I'll do it as fast as possible.

bin/weewx/gettext.py
```diff
@@ -0,0 +1,166 @@
...
+import logging
+
+log = logging.getLogger(__name__)
+
+
+class Gettext(SearchList):
+
+    def __init__(self,generator):
+        """Create an instance of the class"""
+        super(Gettext,self).__init__(generator)
```
tkeffer, line 66: If you don't need any logic in `Gettext.__init__()`, then it can be omitted entirely.

bin/weewx/gettext.py
```diff
@@ -0,0 +1,166 @@
...
+                key: key to look up for
+                page: section name  
+                
+            """
+            _text_dict = self.generator.skin_dict.get('Texts',{})
+
+            # get the appropriate section for page if defined
+            # Note: no hierarchy here
+            if page:
+                if not page in _text_dict:
```
tkeffer, line 83: The usual Python idiom is `if page not in _text_dict`

bin/weewx/gettext.py
```diff
@@ -0,0 +1,166 @@
...
+                    log.error("could not find section [Texts][[%s]] for report %s" % (page,self.generator.skin_dict.get('REPORT_NAME','unknown')))
+                # Note: no 'else:' here, because we need the empty dict
+                # page is specified --> get subsection "page"
+                _text_dict = _text_dict.get(page,{})
+                # if the result is not a dict, make it a dict
+                if not isinstance(_text_dict,dict):
+                    _text_dict={page:_text_dict}
```
tkeffer, line 90: Don't get this one. Why would it not be a `dict`?

bin/weewx/gettext.py
```diff
@@ -0,0 +1,166 @@
...
+            # if key is not empty, get the value for key
+            if key:
+            
+                if key in _text_dict:
+                    # key is in dict _text_dict --> return value of key
+                    val = _text_dict[key]
+                else:
+                    # otherwise return key as value
+                    val = key
+
```
tkeffer, line 101: See class `weeutil.KeyDict`. It can do this for you.

bin/weewx/gettext.py
```diff
@@ -0,0 +1,166 @@
...
+            # entry found in localization file
+            return self.lang_files_dict[attr]
+        elif attr in self.cheetah_dict:
+            # entry found [CheetahGenerator] subsection
+            return self.cheetah_dict[attr]
+        elif attr=='nav':
+            # if $gettext(page='xxx').nav not in localization file, try 
+            # nav_xxx in [Labels][[Generic]] section of skin.conf
+            # helpful for extending Belchertown skin
+            if 'nav_'+self.page in self.skin_labels_dict:
```
tkeffer, line 152: Don't understand why we need this. What's special about the Belchertown skin?

bin/weewx/reportengine.py
```diff
@@ -254,6 +254,113 @@ def _build_skin_dict(self, report):
...
+        # used as a file name for a language file that is located in the
+        # 'lang' subdirectory of the skin directory. Like it is with the
+        # other path settings, a complete path overwrites the skin
+        # directory setting.
+
+        # If an localization file is defined, determine the path.
+        # As config_dict is not merged into skin_dict so far, we
+        # need to check both. config_dict (weewx.conf) has the final say,
+        # so check it first.
+        if 'lang' in self.config_dict['StdReport'][report]:
```
tkeffer, line 270: Don't see why `lang` would not obey the same hierarchical principles as any other option. It shouldn't be necessary to do "look here, if not there."

bin/weewx/reportengine.py
```diff
@@ -254,6 +254,113 @@ def _build_skin_dict(self, report):
...
+                             (lang_config_path, report))
+            except IOError as e:
+                log.debug("Cannot read localization file %s for report '%s': %s",
+                          lang_config_path, report, e)
+            except SyntaxError as e:
+                log.error("Failed to read localization file %s for report '%s': %s",
+                          lang_config_path, report, e)
+                raise
+        
+        # The key 'unit_system' defines the unit system to be used with
```
tkeffer, line 319: Please. No `unit_system`. At least until we discuss how it interacts with the present system of unit groups.

bin/weewx/gettext.py
roe-dl: Done.

bin/weewx/gettext.py
roe-dl: Done.

bin/weewx/gettext.py
roe-dl: It is an error check. It is not a dict if the user uses the name of `page` as a scalar value instead of a section. Just to make sure not to create an error in the next step.

I can remove it, if you don't want that check.

bin/weewx/gettext.py
roe-dl: Done.

bin/weewx/gettext.py
```diff
@@ -0,0 +1,166 @@
...
+        
+            # if key is empty but page is not return further class instance
+            if page:
+                cheetah_dict={}
+                cheetah_dict_key = 'CheetahGenerator'
+                if "FileGenerator" in self.generator.skin_dict and 'CheetahGenerator' not in self.generator.skin_dict:
+                    cheetah_dict_key = "FileGenerator"
+                if cheetah_dict_key in self.generator.skin_dict:
+                    cheetah_dict = Gettext._get_cheetah_dict(self.generator.skin_dict[cheetah_dict_key],page)
+                return FilesBinder(page,_text_dict,self.generator.skin_dict.get('Labels',{}).get('Generic',{}),cheetah_dict)
```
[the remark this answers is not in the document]
roe-dl: Done.

bin/weewx/gettext.py
tkeffer: The Pythonic way of doing things like this is to proceed and let an exception occur if `_text_dict` does not behave like a `dict`. Who knows: `_text_dict` may not be an actual `dict`, but rather a class that looks like one (called "duck typing").

bin/weewx/reportengine.py
roe-dl: You say, `weewx.conf` has the last, final say. If I obey that rule and obey the hierarchical principles, I have to merge the language config into `skin_dict` ***before*** the data from weewx.conf are merged into `skin_dict`. At that moment, `skin_dict['lang']` has not the final value. So I have to take into account not only `skin_dict`, but also `self.config_dict`. 

Anyway. I found a way to replace those `if` - `elif` - `else` sequence by a simple single line of code.

bin/weewx/gettext.py
tkeffer: Good rule of thumb in Python: if you find yourself using `isinstance()` you're probably doing something wrong. Not true all of the time of course, but it's worth re-examining the problem to see if there is a more Pythonic solution.

roe-dl: Done.

bin/weewx/gettext.py
roe-dl: So I explain it, but in the end, I remove it, because `FilesBinder` is not needed any more then.

The Belchertown skin (and I guess some other complex skins, too) contains:
```
                #if $page == "graphs"
                <h1>$obs.label.graphs_page_header</h1>
                #else if $page == "records"
                <h1>$obs.label.records_page_header</h1>
                #else if $page == "reports"
                <h1>$obs.label.reports_page_header</h1>
                #else if $page == "about"
                <h1>$obs.label.about_page_header</h1>
                #else if $page == "imprint"
                <h1>$obs.label.imprint_page_header</h1>
                #else if $page == "gdpr"
                <h1>$obs.label.gdpr_page_header</h1>
                #end if
```
If you want to add a new page, you have to write another `#else if`.

With the new `$gettext` feature that can be replaced by a single line:
```
<h1>$gettext("page_header",$page)</h1>
```
No need to change or add anything when adding a new page any more.

Now, the values are found in `[Labels][[Generic]]` instead of `[Texts]`, and they include the page's name. The code you asked about converted `page_header` and the value of `$page` to the key to look for in `[Labels][[Generic]`. 

But there is an alternative. You can move all those entries from `[Labels][[Generic]]` to `[Texts]`, and then you do not need special handling of those tags any more.

So I removed that. And I did point 6 of your comment with that as well.

bin/weewx/reportengine.py
roe-dl: I guess that's the most difficult part. Let me try to explain. 

`defaults.py` is always `US`. That's the starting point. Imagine a user from a metric country. To get metric units in the report s/he has to copy all the long list of unit groups with their respective metric units to somewhere within `weewx.conf`. One by one. Line by line. A long, annoying and error-prone work.

And then, the user sees a wonderful, complete list of those unit groups together with their metric units in `units.py` (that is `weewx.units.MetricUnits`). But that list (or `dict` in Python) is not available to the user as a starting point when it comes to setting up reports.

Why that?

(And I changed the code for that heavily in the hope it would satisfy your needs then. It is very short now, and I used other names and hope they will be more descriptive.)

tkeffer: OK, you're wearing me down. I just need to make sure that this will not break old skins in unexpected ways.

bin/weewx/gettext.py
```diff
@@ -0,0 +1,123 @@
...
+            if page:
+                _page_dict = weeutil.config.config_from_str('')
+                merge_dict = self.generator.skin_dict.get('Labels',{}).get('Generic',{})
+                if merge_dict:
+                    weeutil.config.merge_config(_page_dict,merge_dict)
+                merge_dict = Gettext._get_cheetah_dict(self.generator.skin_dict.get('CheetahGenerator',{}),page)
+                if merge_dict:
+                    weeutil.config.merge_config(_page_dict,merge_dict)
+                weeutil.config.merge_config(_page_dict,_text_dict)
+                return _page_dict
```
tkeffer, line 103: Trying to understand something here. You allow a page to be specified with an empty key (e.g., `$gettext('', 'some_page')`). Yet, I don't see any place where this is used in the skin.

What's the purpose of this functionality?

roe-dl: 1. Something like `$gettext(page=$page).Title` to set a page-dependent text in include files. Especially titles, meta tags in HTML and the like 
2. Can be used when creating sitemaps. `page` can be any page, not only the actual page.
    * `$gettext(page=$page).template` - the name of the template file
    * `$gettext(page=$page).stale_age` - live time of the page
    * `$gettext(page=$page).encoding` - encoding of the page
    * other lines that are placed in this section.

And as you ask I am not sure why I merged `[Labels][[Generic]]` there. I guess it was obsoleted by some earlier change and can be removed now. I now see no sense in something like `$gettext(page=$page).outTemp` as it does not depend on pages. I would remove that merge.

tkeffer: Why not go all the way, and make `gettext` act like a proper dictionary.

    $gettext[$page][template]

which in Cheetah's world can be abbreviated to

    $gettext[$page].template

This would also allow nesting:

    $gettext[$page][subpage].template
    $gettext[$page].subpage.template

tkeffer: This could be implemented using a class that holds the text dictionary internally. It's like a normal dict, except that it returns the key if the key is missing. Something like:

```python
from six.moves import collections_abc


class ParallelDict(collections_abc.Mapping):

    def __init__(self, source):
        self.source = source

    def __getitem__(self, key):
        try:
            return self.source[key]
        except KeyError:
            return key

    def __len__(self):
        return self.source.__len__()

    def __iter__(self):
        for key in self.source:
            yield key
```

roe-dl: I will try that.

tkeffer:
We are getting close.

With these changes, most `skin.conf` files will consist of nothing but information about what to generate. No labels (which are in the `lang` files), and probably no label formats either (which most skins don't need to modify).

I wish we could slim down the `[StdReport]//[[Defaults]]` section. A lot of the information in there is no longer needed, and is distracting. However, that would break old skins. Any ideas are welcome.

bin/weewx/gettext.py
roe-dl: I changed gettext.py and replaced `$gettext()` by `$gettext[]`. It generally works. 

Actually if there is an undefined key in the second level (`$gettext["defined key"]["undefined key"]`) an error is returned that refers to `configObj` class. So I guess, I have to write a recursive function to convert the subsections to `ParallelDict` or something.

roe-dl: What is the difference between `ParallelDict` and `KeyDict` regarding our needs?

roe-dl:
> Why not go all the way, and make `gettext` act like a proper dictionary.

With that idea I wrote a completely new version of `gettext.py`. The previous code remains as comments for comparison until discussion is completed.

The next commit shows that version of the class `Gettext`.

tkeffer:
Our commits may have crossed in the ether.

I took all of your changes and put them in a new branch [I18N](https://github.com/weewx/weewx/tree/i18n). I missed the changes you made in `gettext.py`. However, I came up with my own version, which I think is much simpler.

Take a look at the branch. If you like what you see, I'll commit it into `development`.

roe-dl:
I copied `gettext.py` from I18N to my installation and let it run. For defined keys it works well.

If I use a key that is not defined the key itself should be returned. That is the general rule, we set. And that is the case for keys at top level of the dict. But for keys deeper down, there are errors. It tried `$gettext["index"]["blub"]` (added it at the end of `index.html.tmpl` for test purposes). `index` is defined, `blub` is not. So the key `blub` should be returned. But it isn't. The following errors are logged:
```
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: Generate failed with exception '<class 'KeyError'>'
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: **** Ignoring template /etc/weewx/skins/Seasons/index.html.tmpl
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: **** Reason: 'blub'
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****  Traceback (most recent call last):
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****    File "/usr/share/weewx/weewx/cheetahgenerator.py", line 327, in generate
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****      unicode_string = compiled_template.respond()
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****    File "cheetah__etc_weewx_skins_Seasons_index_html_tmpl_1620038716_915841_22801.py", line 647, in respond
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****    File "cheetah__etc_weewx_skins_Seasons_index_html_tmpl_1620038716_915841_22801.py", line 246, in __errorCatcher26
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****    File "<string>", line 1, in <module>
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****    File "/usr/lib/python3/dist-packages/configobj.py", line 554, in __getitem__
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****      val = dict.__getitem__(self, key)
May  3 12:45:18 LokalWiki weewx[150826] ERROR weewx.cheetahgenerator: ****  KeyError: 'blub'
```

Surprisingly (at least to me) `$gettext.index.blub` does not result in the same error. Instead the whole tag is returned, what is not, what we want, too.

When I started to re-write `gettext.py` I started with a solution similar to yours. And I got the same error messages. Because `configobj.py` is mentioned in that error message I concluded that by `ParallelDict(_text_dict)` the top level of the dictionary is converted from `ConfigObj` to `ParallelDict` only, but the subsections are not. Therefore I thought I have to copy it all recursively and convert each subsection separately. And at last I asked what the difference between `ParallelDict` and `KeyDict` is. As `KeyDict` is already there, I used it.

You did not merge the values from `[CheetahGenerator]`. But it is strictly spoken no part of internationalization, so it could be omitted.

I will put it both together and then push it.

tkeffer:
I like your solution better. It does have one small problem, although I think we can live with it. If a `lang` is not specified, then we get
```
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: Generate failed with exception '<type 'exceptions.TypeError'>'
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: **** Ignoring template /home/weewx/skins/I18N/index.html.tmpl
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: **** Reason: string indices must be integers, not str
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: ****  Traceback (most recent call last):
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: ****    File "/home/tkeffer/git/weewx/bin/weewx/cheetahgenerator.py", line 327, in generate
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: ****      unicode_string = compiled_template.respond()
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: ****    File "_home_weewx_skins_I18N_index_html_tmpl.py", line 104, in respond
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: ****    File "_home_weewx_skins_I18N_index_html_tmpl.py", line 70, in __errorCatcher1
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: ****    File "<string>", line 1, in <module>
May  3 05:12:57 nuc wee_reports[306996] ERROR weewx.cheetahgenerator: ****  TypeError: string indices must be integers, not str
```

roe-dl:
I can look into it in the evening European time.

roe-dl:
What about replacing
```
class KeyDict(dict):
    """A dictionary that returns the key for an unsuccessful lookup."""
    def __missing__(self, key):
        return key
```
in `weeutil.py` by
```
class KeyDict(dict):
    """A dictionary that returns the key for an unsuccessful lookup."""
    def __missing__(self, key):
        return IndexStr(key)

class IndexStr(str):
    def __getitem__(self,key):
        try:
            return str(self)[key]
        except TypeError:
            return key
```
?

After changing `weeutil.py` in that way, no error message is logged any more. And the correct string is returned. And at first I see no other problem resulting from that change.

roe-dl:
As you did not comment, I have a second chance:
```
class KeyDict(dict):
    """A dictionary that returns the key for an unsuccessful lookup."""
    def __missing__(self, key):
        return IndexStr(key)

class IndexStr(str):
    def __getitem__(self,key):
        try:
            return super(IndexStr,self).__getitem__(key)
        except TypeError:
            return IndexStr(key)
```
I think, that is a little bit cleaner.

tkeffer:
Nice. I've made the changes.

I think we've answered the major questions, so I've merged everything into `development`. There are still some things to do (listed in the [TODO list](https://github.com/weewx/weewx/blob/development/TODO.md)). Let me work on those.

Thanks for a great contribution, Karen!

roe-dl:
Thank you, too, for your patience and all.

tkeffer:
Started modifying test suites and immediately ran into a problem. In `reportengine.py`:
```python
            try:
                units_dict = weewx.units.std_groups[
                    weewx.units.unit_constants[report_units_base]].copy()
                weeutil.config.merge_config(skin_dict, {'Units': {'Groups': units_dict}})
            except (SyntaxError, TypeError, IndexError, ValueError, IOError) as e:
                log.error("Error ('%s') merging unit system '%s' for report '%s'",
                          e, report_units_base, report)
```
The variable `units_dict` is the same type as the standard unit groups defined in `units.py`, that is, `ListOfDicts`, so after the merge, `skin_dict['Units']['Groups']` ends up being of type `ListOfDicts`.

When the final merge happens (line 332)
```python
        weeutil.config.merge_config(skin_dict, self.config_dict['StdReport'][report])
```
the `ListOfDicts` part is ignored, probably because it doesn't inherit from `configobj.Section`. The final results are whatever was in `config_dict['StdReport'][report]` and nothing else.

roe-dl:
That is interesting, as it was not the case in my tests. I tested it with all possible values, and the units changed according to the setting. I will think about it and hope I can come around with a solution.

roe-dl:
I would suggest to change that code snippet to:
```
            try:
                units_dict = weewx.units.std_groups[
                    weewx.units.unit_constants[report_units_base]].copy()
                for __key in units_dict:
                    skin_dict['Units']['Groups'][__key] = units_dict[__key]
            except (SyntaxError,TypeError,IndexError,ValueError,IOError) as e:
                log.error("error merging unit system '%s' for report '%s': %s" % (report_units_base,report,e))
```

The units dicts have no subsections. So calling `merge_config` is not really necessary. The assignment within the `for` loop preserves the type of `[Units][[Groups]]` as `<class 'configobj.Section'>`. 

The section `[Units][[Group]]` already exists because of `defaults.py`. But if you want to make it sure you could insert
```
                weeutil.config.merge_config(skin_dict, 
                  weeutil.config.config_from_str('''[Units]
                  [[Groups]]'''))
```
before the `for` loop. That inserts `[Units][[Groups]]` if it does not exist, but it does not overwrite anything, that was set in the defaults.

roe-dl:
> Update the documentation. I think the I18N sections still use gettext() rather than gettext[].

That I did. See [commit 64cf55d](https://github.com/weewx/weewx/pull/665/commits/64cf55d06a87e461f48568433773f36dadff8175)

tkeffer:
I went with
```python
                skin_dict['Units']['Groups'].update(units_dict)
```

roe-dl:
That's better.

I changed it in my installation. No problem.

tkeffer:
I keep discovering new texts in the Seasons skins. I need German translations for

        "Always down"     = "Always down"
        "Always up"       = "Always up"
        "less than yesterday" = "less than yesterday"
        "more than yesterday" = "more than yesterday"
        "full"            = "full"    # As in "The Moon is 21% full"

roe-dl:
```
    # the sun far north or south
    "Always down"     = "Polarnacht"
    "Always up"       = "Polartag"
    # daylight duration in time units
    "less than yesterday" = "weniger als gestern"
    "more than yesterday" = "mehr als gestern"
    # the moon
    "full"            = ""    # As in "The Moon is 21% full"  *)
```

*) I would not say that in german (like "Der Mond ist 21% voll."). The sentence is grammatically correct in german, but it sounds like the moon had drunken too much. I searched Google, and I found no similar wording. I would write something like "Sichtbarkeit: 21%" or "zunehmender Mond mit 21%" (word by word: "waxing moon with 21%").

roe-dl:
I have another idea:
```
"full" = "sichtbar"  # as in "Der Mond ist zu 21% sichtbar." but I would use "21% sichtbar" alone.
```

tkeffer:
You're going to kill me, but I switched back to your original function style `gettext()`. I also introduced a special context sensitive version `pgettext()`. 

They both follows the GNU and Python `gettext` semantics.

The two changes will make it easier to write tools, and to do alternative implementations in the future, should we want them.

roe-dl:
I did not say anything the last few weeks.

tkeffer:
I know. It's just that I ended up back where you started.

PR #668, opened 2021-04-25
minor doc updates for debian installation instructions

vinceskahan:
This provides minor tweaks for debian based on rather frequently asked questions on weewx-users.

- make the apt related commands a little more obvious since we're still in the python2/python3 transition
- for setup.py users itemize how to use systemd as the init system


tkeffer:
Looks good. Minor change: as a matter of style, the ordering of options should be kept consistent. You (correctly) suggested `systemd` approach before the `/etc/init.d` approach. Removal instructions should follow the same ordering.

vinceskahan:
Done.   Thanks.  Feel free to merge when the spirit moves you :-)

PR #669, opened 2021-04-26
add kwargs to XType `get_scalar()` and `get_series()` signatures

gjr80:
As discussed on Slack #development 28 January 2021.
Pre-cursor to adding support for improved cumulative plots

### DIFF
--- a/bin/wee_database
+++ b/bin/wee_database
@@ -170,14 +170,14 @@ def main():
     if options.rename_column and not options.to_name:
         sys.exit("Option --rename-column requires option --to-name")
 
-    if options.date and not (options.calc_missing or options.rebuild_daily):
-        sys.exit("Option --date can only be used with options --calc-missing or --rebuild-daily")
+    if options.date and not (options.calc_missing or options.rebuild_daily or options.reweight):
+        sys.exit("Option --date can only be used with options --calc-missing or --rebuild-daily or --reweight")
 
-    if options.from_date and not (options.calc_missing or options.rebuild_daily):
-        sys.exit("Option --from can only be used with options --calc-missing or --rebuild-daily")
+    if options.from_date and not (options.calc_missing or options.rebuild_daily or options.reweight):
+        sys.exit("Option --from can only be used with options --calc-missing or --rebuild-daily or --reweight")
 
-    if options.to_date and not (options.calc_missing or options.rebuild_daily):
-        sys.exit("Option --to can only be used with options --calc-missing or --rebuild-daily")
+    if options.to_date and not (options.calc_missing or options.rebuild_daily or options.reweight):
+        sys.exit("Option --to can only be used with options --calc-missing or --rebuild-daily or --reweight")
 
     if options.dest_binding and not options.transfer:
         sys.exit("Option --dest-binding can only be used with option --transfer")
--- a/bin/wee_database
+++ b/bin/wee_database
@@ -170,14 +170,20 @@ def main():
     if options.rename_column and not options.to_name:
         sys.exit("Option --rename-column requires option --to-name")
 
-    if options.date and not (options.calc_missing or options.rebuild_daily or options.reweight):
-        sys.exit("Option --date can only be used with options --calc-missing or --rebuild-daily or --reweight")
+    if options.date and not (options.calc_missing or options.rebuild_daily
+                             or options.reweight):
+        sys.exit("Option --date can only be used with options "
+                 "--calc-missing, --rebuild-daily, or --reweight")
 
-    if options.from_date and not (options.calc_missing or options.rebuild_daily or options.reweight):
-        sys.exit("Option --from can only be used with options --calc-missing or --rebuild-daily or --reweight")
+    if options.from_date and not (options.calc_missing or options.rebuild_daily
+                                  or options.reweight):
+        sys.exit("Option --from can only be used with options "
+                 "--calc-missing, --rebuild-daily, or --reweight")
 
-    if options.to_date and not (options.calc_missing or options.rebuild_daily or options.reweight):
-        sys.exit("Option --to can only be used with options --calc-missing or --rebuild-daily or --reweight")
+    if options.to_date and not (options.calc_missing or options.rebuild_daily
+                                or options.reweight):
+        sys.exit("Option --to can only be used with options "
+                 "--calc-missing, --rebuild-daily, or --reweight")
 
     if options.dest_binding and not options.transfer:
         sys.exit("Option --dest-binding can only be used with option --transfer")
--- a/bin/weecfg/tests/expected/weewx43_user_expected.conf
+++ b/bin/weecfg/tests/expected/weewx43_user_expected.conf
@@ -26,7 +26,7 @@ log_failure = True
 socket_timeout = 20
 
 # Do not modify this - it is used by setup.py when installing and updating.
-version = 4.5.1
+version = 4.6.0a1
 
 ##############################################################################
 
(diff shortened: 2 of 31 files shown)


tkeffer:
Just thinking: this will break backwards compatibility with any custom xtypes out there, won't it? There's not many, but there are definitely a few.

gjr80:
I guess it will, in some cases. I guess we can scrub that idea.

tkeffer:
Don't give up so easily! We have two options:
1. Wrap in a try block, catching any `TypeError` exceptions. Then retrying without the `kwargs`:
    ```python
    try:
        return xtype.get_scalar(obs_type, record, db_manager, **option_dict)
    except TypeError:
        return xtype.get_scalar(obs_type, record, db_manager)
    ```
2. Or, examine the number of arguments first:
    ```python
    if xtype.get_scalar.__code__.co_argcount > 3:
        return xtype.get_scalar(obs_type, record, db_manager, **option_dict)
    else:
        return xtype.get_scalar(obs_type, record, db_manager)
    ```

The former is more Pythonic, but the `TypeError` could catch things other than a mismatch in the number of arguments. Both are ugly, but do maintain backwards compatibility.

Give it a try and see how it feels.

tkeffer:
BTW, it would also be good to see the rest of the solution. In particular, how are you going to pass on `option_dict`?

gjr80:
> BTW, it would also be good to see the rest of the solution. In particular, how are you going to pass on `option_dict`?

I presume you are referring to the cumulative plots solution?

tkeffer:
> > BTW, it would also be good to see the rest of the solution. In particular, how are you going to pass on `option_dict`?
> 
> I presume you are referring to the cumulative plots solution?

Yes.

gjr80:
Have added the `ImageGenerator` changes to the PR. Basically `option_dict` consists of the accumulated plot options plus `plotgen_ts` less the explicitly included `aggregate_type` and  `aggregate_interval`.

All the rest of the work is done within the xtype that produces the series, aside of course from adding the xtype to the xtypes list.

tkeffer:
Still need a solution to the backwards compatibility problem. I think I'd prefer the `TypeError` approach

roe-dl:
I am one of the XType extension authors, and I think I can add both definitions, that the extension works whether there are the old or the new function definitions in the case class.

roe-dl:
I am one of the authors of an XType extension, and I could change the functions if I know when it becomes active.

tkeffer:
> I am one of the authors of an XType extension, and I could change the functions if I know when it becomes active.

This is a kind offer, but doesn't solve the problem, because there will be a mix of new and old xtype software out there for years.

gjr80:
> > I am one of the authors of an XType extension, and I could change the functions if I know when it becomes active.

> This is a kind offer, but doesn't solve the problem, because there will be a mix of new and old xtype software out there for years.

Perhaps you meant change your XType extensions? If so you sure could, but that will mean any users of your upgraded XType extensions will then need to upgrade WeeWX to (at least) the version that first incudes the above XType changes. Not a great problem in itself but just another backwards compatibility complication.

tkeffer:
Yes. Perhaps I didn't state that very well. I was trying to say that there will be a mix of new and old weewx versions for years. If we want to support both, it's not good enough to just upgrade the extensions. Some user with an old version of weewx may try to install your new-style extension, and it will fail.

There are better solutions.

gjr80:
>  1. Wrap in a try block, catching any `TypeError` exceptions. Then retrying without the `kwargs`:
>        ```python
>        try:
>            return xtype.get_scalar(obs_type, record, db_manager, **option_dict)
>        except TypeError:
>            return xtype.get_scalar(obs_type, record, db_manager)
>        ```
>  2. Or, examine the number of arguments first:
>        ```python
>        if xtype.get_scalar.__code__.co_argcount > 3:
>            return xtype.get_scalar(obs_type, record, db_manager, **option_dict)
>        else:
>            return xtype.get_scalar(obs_type, record, db_manager)
>        ```
Getting back to these, I've implemented 1 and it runs fine. Now thinking aloud about how things might break or mask errors. The issue will be `TypeError` exceptions, all other exceptions encountered in an XType `get_scalar()` or `get_series()` method will passed up as per normal and logged/cause WeeWX to fail, no issue there. For `TypeError` exceptions there are two cases we need to consider:

1. 'legacy' XTypes whose `get_scalar()` or `get_series()` methods **do not** support additional kwargs
2. 'new style' XTypes whose `get_scalar()` or `get_series()` methods **do** support additional kwargs

Considering 1. When called by `xtypes.get_scalar()` or `xtypes.get_series()` legacy XTypes will immediately raise a `TypeError` if they receive any kwargs. This will force another call to the legacy XType this time without the kwargs. If there are no exceptions (other than `UnknownType`, `UnknownAggregation` or `CannotCalculate`) then all is good, no problems. If a `TypeError` is encountered this will be passed up and actioned accordingly as there is no longer any handler intercepting the `TypeError` exception.

Considering 2. When called by `xtypes.get_scalar()` or `xtypes.get_series()` new style XTypes will not immediately raise a `TypeError` if they receive any kwargs. If a `TypeError` is raised due to an error in the XType this will be raised and caught by the handler in `xtypes.get_scalar()` or `xtypes.get_series()`. This will force another call to the new style XType but this time without the kwargs. If the same `TypeError` exception is encountered it will again be passed up and actioned/reported accordingly as there is no longer any handler intercepting the `TypeError` exception. If the original `TypeError` was due to say, some processing of the kwargs (that no longer exist), it is possible the `TypeError` will no longer occur and the net effect is that the original `TypeError` is swallowed. I suspect though that if the new style XType is called without the kwargs you have a pretty good chance of seeing exceptions other than `TypeError` if you see any exceptions at all.

So we can draw a couple of things from this:

1. the issue of swallowing `TypeError` exceptions does not lay with legacy XTypes but rather with new style XTypes
2. new style XTypes will only swallow `TypeError` exceptions if the exception is raised when processing kwargs

Would any half way decent developer writing an XType that will use kwargs be doing some checks in his XType code to ensure his kwarg(s) in fact exist? I think I would, certainly knowing about this aspect of XTypes I would. Perhaps we can mitigate the issue with more guidance to developers on writing XTypes? Let's face it I don't think that many folks will be developing XTypes and those few that are I would expect would need to read about what they need to do before they actually develop any code.

Bottom line is I don't think there is a great problem in adopting your option 1 provided we give include some appropriate guidance in the XType documentarian. 

I'm sorry, this is necessarily long.

tkeffer:
Good analysis. I think we can safely conclude that a swallowed kwargs is very unlikely in all but the most corner of corner cases.

Do you want to implement it?

BTW, I don't think you need to `pop()` arguments `aggregate_type` and `aggregate_interval`. Just remove them from your call to `get_series()`.

gjr80:
Ironic I think that new style rather than legacy XTypes will be the ones to suffer.

I have a suspicion that `aggregate_type` and `aggegate_interval` are already named arguments and including them in the kwargs dict throws an error. But perhaps I missed something. Will have another look tomorrow and sort out the PR.

gjr80:
I knew this would happen; my fork has added a whole pile of commits that just muddy the waters and test my Git-fu. Closing this PR and will resubmit with a link back to this conversation.

tkeffer:
It would have turned out fine if you had pressed on. All of those commits already existed in `development`.

gjr80:
> It would have turned out fine if you had pressed on. All of those commits already existed in `development`.

That was my first thought, but when I saw entries that indicated there were conflicts that needed to be dealt with (which no longer seem to appear anywhere) I decided to bail and run. Have seen too many other PRs that end up adding crud to the repo and I wanted to avoid that.

PR #670, opened 2021-05-03
Airmar (or NMEA0183/NMEA2000) driver

AlexMclellan:
FYI; As you'll be aware, I'm not the author of this, it comes from here; https://groups.google.com/g/weewx-development/c/q6IHzkmyhBo?pli=1, and appears to originally be developed by Aleksandar Tsankov with updates applied by various people.

It has been floating around on the google group for awhile now, so I thought I'd open a pull request, so future development on it could be tracked, and also means that, if anyone wants to start using it, then it should come packaged with weewx and they don't need to hunt around for it on the google group.

Before making this merge request, I cleaned it up, so it's slightly more PEP8 compliant and may even run with python 3. Although it appears that my installation of weewx 4.5.1 is using python 2.7. So that's what it's tested against.

I've flicked through the notes for developers, however as I'm not 100 % intimate with the driver there may be things that I've missed. If you want me to fix them, just mention it in a code review.

I've had a look around at the principal developers accounts, and noticed that @tkeffer has a NMEA0183 weewx service (weewx-nmea-xdr). I'm now wondering if there's a reason why this driver hasn't been commit to the repo yet. And I'm also wondering if this should be renamed to NMEA2000.py or NMEA0183.py, because it should at least be partially compliant with those standards.

Anyway, I've at least started a discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Support for over 70 types of hardware including, but not limited to:
 * Argent Data Systems WS1;
 * KlimaLogg Pro;
 * New Mountain;
-* AirMar 150WX;
+* AirMar weatherstations (150WX, 200WX, etc) and other NMEA 2000 or NMEA 0183 compliant devices;
 * Texas Weather Instruments;
 * Dyacon;
 * Meteostick;
--- a/bin/weewx/drivers/airmar.py
+++ b/bin/weewx/drivers/airmar.py
@@ -1,0 +1,605 @@
+#!/usr/bin/env python
+
+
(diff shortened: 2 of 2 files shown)


tkeffer:
For drivers to be included in the WeeWX package they need to be popular. I'm not seeing any AIRMAR stations on the WeeWX registry. They also need to work under Python 2 and 3.

Why not package your driver as an extension? That's the intention of extensions.

The NMEA driver you referred to works with XDR sentences only. That's why it's named the way it is.

PR #671, opened 2021-05-16
Support for canonical URL <link rel="canonical" … />

roe-dl:
Google suggests to include a [canonical URL](https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls) into the header of HTML files. Often websites are available by HTTPS as well as HTTP, and with `www` and without. To help Google to sort out those variants and duplicate content, web administrators shall inform about the real URL bei setting up a canonical URL. 

To my knowledge the filename to be created (including relative path) is not available in Cheetah unfortunately. 

With the changes from this pull request setting a canonical URL would be possible like this:
```
<link rel="canonical" href="$station.station_url/$filename" />
```

One could argue that the filename could be set literally in the template. But that is not always possible:
* in case of templates in the `[[SummaryByMonth]]` or `[[SummaryByYear]]` section because the file name includes variable parts
* in case there is a single include file to define the HTML header section for all files to be created.

### DIFF
--- a/bin/weewx/cheetahgenerator.py
+++ b/bin/weewx/cheetahgenerator.py
@@ -309,7 +309,10 @@ class CheetahGenerator(weewx.reportengine.ReportGenerator):
                     pass
 
             searchList = self._getSearchList(encoding, timespan,
-                                             default_binding, section_name)
+                                             default_binding, section_name,
+                                             os.path.join(
+                                               os.path.dirname(report_dict['template']),
+                                               _filename))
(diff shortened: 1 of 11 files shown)


tkeffer:
Seems reasonable. Let me finish the I18N material first, then I'll get to this.

tkeffer:
How about setting `<link>` in the Seasons skin?

roe-dl:
Yes, I think this is a good idea.

PR #672, opened 2021-05-18
Clean shutdown for wxxtype services

edi-x:
I had problems with the rain rate values. When I restarted weewx the rain rates were calculated correctly for a few days and after different periods of time the RainRater always calculated a rain rate of 0, even when it was raining. After some investigations I figured out that my service is restarted "softly" by the initd script due to occasionally happening USB problems. After such a soft resart the RainRater did no longer work properly. According to some added log messages there were at least two active instances of self.rain_events (so I assume there were two RainRater instances active). One was used to collect data about new rain events and the other on to calculate the rain rate. Because the calculating RainRater did never see new rain events it always calculated a rain rate of 0.

Long story short, I saw that a engine shutdown should trigger the shutdown function of the RainRater, but this function was named "shut_down" there. So I renamed the method here (and for some other services, too) and now the RainRater seems to work correct again, even after such soft restarts.


tkeffer:
I certainly see the misnamed method, but shouldn't the correct name be `shutDown()`, not `shutdown()`?

edi-x:
Oh dear! Absolutely! I made a mistake while transferring the changes to the branch :-(
In my already tested version the functions are named shutDown(). I updated the merge request accordingly.
Sorry for that and thanks for your hint.

PR #673, opened 2021-05-24
add kwargs to XType `get_scalar()` and `get_series()` signatures to support complex imagegenerator plots

gjr80:
Refer to previous PR (#669) conversation.
Also discussed on Slack #development 28 January 2021.

### DIFF
--- a/bin/weewx/imagegenerator.py
+++ b/bin/weewx/imagegenerator.py
@@ -190,12 +190,23 @@ class ImageGenerator(weewx.reportengine.ReportGenerator):
                     # Now its time to find and hit the database:
                     binding = line_options['data_binding']
                     db_manager = self.db_binder.get_manager(binding)
+                    # we need to pass the line options and plotgen_ts to our xtype
+                    # first get a copy of line_options
+                    option_dict = dict(line_options)
+                    # but we need to pop off aggregate_type and
+                    # aggregate_interval as they are used as explicit arguments
+                    # in our xtypes call
+                    option_dict.pop('aggregate_type', None)
+                    option_dict.pop('aggregate_interval', None)
+                    # then add plotgen_ts
+                    option_dict['plotgen_ts'] = plotgen_ts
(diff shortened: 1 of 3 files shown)


bin/weewx/imagegenerator.py
```diff
@@ -190,12 +190,23 @@ def genImages(self, gen_ts):
                     # Now its time to find and hit the database:
                     binding = line_options['data_binding']
                     db_manager = self.db_binder.get_manager(binding)
+                    # we need to pass the line options and plotgen_ts to our xtype
+                    # first get a copy of line_options
+                    option_dict = dict(line_options)
+                    # but we need to pop off aggregate_type and
+                    # aggregate_interval as they are used as explicit arguments
+                    # in our xtypes call
+                    option_dict.pop('aggregate_type', None)
```
tkeffer, line 199: Don't see why you need to pop `aggregate_type` and `aggregate_interval`. Just remove them from the call to `get_series()`. The `**option_dict` would expand to include them.

gjr80: Maybe I am doing something wrong but when I leave them in I get a `TypeError: get_series() got multiple values for keyword argument 'aggregate_type'`

gjr80:
As per end of #669 Conversation `xtypes.get_scalar()` and `xtypes.get_series()` now catch any `TypeError` exceptions and then try call again but without hte kwargs. This will maintain backwards compatibility with 'legacy' XTypes but also imposes some (minor) limitations on the XTypes that do support kwargs for their `get_scalar()` and `get_series()` methods.

@tkeffer. I don't like nested `try..except` but could not see how to avoid it in this case, at least it doesn't detract from the readability of the code too much. Perhaps you have a better construct?

tkeffer:
See commit https://github.com/weewx/weewx/commit/30b16de848f1412c39c7389a2cccad5670fa3705

It's still nested, but a bit simpler

PR #675, opened 2021-05-26
german localization for Mobile and Smartphone skins

roe-dl:
These are german localization files for Mobile and Smartphone skins.

There is a difficulty. If words are put together in the german language, a hyphen is needed between the words or the words are put together without space between them. "Last 7-day Temperatur" translates to "7-Tage-Temperatur" or better "Temperaturen der letzten 7 Tage". This is not possible with the current structure of the language file.


tkeffer:
I assume you're referring to lines like
```
<h4>$gettext["Mobile"]["7-day"] $obs.label.outTemp</h4>
```

This could be changed to

```
<h4>$gettext["Mobile"]["7-day temperature"]</h4>
```

tkeffer:
Unfortunately, this also means we have to go back and get translations for all the other languages.

Sigh!

PR #691, opened 2021-08-22
Proposal : Adding a requirements.txt file.

moonpyk:
Using it is purely optional but gives a good view of what is needed
immediately to get started in coding for weewx, it also makes stuff
easier when using virtualenvs.
Compatible with Python 2 and Python 3.

Kind regards

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,18 @@
+configobj
+
+# Cheetah or Cheetah3 depending on Python version
+Cheetah; python_version < '3.0'
+Cheetah3; python_version >= '3.0'
+
+# Pillow with fixed version depending on Python version
+Pillow==5.1; python_version < '3.0'
+Pillow; python_version >= '3.0'
+
+pyserial
+pyusb
+
+# Ephem is optional, you can comment this line if you don't intent to use it
+pyephem
+
+# MySQL is optional, you can comment this line if you don't intent to use it
+mysqlclient


tkeffer:
Thanks for the PR.

We support both Python 2 and Python 3, which have slightly different requirements. Also, not everyone needs pyserial, pyusb, and/or mysqlclient. How would you solve this?

moonpyk:
Hi @tkeffer, you are absolutely right, I'm gonna have a look in that direction

moonpyk:
@tkeffer for the sightly different versions of packages needed, requirements.txt are supporting that with python version specifiers, for the other part (not needed packages for some persons), we would need to create different requirements-<variant>.txt referring to a basic blunt requirements.txt for absolute basic needs.

While this totally possible, I think this beats the "quick and simple" way I wanted to offer with that : requirements.txt are mainly targeted at developers to get a quick dev env running. I think they won't mind pulling mysql + ephem even if they won't use it, as unit tests indirectly depend on them, it would be a good idea to make sure a little change somewhere doesn't break anything for mysql and/or ephem users.

If you are okay with a single requirements.txt file with proper python version specifiers and maybe a comments saying like "comment this line if you don't need mysql", "you might want to use mariadb-client" inside I would be very glad to pursue in this direction.

tkeffer:
That should work. I would definitely prefer a single requirements.txt file, even if it pulls in more than the average user is likely to need.

Thanks for researching this. I often use pip and always have to look up the exact package names!

moonpyk:
@tkeffer we should be good now :)

tkeffer:
1. `PIL` is not available on pypi. Only `Pillow`. However, Python 2.7 is supported only by older versions of `Pillow.` See this [support matrix](https://pillow.readthedocs.io/en/stable/installation.html#python-support). So, you'll need something like:

    ```
    Pillow==5.1; python_version < '3.0'
    Pillow; python_version >= '3.0'
    ```

2. `pysqlite` is not necessary at all. All versions of Python that we support come with sqlite client libraries.

Make sure you try your requirements.txt file!

moonpyk:
@tkeffer sorry, I didn't paid enough attention, I based my decisions around http://www.weewx.com/docs/macos.htm (with Python 2 selected). I don't have any python2 interpreter around anymore.

Cheers.

tkeffer:
Looks good. Thanks.

As for using and testing on Python 2, as your notes say, use a virtualenv. 

I use [`pyenv`](https://github.com/pyenv/pyenv), and can highly recommend it, Makes it easy to have multiple versions of Python around, including oddballs like `pypi`.

PR #693, opened 2021-08-22
Fix Davis Vantage backfilling (#692)

### LINKED ISSUE #692: Vantage: wee_device --dump backfills timeout every 2 pages (10 records)
# Problem
Occasionally my station will suffer a multi-day power outage, and upon restart, does not properly backfill the `archive` with records on the Vantage console. To recover missing data I use the `wee_device --dump` command. In my setup, this backfill can take many minutes as the archive database (MySQL) is offsite across a high latency WAN connection.

When backfilling more than 2 pages of data, the command will consistently throw a `weewx.WeeWxIOError: Timeout in get_data_with_crc16` exception. This occurs when attempting to fetch the third page of "unseen" records. If you re-run the command, it will properly read through the recently inserted records, insert two more pages worth of records, then fail again.

I've tracked the issue down to the time it takes to handle a single page of data from the Vantage console. If handling of a full page of console data happens within a few seconds, the next page is fetched successfully. If handling a single page takes more than ~10 seconds, page fetches eventually fail. Because my archive fetches/reads can be slow, this happens consistently in my environment.

```Using configuration file /home/weewx/weewx.conf
Using Vantage driver version 3.2.2 (weewx.drivers.vantage)
Proceeding will dump all data in the logger.
Are you sure you want to proceed (y/n)? Starting dump ...
 Traceback (most recent call last):: 2021-08-13 23:50:00 PDT (1628923800)
  File "/home/weewx/./bin/wee_device", line 78, in <module>
    main()
  File "/home/weewx/./bin/wee_device", line 74, in main
    device.configure(config_dict)
  File "/home/weewx/bin/weewx/drivers/__init__.py", line 67, in configure
    self.do_options(options, parser, config_dict, not options.noprompt)
  File "/home/weewx/bin/weewx/drivers/vantage.py", line 2101, in do_options
    self.dump_logger(station, config_dict, options.noprompt)
  File "/home/weewx/bin/weewx/drivers/vantage.py", line 2703, in dump_logger
    for record in converted_generator:
  File "/home/weewx/bin/weewx/units.py", line 1682, in __next__
    _record = next(self.input_generator)
  File "/home/weewx/bin/weewx/drivers/vantage.py", line 697, in genArchiveDump
    _page = self.port.get_data_with_crc16(267, prompt=_ack, max_tries=self.max_tries)
  File "/home/weewx/bin/weewx/drivers/vantage.py", line 233, in get_data_with_crc16
    raise weewx.WeeWxIOError("Timeout in get_data_with_crc16")
weewx.WeeWxIOError: Timeout in get_data_with_crc16
```

*NOTE: The reason this backfill is necessary is a byproduct of another local issue that prevents weewx from properly backfilling data when restarted. I've explained it below.*

# Verification
After a failure (which occurs quickly when starting the backfill), you can re-start the command, and it will successfully fetch N+2 pages, insert the 10 new archive records, then timeout. You can repeat this process to fully backfill the archive. There is no limit to how many pages of already-inserted data can be read.

There are no documented timeouts in the Vantage serial console documentation.

# Solution
The fix appears to be simple: fetch the entire Vantage console database before attempting to insert into the archive. This is simple as changing L2703 in `vantage.py` to `for record in list(converted_generator):` . By wrapping the generator in a list, Python fetches the entire console (max of 2650 records),  I have tested locally, and the change worked: the entire console was dumped in a few seconds, then the inserts handled slowly later.

While the generator approach is elegant if the archive is large, the Vantage console can only store 2560 records, each of which is less than 100 bytes of data (hard limits already implemented in the driver), so the risk to memory exhaustion with this change is low.

https://github.com/weewx/weewx/blob/master/bin/weewx/drivers/vantage.py#L2703 

# Why does the backfill not work?
If your station operates without a battery-backed clock and weewx is restarted before you have properly resynchronized the clock, weewx's backfill procedures discards valid records that appear in the "future." When the Vantage generates its next archive record, weewx will store it correctly, leaving a gap in the archive.

Subsequent restarts of weewx with a properly set clock will not trigger a new backfill, as WeeWx looks for records after the last known record in the archive (specifically it issues a `DMPAFT` to the Vantage to fetch only records after the last archive record.)

Using `wee_device --dump` is the only way to backfill the records.

gashalot:
Fix backfilling weewx archive for Davis Vantage users by fetching entire console archive before attempting to insert into weewx archives. Prevents slow archive inserts from causing Davis Vantage console timeouts.

Closes #692

### DIFF
--- a/bin/weewx/drivers/vantage.py
+++ b/bin/weewx/drivers/vantage.py
@@ -693,6 +693,9 @@ class Vantage(weewx.drivers.AbstractDevice):
         
         # Cycle through the pages...
         for ipage in range(512):
+            # Provide users with some status as the dumper
+            if ipage % 16 == 0:
+                log.debug("Requesting page %d / 512 of archive", ipage)
             # ... get a page of archive data
             _page = self.port.get_data_with_crc16(267, prompt=_ack, max_tries=self.max_tries)
             # Now extract each record from the page
@@ -2045,9 +2048,9 @@ class VantageConfigurator(weewx.drivers.AbstractConfigurator):
                           help="Dump all data to the archive. "
                                "NB: This may result in many duplicate primary key errors.")
         parser.add_option("--dump-quick", action="store_true", dest="dump_quick",
-                          help="Dump all data to the archive, fetching all data off the console"
-                               "before inserting to the archive. This can improve performance, but"
-                               "requires sufficient memory to hold all station data."
+                          help="Dump all data to the archive, fetching all data off the console "
+                               "before inserting to the archive. This can improve performance, but "
+                               "requires sufficient memory to hold all station data. "
                                "NB: This may result in many duplicate primary key errors.")
         parser.add_option("--logger-summary", type="string",
                           dest="logger_summary", metavar="FILE",
--- a/bin/weewx/drivers/vantage.py
+++ b/bin/weewx/drivers/vantage.py
@@ -2044,6 +2044,11 @@ class VantageConfigurator(weewx.drivers.AbstractConfigurator):
         parser.add_option("--dump", action="store_true",
                           help="Dump all data to the archive. "
                                "NB: This may result in many duplicate primary key errors.")
+        parser.add_option("--dump-quick", action="store_true", dest="dump_quick",
+                          help="Dump all data to the archive, fetching all data off the console"
+                               "before inserting to the archive. This can improve performance, but"
+                               "requires sufficient memory to hold all station data."
+                               "NB: This may result in many duplicate primary key errors.")
         parser.add_option("--logger-summary", type="string",
                           dest="logger_summary", metavar="FILE",
                           help="Save diagnostic summary to FILE (for debugging the logger).")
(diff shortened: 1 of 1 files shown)


bin/weewx/drivers/vantage.py
```diff
@@ -2044,6 +2044,11 @@ def add_options(self, parser):
         parser.add_option("--dump", action="store_true",
                           help="Dump all data to the archive. "
                                "NB: This may result in many duplicate primary key errors.")
+        parser.add_option("--dump-quick", action="store_true", dest="dump_quick",
+                          help="Dump all data to the archive, fetching all data off the console"
```
tkeffer, line 2048: You need a space at the end of these lines. Otherwise the `--help` command will merge the two words. 

Try it.

tkeffer (changes requested):
I had a chance to look at these changes. They work well!

One minor problem: because the `--dump-quick` command dumps everything off the console before inserting anything into the database, the command line goes dead for a long period of time. It would be useful to add some progress indicator to the `genArchiveDump()` function.

bin/weewx/drivers/vantage.py
gashalot: Indeed, I didn't even catch that when I ran my tests. Both requests merged!

tkeffer:
Hope you don't mind, but I generalized what you started, as well as added documentation.

Now you can get what you want with

```
wee_device --dump --batch-size=0
```
Default is a batch-size of 1, which is the behavior from before. Other batch sizes are possible.

PR #694, opened 2021-08-23
units.py: fix space unit label and unit format

l30n1d4:


tkeffer:
I can understand the changes for labels and formatting for `microgram_per_meter_cubed`, but why for `NONE`?

l30n1d4:
Because, in the rare cases where the value does not exist, unnecessary spaces are added before the "N/A" label. Normally, when this happens, it is because there is a communication error with the station and all current values are missing. In my opinion, the resulting template with these spaces is not perfect.

Issue #695, opened 2021-08-27
`wee_database` `-fix-strings`/`--fix-strings` fails on unpatched databases containing 'null strings'


gjr80:
`wee_database` `--check-strings`/`--fix-strings` fails on v1.0-v3.0 databases containing 'null strings' as the database cannot be patched to v4.0 due to the embedded 'null strings'. The 'null strings' cannot be removed/repaired due to the unpatched state of the database and the database cannot be patched due to the 'null strings' embedded in the database.

Refer to [this weewx-user thread](https://groups.google.com/g/weewx-user/c/HXGIfGCI4wc/m/hVuBQhLpAgAJ).

gjr80:
PR #696 should address this corner case, need to wait and see how the user with the problem fares. Not 100% happy how it turned out so happy to take on any advice, hence the RP rather than a direct commit.

tkeffer:
I guess another solution would be to catch the exception that occurs when trying to convert from string to float and, if it occurs, set the value to `None`.

This is all that `--fix-strings` does anyway. It would obviate the need for `--check-strings` and `--fix-strings`.

gjr80:
Been a while since I worked on this and my brain is not too sharp tonite. When you say '_catch the exception that occurs when trying to convert from string to float_', do you mean catching the exception when the malformed field is added to the accumulators? Will that fix a string in the archive? Can see it will keep the daily summaries string free but that was not the issue. Or do I have this all wrong - refer to my first sentence.

tkeffer:
If you go back to the original [weewx-user thread](https://groups.google.com/g/weewx-user/c/HXGIfGCI4wc/m/hVuBQhLpAgAJ) that started this, you'll see that the problem is that the user had a string in his database, which caused a `ValueError` exception. Normally, we tell the user to fix their database using `--fix-strings`. What `--fix-strings` does is search for all strings, and set them to `None`.

I'm suggesting that we suppress the original exception and, instead, set the value to `None`. Change this

```python
        #  If this is a string, try to convert it to a float.
        if isinstance(val, (six.string_types, six.text_type)):
            # Fail hard if unable to do the conversion:
            val = to_float(val)
```

to this

```python
        #  If this is a string, try to convert it to a float.
        try:
            val = to_float(val)
        except ValueError:
            val = None
```

This will also cause the "upgrade" of the daily summaries to finish cleanly.

gjr80:
[...]

PR #696, opened 2021-08-29
Fix for issue #695

gjr80:
Implemented a `check_patch` argument for `Manager` objects and their children. `wee_database` now disables patching of the database when performing `--check-strings` and `--fix-strings` actions.


gjr80:
Superseded by simpler approach at #695.

PR #701, opened 2021-09-21
simplify seasons

matthewwall:
parameterize the observations in `statistics.inc`

this results in:
* less copy/paste of html and cheetah code
* automatic display of observations so users do not have to modify the skin for 'extended' sensors


matthewwall:
accidentally made the pull request from master, not development.  lets try this again...

PR #704, opened 2021-09-24
Seasons fix

itec78:
Fixed missing unit labels on hilo bar
Improved the readability of graphs changing x-axes label spacing
Fixed year graphs x-axis: %m/%d format is wrong for EU. I removed the day

### DIFF
--- a/skins/Seasons/hilo.inc
+++ b/skins/Seasons/hilo.inc
@@ -40,7 +40,10 @@
             $archive[1].outTemp.min.format(add_label=False)</span>
         </td>
         #end for
-        <td class="units">$unit.label.outTemp</td>
+        <td class="units">
+	        $unit.label.outTemp<br/>
+	        $unit.label.outTemp
+        </td>
       </tr>
       <tr>
         <td class="label">$obs.label.heatindex</td>
@@ -72,7 +75,10 @@
             $archive[1].dewpoint.min.format(add_label=False)</span>
         </td>
         #end for
-        <td class="units">$unit.label.dewpoint</td>
+        <td class="units">
+          $unit.label.dewpoint<br/>
+          $unit.label.dewpoint
+        </td>
       </tr>
       <tr>
         <td class="label">$obs.label.outHumidity</td>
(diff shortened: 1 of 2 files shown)


skins/Seasons/hilo.inc
```diff
@@ -40,7 +40,10 @@
             $archive[1].outTemp.min.format(add_label=False)</span>
         </td>
         #end for
-        <td class="units">$unit.label.outTemp</td>
+        <td class="units">
```
matthewwall, line 43: not sure what your intent is with these extra unit entries.  the units are already specified for each observation.

skins/Seasons/skin.conf
```diff
@@ -213,6 +213,7 @@
 
     [[day_images]]
         x_label_format = %H:%M
+        x_label_spacing = 1
```
matthewwall, line 216: could you include some screen shots that show the effect of these pixel-tweaking changes?  we need to be sure they do not break other configurations.

tkeffer:
I'm not finding these changes better or worse ... just different. 

For example, why is this
image

better than this? 
image

Also, the x-axis uses `%x %X` formatting, which should work in any locale. See the Python documentation on [strftime()](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes). Is it not working for you?

itec78:
> I'm not finding these changes better or worse ... just different.

I think it's easier to read. The vertical line between 15:00 and 21:00, for example, what time is it? I have to find the number thinking about it...

> Also, the x-axis uses `%x %X` formatting, which should work in any locale. See the Python documentation on [strftime()](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes). Is it not working for you?

What I was meaning is that formatting works fine, except in year graphs that works for US only:
```
[[year_images]]
        x_label_format = %m/%d
```

skins/Seasons/hilo.inc
itec78: I think it makes sense to have every value with his unit, it's easier to read. But it's my opinion

tkeffer:
> What I was meaning is that formatting works fine, except in year graphs that works for US only:
> 
> ```
> [[year_images]]
>         x_label_format = %m/%d
> ```

OK. That makes sense. In V4.6 (not released yet), there will be the ability to specify locale-specific formatting. Find it in the `development` branch.

tkeffer:
Summary: I like your change for `x_label_format` for the year graphs, but the other changes are a matter of taste, and best left to the user.

Commit https://github.com/weewx/weewx/commit/ebd6d6e1caa2106b890dc834585abf26dcebfc8e

PR #709, opened 2021-10-27
parameterize plots

matthewwall:
* parameterize the sensor status (communications, battery status, voltages) as well as the observations
* use a list of observations to specify which plots will be shown in the seasons skin
* add lightning count to current conditions and statistics
* convert more getattr to getVar


bin/weewx/cheetahgenerator.py
```diff
@@ -766,9 +766,10 @@ class SkinInfo(SearchList):
 
     def __init__(self, generator):
         SearchList.__init__(self, generator)
-        self.SkinInfo = {}
-        for field in ['SKIN_NAME', 'SKIN_VERSION']:
-            self.SkinInfo[field] = generator.skin_dict.get(field, 'unknown')
+        self.SkinInfo = {
```
matthewwall, line 769: how is one supposed to reference an item from SkinInfo now?  before i did it like this:

```
<td class="data">$SkinInfo.get('SKIN_NAME') $SkinInfo.get('SKIN_VERSION')</td>
```

PR #713, opened 2021-11-10
Fixed always up - always down

itec78:
Always up - Always down was not printed correctly.
Code was visible, and text was not translated to local language


tkeffer:
Thanks for spotting this!

The Python formatting notation is preferred over string concatenation.

PR #715, opened 2021-11-12
acurite: enable continuous calculation of barometer

### LINKED ISSUE #714: Acurite intermittent calculation of barometer
Executive Summary: When using the Acurite driver, because barometer is calculated from absolute pressure and outside temperature, it will only be reported when both readings are available simultaneously.

[Originally posted in weewx-users](https://groups.google.com/g/weewx-user/c/pbJyKmYy3mA/m/Hsk9AWaSCQAJ)

This has been bugging me for a few years, but I just haven't been able to figure out where/how to fix it best. In the Acurite driver, there are 3 different data packets (and two variants of one of those), each updated at different intervals (18, 60, and 12*60 seconds)  Indoor air pressure is reported in R2, while outside temperature is reported in alternating R1 variants. This means that we *often* don't have a current reading for outside temperature at the same time we get a pressure reading, especially if the USB connection is being finicky, which is common with Acurite hardware.

As a result, there can be long intervals (I've seen 10+ minutes) between successful barometer calculations. When generating static webpages or otherwise working from archive data, this isn't a huge issue. When trying to do something near-real-time with loop data (for example MQTT to Home Assistant), though, this can be really annoying.

bwarden:
For Acurite systems, the hardware reports absolute pressure and
weewx calculates barometer from that and outside temperature.
Pressure is reported in R2 at 60 second intervals. Outside temperature
is reported in *alternating* R1 packets, so only at 18*2 second
intervals. So pressure and outside temperature are only available
simultaneously every 5 minutes, and weewx will only calculate barometer
when it has readings for both.

Here, we'll cache the most recent outside temperature reading. If it's
less than 60 seconds old when we get a new pressure reading, we'll insert
the cached value into the new packet, enabling calculation of barometer.

Fixes #714


tkeffer:
The problem is real, but I'd rather see it fixed in class `[PressureCooker](https://github.com/weewx/weewx/blob/development/bin/weewx/wxxtypes.py#L327)`. 

The acurite is not the only system to generate "partial packets," so these others will also be susceptible to the same problem. By putting the solution farther downstream, we can fix it for everyone.

bwarden:
In that case, would it make sense to fetch a temperature from `dbmanager` like we do for 12hr, but only in the last minute (or 5 minutes) when `record['outTemp']` is `None` here?

https://github.com/weewx/weewx/blob/644e794abd6b01fb458f757bf3c687694f48b56a/bin/weewx/wxxtypes.py#L394-L401

tkeffer:
For LOOP packets, a value off the database will be an old value. I would think you'd want to use a cached value, replenished with every LOOP packet where `outTemp` is available. Incude a check to make sure the value is not "stale."

Could you please make your changes in a branch off of `development`? It's getting very late to include new features in V4.6.

bwarden:
OK, I'll keep poking.

PR #716, opened 2021-11-16
acurite: enable continuous calculation of barometer

### LINKED ISSUE #714: Acurite intermittent calculation of barometer

bwarden:
For Acurite systems, the hardware reports absolute pressure and
weewx calculates barometer from that and outside temperature.
Pressure is reported in R2 at 60 second intervals. Outside temperature
is reported in *alternating* R1 packets, so only at 18*2 second
intervals. So pressure and outside temperature are only available
simultaneously every 5 minutes, and weewx will only calculate barometer
when it has readings for both.

Here, we'll cache the most recent outside temperature reading. If it's
less than 60 seconds old when we get a new pressure reading, we'll insert
the cached value into the new packet, enabling calculation of barometer.

Fixes #714

### DIFF
--- a/bin/weewx/drivers/acurite.py
+++ b/bin/weewx/drivers/acurite.py
@@ -447,6 +447,7 @@ class AcuRiteDriver(weewx.drivers.AbstractDevice):
         self.r3_max_fail = 3
         self.r1_next_read = 0
         self.r2_next_read = 0
+        self.cached = dict()
         global DEBUG_RAW
         DEBUG_RAW = int(stn_dict.get('debug_raw', 0))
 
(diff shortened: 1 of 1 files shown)


tkeffer:
Closed in anticipation of a more general solution.

PR #717, opened 2021-11-19
Add `try..except` to handle earlier python versions that do not support the ftplib FTP encoding parameter

gjr80:
As previously discussed. Did not see any need to elaborate in the User's Guide; if what I suspect is correct earlier python versions that do not support the encoding parameter will silently work if they encounter Latin-1 encoding.


tkeffer:
Adds a lot of lines, but I don't have a better idea!

gjr80:
Just think of them as temporary; once we support only python 4+ they can be removed :)

PR #719, opened 2021-12-02
update apt-key command to debian10 and future syntax

vinceskahan:
This updates the debian apt-key instructions to future-proof it.  We've seen multiple people run into confusion with the new debian deprecation warnings and even some failures of the key to import.  This seems to be the new way.

### DIFF
--- a/docs/debian.htm
+++ b/docs/debian.htm
@@ -25,7 +25,8 @@
 <p>Tell <span class="code">apt</span> where to find the WeeWX releases.  This only has to be done once - the first time you install WeeWX.</p>
 
 Tell your system to trust weewx.com:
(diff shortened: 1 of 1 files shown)


tkeffer:
Vince,

Can we assume people will have `gpg` on Debian systems?

vinceskahan:
'Can' have - yes dating back to at least debian-7 which is the oldest variant I have here.

'Will by default have installed' - then perhaps.   FWIW current Raspbian for the pi 'does' have it installed by default even in the lightest os they make available.

I did find that pure debian-11 minimalist vagrant and docker images do not have gpg installed, but they're perhaps edge cases.  To add them you'd do the expected:

``` sudo apt-get update && sudo apt-get -y install gpg ```

Issue #421, opened 2019-06-12
option for Log to stdout 


phinnay:
I am running WeeWX inside Docker and it's tricky to troubleshoot as all logging is dumped to syslog. I'm working around it by running a syslog daemon inside the container but this goes against best practice. 

Would it be possible to add a flag to dump all output to stdout/stderr?

tkeffer:
Yes, of course it's possible. As always, PRs are welcome.

kadaan:
@phinnay @tkeffer I've created PR #721 to address this issue.

PR #721, opened 2021-12-15
Change logger to build up the default configuration.  By default it w…

kadaan:
By default WeeWX is configured with a log handler that writes to Syslog. In some instances (ie. running in a Docker container) this is undesirable and using StdOut as the default handler would be preferred. This can be accomplished by specifying the WEEWX_LOGGING_DEFAULT_STDOUT environment variable, for example:

```bash
export WEEWX_LOGGING_DEFAULT_STDOUT=true
./weewxd
```

Solves: #421

### DIFF
--- a/bin/weeutil/logger.py
+++ b/bin/weeutil/logger.py
@@ -7,9 +7,11 @@
 
 from __future__ import absolute_import
 
+import os
 import sys
 import logging.config
 import six
+from textwrap import indent
 from six.moves import StringIO
 
 import configobj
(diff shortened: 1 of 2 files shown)


kadaan:
@tkeffer Feedback?

tkeffer:
Thanks for the PR!

Since issue #421 was created, version 4.0 has come out, which uses the Python logging facility. This makes it easy to change the logging destination. For example, to log to both syslog and `stdout`, add this to weewx.conf (NOT TESTED)

```ini
[Logging]
  [[root]]
    handlers = syslog, console
```

I should also mention that we generally favor putting switches in weewx.conf, instead of using environment variables. This puts all the configuration state in one place, making it much easier to debug remotely.

kadaan:
@tkeffer I don't think this should  be closed because this fixes the **default** log handler.  This is important for use in a Docker container so those logs are not dropped.

tkeffer:
OK, makes sense. We'll see what we can do. 

Note:

1. Logging has always been to syslog, never stdout, and I'm not inclined to change that, even on start up.
2. Version 4.6, which is still under development, does no logging before weewx.conf is loaded, for just the reason you cite. See commit https://github.com/weewx/weewx/commit/6f3f6c3339cc46a3986535d13f644d2dd2236fe7#diff-3d73a7b05810664f41c517c25b8411f8283810ad89204b631013d3c8ec9f9b24 among others.

If you're comfortable working with git, give V4.6 a try. It's in the `development` branch. If not, we'll figure something else out.

kadaan:
@tkeffer I'm fine working with from develop.  Without logging before the config is loaded how do you indicate that there are errors in the config?  Looking at the current development [code](https://github.com/weewx/weewx/blob/34140e584dbd85aaa206f772a1956881f71a39c3/bin/weewxd#L91), it seems to have the same issue.  If the config fails to load it relies on the default logger (syslog) to communicate that information.  If the plan is to not need the default log configuration before logging is loaded, then the CLI should communicate the config load failure via `print` like done a [few lines earlier](https://github.com/weewx/weewx/blob/34140e584dbd85aaa206f772a1956881f71a39c3/bin/weewxd#L81).

PR #726, opened 2022-01-04
When reports  done, shutdown engine and services

bellrichm:
I have a service that creates a thread. Because its shutDown method is not called, it is not terminating cleanly.


tkeffer:
Thanks for noticing this!

PR #728, opened 2022-01-07
implement a 'natural' sort of field names in `to_sorted_string()`

gjr80:
Default action is to sort 'naturally' where `extraTemp1` will appear before `extraTemp10`. The previous simple sort using `sorted()` is still available if `to_sorted_string()` is called with the `simple_sort` parameter set to True.

### DIFF
--- a/bin/weeutil/weeutil.py
+++ b/bin/weeutil/weeutil.py
@@ -16,6 +16,7 @@ import cmath
 import datetime
 import math
 import os
+import re
 import shutil
 import time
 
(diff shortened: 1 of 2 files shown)


tkeffer:
I can see the usefulness of this. Two suggestions:
1. Move `atio()` and `natural_keys()` to file scope. They don't use closure, so there's no reason to compile them again and again with every invocation of `natural_sort_keys()`.
2. Along the same line, compile the regular expression `re.split(r'(\d+)'` just once at file scope. Rather than have it float around on its own, you can do a trick:
    ```python
    def natural_keys():
        return [atoi(c) for c in natural_keys.compiled_re(text.lower())]
    natural_keys.compiled_re = re.split(r'(\d+)'
    ```
    so the compiled RE is unambiguously related to the function `natural_keys()`.

gjr80:
One small change  to your suggestion 2 - 5de8da4 refers.
I just ran the PR'd code on my test machine and it worked fine, I then rigged it to use the old `.sorted()` code and it put `extraTemp1`, `extraTemp10`, `extraTemp11` and `extraTemp13` in the correct order. It may be that it was another case that I struck and fixed in the GW1000 driver. Though Matthew said `ch10` was put ahead of `ch1`.

gjr80:
Ahh, just used `ch1`, `ch2`, `ch10` and `ch21`. The old sort gave me `ch1`, `ch10`, `ch2` and `ch21`. The natural sort gave me `ch1`, `ch2`, `ch10` and `ch21`. Happy it's working as it should.

PR #734, opened 2022-01-22
Loop packet weight: Possibility to use weighted loop packets

edi-x:
I have to deal with lightning data in a way where a loop packet contains the lightning strike count and the average distance of these strikes. To get a correct average distance for the archive entry these packets have to be handled as weighted events.  The accumulator already supports weighted data and so I added a possibility to use it for loop data, too. I use the lightning strike count as weight for these loop packets and the driver comes up with correct results now.

@tkeffer, do you think this is a useful feature for weewx? Can or should it be added?


tkeffer:
I'm not following what problem this is trying to solve. What is the weighting that is being applied? What is `loop_packet_weight`?

edi-x:
Well, an example could make things clearer. I receive loop packets with lightning data every minute just like this:
[dateTime=1642939200, lightning_strike_count=1, avg_lightning_distance=5]
[dateTime=1642939260, lightning_strike_count=2, avg_lightning_distance=4]
[dateTime=1642939320, lightning_strike_count=3, avg_lightning_distance=3]
[dateTime=1642939380, lightning_strike_count=4, avg_lightning_distance=2]
[dateTime=1642939420, lightning_strike_count=5, avg_lightning_distance=1]

Currently this would result in an archive package like this:
[dateTime=1642939420, lightning_strike_count=15, avg_lightning_distance=3]

However, this result is wrong. It's not enough to calculate the average lightning distance out of the five loop values (5+4+3+2+1 divided by 5 reported loop events which gives us a result of 3). The distance has to be multiplied with the number of lightning strikes that have appeared in the reported minute and the result has to be devided by the number of reported lightning strikes during the archive interval. So the right calculation would be (5*1 + 4*2 + 3*3 + 2*4 + 1*5) / 15 = 2.333.

As far as I could see the accumulator of StdArchive can solve this problem, when we use the "weight" parameter of the addRecord method there. The weight in this case would be the number of reported lightning strikes during one minute. So my driver would send the additional value for "loop_packet_weight" which has always the same value as "lightning_strike_count":
[dateTime=1642939200, lightning_strike_count=1, avg_lightning_distance=5, loop_packet_weight = 1]
[dateTime=1642939260, lightning_strike_count=2, avg_lightning_distance=4, loop_packet_weight = 2]
[dateTime=1642939320, lightning_strike_count=3, avg_lightning_distance=3, loop_packet_weight = 3]
[dateTime=1642939380, lightning_strike_count=4, avg_lightning_distance=2, loop_packet_weight = 4]
[dateTime=1642939420, lightning_strike_count=5, avg_lightning_distance=1, loop_packet_weight = 5]

The engine applies the "loop_packet_weight" as "weight" parameter for the addRecord method of the accumulator and then the calculated average for the archive packet would be correct.

tkeffer:
That makes sense.

I think I'd rather stick with the present model, where you use different "adders" and "extractors" for different observation types. Your approach results in some unwanted coupling between driver and the accumulator.  

I'd always considered the structures `ADD_FUNCTIONS` and `EXTRACT_FUNCTIONS` to be read-only (hence the upper-case names), but there's no reason why we can't allow them to be extended with specialized functions.

As an example, the accumulators have specialized functions for adding and extracting wind values. Perhaps you could come up with something for lightning? 

At that point, we can either add them permanently to `accum.py` or, more likely, have them added by whatever extension implements the lightning count.

edi-x:
@tkeffer, thanks for sharing your thoughts about this topic. I will check how I could place all the logic into the extension. Perhaps I find a way to place it completely within the driver or another idea would be to let a service handle the data in a proper way.

edi-x:
Just in case anyone is interested in the solution I'm using now:

```
class LightningAccum(Accum):

    def __init__(self, timespan, unit_system=None):
        super().__init__(timespan, unit_system=unit_system)
    
    def add_lightning_distance_value(self, record, obs_type, add_hilo, weight):
        lightning_distance_weight = weight
        if obs_type and record and 'lightning_strike_count' in record and str(obs_type) == 'lightning_distance':
            lightning_distance_weight = lightning_distance_weight * record['lightning_strike_count']
        self.add_value(record, obs_type, add_hilo, lightning_distance_weight)

weewx.accum.ADD_FUNCTIONS.update({'add_lightning': LightningAccum.add_lightning_distance_value})
```

Many thanks to @tkeffer for the implementation idea.

gjr80:
Just as a point of clarification for any Ecowitt WH57 users that may stumble on this in the future. The WH57 provides distance to the last detected strike only, not the distance to every detected/reported strike, so any weighting of WH57 lightning strike distance data based on lightning strike count will give statistically meaningless results.

PR #741, opened 2022-02-08
Fixes handling missing $gettext issues

aslak47:
I have prepared translation to Norwegian, and I found some missing $gettext occurrences


tkeffer:
Thanks for finding these!

PR #745, opened 2022-02-09
Added "from" (direction context) to language files in skin Smartphone

aslak47:
Sorry for another change.
This will be the last for a while, I hope.
Best regards, Aslak


tkeffer:
It's amazing how they can hide.

PR #748, opened 2022-02-12
Option to only generate cheetah templates once

bellrichm:
I am using the series capability to feed json data into Apache ECharts. The javascript to control the charts is created by cheetah templates. This allows the charts to be configured. Once generated, the chart javascript (and html) will not change until the configuration data has changed. The json with the data will change as the WeeWX data changes. I am using the stale_age to limit regeneration for charts with aggregation. Doing some simple profiling, I can see that generating the charts is the bottleneck for charts without aggregation. Also, since these files are regenerated they are rsync’ed again. So I'd like to only generate them once at startup, saving some processing time and a few pennies.
rich


tkeffer:
Useful addition, Rich! Thanks.

PR #765, opened 2022-04-10
added unit groups for sunshine and rain duration, cloudcover, pop

roe-dl:
pop = probability of precipitation

used for driver enhancement https://github.com/grebleem/weewx-weatherlinkliveudp/pull/26 and german forecast https://github.com/roe-dl/weewx-DWD


roe-dl:
It is the same as with `rain` and `dayRain`. There are also 2 observation types.

The Davis Vantage 2 Pro provides the daily sum of rain. To get the the rain for  the LOOP or ARCHIVE interval the driver has to calculate the difference. So that is a secondary reading. The daily sum is the primary reading.

When extending the driver for sunshine duration I thought I provide the primary sum value (as sent by the hardware) as well as the calculated difference. So it is not only the similarity to rain.

tkeffer:
If I understand you correctly, `sunshineDur` would be the amount of sunshine in a single LOOP or archive record, and `daySunshineDur` would be the *total* amount of sunshine for the whole day? So, the sum of all `sunshineDur` for the day should equal `daySunshineDur`, which would. be analogous to `rain`.

If that's the case, perhaps we could pick a better name than `sunshineDur`? Perhaps just `sunshine` and `daySunshine`? Or, `sunTime` and `daySunTime`?

roe-dl:
Yes, you understand me correctly.

I thought about how to call the observation type. And indeed I started with simple `sunshine`. But then I realized that the meaning is not clear. `sunshine` could be the amount of time but also the intensity. There are instruments that track the sun and measure the direct radiation of the sun. Like the observation type `radiation` measures the overall radiation of the sky. So I see need to distingish between sunshine duration and sunshine radiation.

`sunTime` could mean the total time the sun is up, too, I guess. For my understanding it could be misleading therefore.

And I searched the Internet for an appropriate name. `sunshineDur` someone already used.

That were my thoughts why I ended up with `sunshineDur`.

PR #767, opened 2022-04-11
systemctl stop sends 2 SIGKILL signals

bellrichm:
See, https://groups.google.com/g/weewx-user/c/Yg8OJ7uot7U/m/-QxTQr1VCwAJ

Here is some reference material I used
From, https://wiki.debian.org/systemd/Services
“Most services should use the simple type, which means a program that runs in the foreground. If your service normally runs itself in the background, search the documentation to see if it has an option to disable that. Running in the foreground is preferred.”

From, https://www.freedesktop.org/software/systemd/man/systemd.service.html
For StandardOutput
[This setting defaults to the value set with DefaultStandardOutput= in systemd-system.conf(5), which defaults to journal.](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#StandardOutput=)
For StandardError
[The available options are identical to those of StandardOutput=, with some exceptions: if set to inherit the file descriptor used for standard output is duplicated for standard error, while fd:name will use a default file descriptor name of "stderr".](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#StandardError=)
For StandardInput
[This setting defaults to null, unless StandardInputText=/StandardInputData= are set, in which case it defaults to data.](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#StandardInput=)

weewxd file
```
daemon.daemonize(pidfile=options.pidfile)
```
daemon.py file
```
def daemonize(stdout='/dev/null', stderr=None, stdin='/dev/null',
              pidfile=None, startmsg = 'started with pid %s' ):
```
```
if not stderr: stderr = stdout
```
I read this as currently stdout goes to /dev/null, stderr goes to stdout, and stdin goes to /dev/null. So I only overide the default for StandardOutput.

### DIFF
--- a/util/systemd/weewx.service
+++ b/util/systemd/weewx.service
@@ -7,10 +7,9 @@ After=time-sync.target
 RequiresMountsFor=/home
 
 [Service]
-ExecStart=/home/weewx/bin/weewxd --daemon --pidfile=/var/run/weewx.pid /home/weewx/weewx.conf
+ExecStart=/home/weewx/bin/weewxd /home/weewx/weewx.conf
 ExecReload=/bin/kill -HUP $MAINPID
-Type=forking
-PIDFile=/var/run/weewx.pid
+StandardOutput=null
 #User=weewx
 #Group=weewx
 


matthewwall:
one thing to consider is that when you run weewxd directly, it emits LOOP and REC data to stdout.  when invoked with the daemonize option, weewx does not emit this information.  in either case, weewx emits log messages to whatever log was configured withing weewx.

so when you remove the daemonize option, does a `systemctl status weewx` report all of the LOOP and REC data?

i would assume that the only thing we want to see from a systemctl status would be any error messages, more specifically, only error messages that result in a hard failure.  systemctl should indicate status, but for details users should look at the log file.

we explicitly do *not* want to introduce anything within weewx that is specific to systemd that would cause problems on other platforms.  there are many other systems that do not use systemd.

matthewwall:
another thing to test is how systemd deals with weewx when there are weewx threads that have not yet completed.

weewx should respond to INT, TERM, and KILL.  as of 4.6 it no longer responds to HUP.

INT should only happen when you are running weewx directly, e.g., when you control-c.  i *think* the intent is that one could resume after an INT (e.g., for stepping through a debugger), but that is not built into weewx.  effectively INT is like a KILL, but with a stack trace.

there are many cases where one sends TERM, but there are still cleanup tasks that must be completed.  for example, if report generation is in progress, a TERM should not shut down weewx in the middle of that report.  similarly for writing to database, e.g., using mysql on a slow connection.  in my experience, the report generation is usually the thread that is most likely to cause weewx a long time to shut down. but there are some drivers that could block a TERM until they timeout.

a KILL should be sent only if weewx is truly hung.

util/systemd/weewx.service
```diff
@@ -7,10 +7,9 @@ After=time-sync.target
 RequiresMountsFor=/home
 
 [Service]
-ExecStart=/home/weewx/bin/weewxd --daemon --pidfile=/var/run/weewx.pid /home/weewx/weewx.conf
+ExecStart=/home/weewx/bin/weewxd /home/weewx/weewx.conf
 ExecReload=/bin/kill -HUP $MAINPID
-Type=forking
-PIDFile=/var/run/weewx.pid
+StandardOutput=null
```
bellrichm, line 12: This

bellrichm:
The `StandardOutput=null` option redirects stdout to `/dev/null`. From my reading of the daemonize code and WeeWX's invocation of it, this is what currently happens. 
This change should only result in WeeWX no longer forking, which seems to make systemd `systemctl stop` behave better/correctly.

tkeffer:
> one thing to consider is that when you run weewxd directly, it emits LOOP and REC data to stdout. when invoked with the daemonize option, weewx does not emit this information.

Just to clarify: `weewxd` still prints LOOP and REC data when invoked with the `--daemon` option, but `stdout` is closed as part of the daemonization procedure, so you never see it. 

If necessary, we could test for whether the `--daemon` flag is present and, if it is, suppress the print.

tkeffer:
One other thought: I've thought of replacing the present `daemon.py` that we distribute with WeeWX, with the PEP 3143 compliant [`python-daemon`](https://pagure.io/python-daemon). It has a very permissive Apache V2 license, allowing us to include it in the distribution. 

I don't think the functionality is very different, but `python-daemon` is better known.

bellrichm:
I’m curious, why does WeeWX ‘daemonize’ when invoked via systemd? From what I read it is not necessary and my theory is the forking is causing system to send the SIGKILL signals immediately.

bellrichm:
> If necessary, we could test for whether the `--daemon` flag is present and, if it is, suppress the print.

I don’t think that is necessary.

matthewwall:
>> If necessary, we could test for whether the --daemon flag is present and, if it is, suppress the print.

> I don’t think that is necessary.

agreed.  i think the daemonize was put into the unit file due to lack of understanding of systemd conventions/behaviors.

tkeffer:
WeeWX has used `daemon` for as long as I can remember, certainly for 10+ years. Way before systemd became common.

tkeffer:
I've done some testing on my systems, and these changes seem to work correctly. 
Thanks again, Rich!

tkeffer:
Rich, did you see this email thread? 

https://groups.google.com/g/weewx-development/c/Ym9KOpYKZmE

WeeWX V4.8, using the new service file, receives a `SIGTERM` under SuSE.

tkeffer:
@matthewwall 

I was able to fix by removing the `--daemon` option from the systemd service file. I'm not sure how it's getting in there.

bellrichm:
> Rich, did you see this email thread?
> 
> https://groups.google.com/g/weewx-development/c/Ym9KOpYKZmE
> 
> WeeWX V4.8, using the new service file, receives a `SIGTERM` under SuSE.

Thanks I had not. At the moment I am puzzled.

tkeffer:
Mystery solved. According to Matthew, the install program for SuSE "inserts pieces into `weewx.service`, including a daemon flag."

He's going to create a  commit that prevents this. Hopefully, that will fix the problem.

PR #778, opened 2022-06-29
Do calculations in Cheetah templates using the #set directive

roe-dl:
I found it nowhere else explained and hope it is useful.

### DIFF
--- a/docs/customizing.htm
+++ b/docs/customizing.htm
@@ -3791,7 +3791,7 @@ $dawntodusk_vh</pre>
 #set $maxmin=$day.outTemp.max.degree_C.raw-$day.outTemp.min.degree_C.raw
 #set $maxmin_vh=ValueHelper(ValueTuple($maxmin,'degree_C','group_temperature'),formatter=$station.formatter)
 
-$maxmin_vh.format(add_label=False) Kelvin</pre>
+$maxmin_vh</pre>
 
 
         <!-- -------------------------------------------------------------- -->
--- a/docs/customizing.htm
+++ b/docs/customizing.htm
@@ -3761,12 +3761,8 @@ ephem.Eros = eros</pre>
         </p>
         <p>
             To get an observation type value in a way that can be used in formulae you can add <span class="code">.raw</span>
-            after an observation type tag.
+            after an observation type tag as described in section <a href="#formatting_options">Formatting options</a>.
         </p>
-        <p>
-            Example:
-        </p>
-        <pre class="tty">$almanac.sun.rise.raw</pre>
         <p>
             To calculate - for example - the time from dawn to dusk:
         </p>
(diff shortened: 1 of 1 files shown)


vinceskahan:
http://cheetahtemplate.sourceforge.net/docs/users_guide_html/users_guide.html#SECTION000940000000000000000 explains it pretty well

roe-dl:
But it does not explain the use of `weewx.units.ValueHelper` and `weewx.units.ValueTuple` in that circumstance. And if you do not use those classes the values are not formatted appropriately.

roe-dl:
https://groups.google.com/g/weewx-user/c/_8Jix9BIuCA

tkeffer:
I think a short section explaining how to manipulate values in the templates is a good idea, but I don't think this is the best example because there is already a tag for displaying the amount of daylight (`$almanac.sun.visible`).

roe-dl:
Ok, I will think about a better example.

roe-dl:
I thought about a better example for some time. There is already another example involving temperature, so I thought to stick to an example involving time. After all I the new example calculates the time between the beginning of civil twilight in the morning to the end of civil twilight in the evening, which is the time there is some light outside.

tkeffer:
1. The code as written would not compile --- missing parenthesis.
2. HTML issues. A `<span>` block was closed with `</code>` instead of `</span>`.
3. Explaining the `.raw` suffix is not necessary because it is already defined elsewhere in the guide. Just provide a link.
4. Types `ValueHelper`, `ValueTuple`, and tag `$almanac` are all defined elsewhere but no links are provided.
5. The prose is choppy. For example, you warn how displaying units is "tricky", but never show what's special about your code to avoid those hazards.
6. The example of difference in Kelvin (vs Celsius) is, well, tricky, because your solution won't work if the user were to innocently try it with Fahrenheit:
    `C_to_F(20 - 10) ≠ C_to_F(20) - C_to_F(10)`

roe-dl:
To item 6.: The values are converted to Celsius within the tags. So the difference is always in Celsius. And for the Celsius scale differences are marked "K" or "Kelvin" instead of "°C" by convention. But anyway, this is about how to make calculated values formattable, so unit issues can confuse there. I removed the unit and let it to WeeWX to include the appropriate unit.

To item 5: I changed the text a little bit and added some additional explanation. Is that what you looked for? Should I provide more details?

tkeffer:
Still didn't compile. The format override also had errors. 

In general, it's always a good idea to check any example code so as to avoid confusing the reader.

I fixed those problems, plus tried to improve the readability. See if you like the results.

tkeffer:
I mistakenly did that push from my own copy of the repository, instead of your fork, so it includes a few extra things. Hopefully, you'll get the idea.

roe-dl:
Yes, I like it. Thank you.

PR #789, opened 2022-08-30
Resolves #788 adding rsyslog example for CentOS journal.

### LINKED ISSUE #788: Rsyslog example to catch wee-utils logs sent to CentOS / RHEL journal
On CentOS Stream, and most likely earlier versions of CentOS and RedHat Enterprise Linux the logs for wee-utils such as `wee_import` and `wee_reports` are sent through `journald` so the `$programname` is `journal` instead of `wee_import`, etc.

The rsyslog configuration example does not account for this.

I'll add a PR to address it.

sastorsl:


matthewwall:
thank you for this.  that journald behavior has been a thorn in my side for a long time, and your use of the msg field to distinguish weewx entry points is good.  also thank you for the explicit comments about rsyslog versions and syntax.

not sure if we want to do this for general weewx use (or this merge request), but i typically send weewxd output to weewxd.log, and wee_xxx output (and output from invoking drivers directly) to weewx.log.  even more generally, if a system is using weewx-multi, then i will send output from each weewx instance to a separate log file.

PR #790, opened 2022-09-01
Resolves #788 update rsyslog example for RainerScript and CentOS journal

### LINKED ISSUE #788: Rsyslog example to catch wee-utils logs sent to CentOS / RHEL journal

sastorsl:
* Separate the weewx daemon into the `weewxd.log` file while weewx util messages are sent to `weewx.log`.
* Handle both weewx and utils messages written through the systemd journal in i.e. CentOS and RHEL.

### DIFF
--- a/util/rsyslog.d/weewx.conf
+++ b/util/rsyslog.d/weewx.conf
@@ -1,9 +1,12 @@
 #
 # Configuration options for rsyslog / log handling
+# "weewxd.log" - Messages from the weewx daemon
+# "weewx.log"  - Messages from the weewx utils
 #
 # Default to old-style Property-Based Filters syntax for rsyslog
-:programname,isequal,"weewx" /var/log/weewx.log
+:programname,isequal,"weewx" /var/log/weewxd.log
 :programname,isequal,"weewx" stop
+# Ship logs for wee-utils to a separate file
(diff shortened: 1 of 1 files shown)


matthewwall:
do you know why weewx output ends up in 'journal', while other systems that use systemd (e.g., debian) do not have this behavior?

sastorsl:
AFAIK logs are picked up by `systemd-journald` first, then sent to the `rsyslog` socket on RHEL / CentOS.
Now this seems to be the case with Ubuntu aswell (I'm running both), so the differences might be in the configuration of journald.

There are a lot of resources on the subject, but I didn't find *this* particular detail.

sastorsl:
The PR example still will manage both scenarios.

sastorsl:
@matthewwall any comments or issues with this PR?

matthewwall:
adopted at commit f2ac7502dc3a216f9d7cc9445a2fafb10231c1ab

sastorsl:
Any reason you skipped this PR?

PR #792, opened 2022-09-01
Resolves #791 Add copytruncate to logrotate and update comments.

### LINKED ISSUE #791: Logrotate rotation
The logrotate example relies on rsyslog to either discover the file to be gone or that you have to reload rsyslog.
This can be managed by the `copytruncate` directive for logrotate, which will copy the file first, then truncate it `cp /dev/null <file>`

sastorsl:
* Add copytruncate to avoid having to reload rsyslog
* Update and structure comments

### DIFF
--- a/util/logrotate.d/weewx
+++ b/util/logrotate.d/weewx
@@ -1,30 +1,44 @@
-/var/log/weewx.log {
+/var/log/weewx.log /var/log/weewxd.log {
   weekly
   missingok
   rotate 52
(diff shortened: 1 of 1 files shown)


matthewwall:
this merge request was superceded by a few commits, ending with 87046445

my apologies for not incorporating this merge request directly, but i think i captured all of the critical details, and i had to clarify a few more things in the comments.

PR #793, opened 2022-09-08
Changed translation of moon phases

aslak47:
Minor changes


tkeffer:
Some of these were changed in PR #764. Is this the final version?

aslak47:
Yes, the new version is better.

PR #801, opened 2022-10-23
Fixes pgettext example in docs

Daveiano:
The `pgettext` code example actually uses `gettext` not `pgettext`.


tkeffer:
Thanks for spotting this.

Issue #646, opened 2021-02-05
NOAA text files not displayed as UTF-8


tkeffer:
If run through a webserver, the two-byte utf-8 encoding of the degree sign is being interpreted as two separate characters encoded in Windws-1252, that is, the character 'Á' and the character '⁰'. 

See the weewx-user thread starting here: https://groups.google.com/g/weewx-user/c/lQ-Q_9Zd04s/m/eEZCYAaSAwAJ

tkeffer:
I see two options:
1. Encode in Windows-1252 (instead of UTF-8). Pro: simple. Con: is this always the default encoding used by all browsers?
2. Instead of formatting a simple text file, format an HTML file with `<pre>` tags. Pro: HTML is what all browsers expect, so this will format correctly on all browsers. Con: requires replacing the Cheetah templates for NOAA files.

I'm inclined to go with the first option.

tkeffer:
The problem with encoding as Windows-1252, is that, depending on what characters it uses,  it may not be possible to encode the user's station location. 

So, now I'm thinking a third option, `normalized_ascii` may be best.

tkeffer:
Went with `normalized_ascii`. Fixed in commit https://github.com/weewx/weewx/commit/cb0d0e50362e84dd64a2b6755af40ee6ff8c9662

PR #804, opened 2022-10-29
 skin.conf  in seasons skin about NOAA enconding

metfm:
enconding in NOAA   [[SummaryByMonth]]  and  [[SummaryByYear]]   must be utf-8, because in special characters like greek for example, in noaa reports the name of the station dosent apears.


tkeffer:
Please see issue #646.

In short: the default for many browsers is not utf-8. For those browsers, your fix will cause strange characters to appear.

Of course, you can always modify your own personal copy.

PR #807, opened 2022-11-05
make formatting to decimal hours possible

roe-dl:
This PR makes it possible to format `group_deltatime` values to decimal hours, not only hours, minutes, and seconds.

* `$almanac.sun.visible` still displays a result like "9 hours, 30 minutes, and 0 seconds".
* `$almanac.sun.visible.hour` displays a result like "9.5 hours" if this PR is applied.

I see no problem to existing skins as `$almanac.sun.visible.hour` displayed nothing useful so far.


tkeffer:
I am not very keen on this approach.

The unit in the `ValueTuple` is simply the unit the value is measured in. It should have nothing to do with how it is displayed. It is best to keep these two concepts separate.

Instead, it is the job of the `format` parameter to choose how the value should be displayed.

Right now, the function `delta_secs_to_string()` decomposes an elapsed time into days, hours,
minutes, and seconds, which are then used to interpolate a proposed format using keys `day`, `hour`, `minute`, and `second`.

An alternative to your approach would be to add other possibilities to the interpolation
dictionary, such as `total_hours`. So, the format would look like:

```python
"%(total_hours).2f%(hour_label)"
```

roe-dl:
I thought on `$current.outTemp.degree_C` vs. `$current.outTemp.degree_F` vs. `$current.outTemp` when I looked for a solution. So I think `$almanac.sun.visible.hour` should at least work in some way. But it actually does not.

But ok, I will think about another solution, involving `delta_secs_to_string()`.

tkeffer:
> I thought on $current.outTemp.degree_C vs. $current.outTemp.degree_F vs. $current.outTemp when I looked for a solution. So I think $almanac.sun.visible.hour should at least work in some way. But it actually does not.

Not an unreasonable assumption. Let me think about this a bit. It may take a few days
because I will be traveling.

roe-dl:
In the meanwhile let me present another solution to choose from.

Here the format string defines which way to use. If the format string contains `%(` (which indicates that it contains clauses like `%(hour)d`) then `delta_secs_to_string()` is used. Otherwise `locale.format_string()` or `format_string % val_t[0]` is used.

If `delta_secs_to_string()` is to be used and the unit is not `second` the value is converted to seconds.

And last but not least if there is no given format string and the context is `day`, `week`, `month`, or `year` the context is converted to something reasonable before looking up the default format string.

roe-dl:
Real live example: https://www.woellsdorf-wetter.de/photovoltaics/

Press "Dezimalstunden" (for "Decimal hours") in the middle of the page to toggle display between decimal hours and hours:minutes:seconds in the table.

roe-dl:
As this does not merge any more, I write my second proposal to this comment.

current content of `units.py` at line 758:
```python
            # Now format the delta time, using the function delta_secs_to_string:
            val_str = self.delta_secs_to_string(val_t[0], format_string)
            addLabel = False
```

proposed change:
```python
            if '%(' in format_string:
                # Now format the delta time, using the function delta_time_to_st
                val_str = self.delta_time_to_string(val_t, format_string)
                addLabel = False
            elif localize:
                # Localization requested. Use locale with the supplied format:
                val_str = locale.format_string(format_string, val_t[0])
            else:
                # No localization. Just format the string.
                val_str = format_string % val_t[0]
```

tkeffer:
I think there's a better way. The analogy is not `$current.outTemp.degree_C`, but
`$current.windDir.ordinal_compass`. 

The tag `$day.sunshineDur.sum` would give you `19,200`. The tag `$day.sunshineDur.sum.hour` would give you `5.33`. The tag `$day.sunshineDur.sum.long_form` would give you `5 hours, 20 minutes`.

Of course, this would break backwards compatibility. Every tag that involved `group_deltatime` would need a `.long_form` appended to it. To avoid this, we could go the opposite way. For decimal values: `$day.sunshineDur.sum.decimal` and `day.sunshineDur.sum.hour.decimal`. That wouldn't break any skins, but it feels unnatural.

roe-dl:
May be we could define both `.long_form` and `.decimal` and set a default in `weewx.conf` which is `decimal` for new installations and `long_form` for updated installations. So it would not break existing skins. But I am not sure if this is a good idea either.

roe-dl:
Another thought.

And what if you see the long form and the decimal form like `"%.1f"` vs. `"%+.1f"`? One of them presents the sign (plus, minus) for negative values, only, and the other one prepends the sign everytime. You can change the output in a wide range using those format strings.

I tried to keep it small and simple: Using a format string the user sets the look and feel of the value. If they include time parts in the format string, the value is divided into days, hours, minutes, and seconds. If they use a simple number format string, they get a simple number. No special explanation needed. All kinds of format strings work. That was behind my thinking.

tkeffer:
Here's a candidate fix. Branch `long_form`, a branch off `development`

https://github.com/weewx/weewx/tree/long_form

roe-dl:
I think that is a possible solution. And I hope people will not be that much "unamused" about the breaking change. It will affect almost all skins, as almost all use `$almanac.sun.visible`.

tkeffer:
Yes, it will, but the fix is simple. Even without the fix, the results are correct, just hard to understand.

roe-dl:
So, what is the next step? Do you want me something to do?

tkeffer:
I think we're done. Thanks, Karen.

PR #816, opened 2022-11-26
add $to_bool and $to_list to the description of helper functions

roe-dl:
Added the missing functions `$to_bool` and `$to_list` to the list of Cheetah helper functions.


tkeffer:
Thanks, Karen.

PR #817, opened 2022-11-30
allow the weewx.UnknownBinding exception to bubble up

bellrichm:
Currently a bad `data_binding` with the `$current` tag does not error. It actually results in the ‘tag line’ being in the generated file.
So for example if `$current($data_binding="bad_binding").outTemp.format(add_label=False, localize=False)` is in the template, the generated file will have `$current($data_binding="bad_binding").outTemp.format(add_label=False, localize=False)`.
This allows the `weewx.UnknownBinding` exception to bubble up and be logged as an error. This the same behavior as something like `$day($data_binding="bad_binding").outTemp.avg.format(add_label=False, localize=False)`

### DIFF
--- a/bin/weewx/tags.py
+++ b/bin/weewx/tags.py
@@ -567,28 +567,23 @@ class CurrentObj(object):
         else:
             # A binding has been specified, or we don't have a record, or the observation type
             # is not in the record, or the timestamp is wrong.
-            try:
-                # Get the appropriate database manager
-                db_manager = self.db_lookup(self.data_binding)
-            except weewx.UnknownBinding:
-                # Don't recognize the binding.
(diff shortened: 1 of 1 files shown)


tkeffer:
It's certainly reasonable to have the behavior of `$current` be the same as `$day` and others, but, unfortunately, this change breaks the test suites, in particular, test `test_templates.py`. You'll have to catch the resultant exception.

tkeffer:
I ended up fixing this by going the other way: an aggregation tag with an unknown binding now swallows the exception and displays the offending tag in the output.

Commit https://github.com/weewx/weewx/commit/592faa16d38a473bb1f64a2515cb9d740ea3692c

PR #821, opened 2023-01-05
convert suse.htm and debian.htm to md

vinceskahan:
Similarly, rename htm=>md, update content, fix up .yml


vinceskahan:
oops, I forgot GH will update a pre-existing PR if you push more to the branch being PR'd from.  So this one does two quickstarts.  I'll change the title to match.

tkeffer:
Same issue. Both the Debian and SuSE instructions include Python 2.

PR #827, opened 2023-01-09
Converted existing user guide and troubleshooting guides to markdown. …

djenk41us:
This is based on the v4.9x code line. I broke up the User Guide and the WeeWX.conf documentation into a series of markdown documents for easier management.


tkeffer:
Thanks, Doug.

Having gone through this myself with the Customization Guide, I know how much work it is.

PR #831, opened 2023-01-15
Add option to set decimal character for csv import

c0by85:
Add option to set decimal character for csv import. Adopting the option from [wd-example.conf](https://github.com/weewx/weewx/blob/master/util/import/wd-example.conf)

https://groups.google.com/g/weewx-user/c/QV4UKRnR9v0


gjr80:
Thanks. I will separately add so extra words to the relevant section of the Utilities Guide.

PR #832, opened 2023-01-15
support to configure wunderground api urls

gschaden:
support to change the url for the wunderground api in the config file


tkeffer:
Sorry, but I'm not following. You can already do something like this:

```
    [[Wunderground]]
        server_url = https://your.custom.api
```

What does this PR add? Am I missing something?

Issue #833, opened 2023-01-19
for aggregation of `wind` the reading of `windGust` is used for `max` but `windSpeed` for `maxtime`


roe-dl:
In `xtypes.py` the derived observation type `wind` is replaced by `windGust` for the aggregation type `max` and by `windSpeed` for other aggregation types (see line 311). This means `...wind.max` is based on observation type `windGust` but `...wind.maxtime` is based on `windSpeed`. Is that intended?

tkeffer:
You have an extraordinary eye for catching these things. I guess we want

``` python
        if obs_type == 'wind':
            sql_type = 'windGust' if aggregate_type in {'max', 'maxtime'} else 'windSpeed'
        else:
            sql_type = obs_type
```

roe-dl:
> You have an extraordinary eye for catching these things. I guess we want

Sometimes people called me autistic. May be they are right a little bit. But in this case it was not me but @hoetzgit who found out.

roe-dl:
I thought about other aggregation types that may be affected, but I found none. So I guess I send a PR about this.

PR #834, opened 2023-01-20
'windGust' for 'maxtime', too. #833

roe-dl:


tkeffer:
Thanks, but this is the easy part. It really needs a regression test and an entry in the change log.

roe-dl:
I included the change in my running installation of WeeWX. There is no issue so far.

roe-dl:
@hoetzgit included this change into his running WeeWX installation, too. It is working well so far.

PR #838, opened 2023-01-28
switch RH pip to systemd, add pip install wheel to Preparation

vinceskahan:
Added RH tab to Preparation to mention installing wheel, switch to systemd and mention selinux command to run in RH tab for 'run as a daemon'.  Tested on rocky/9 in a vagrant vm.

### DIFF
--- a/docs/quickstarts/pip.md
+++ b/docs/quickstarts/pip.md
@@ -32,6 +32,15 @@ install WeeWX:
     python3 -m pip install wheel
     ```
 
+=== "Redhat"
+    ```shell
+    sudo yum install -y gcc
+    sudo yum install -y python3-devel python3-pip
+    # This makes the install of pyephem go more smoothly:
+    python3 -m pip install wheel
+    ```
+
+
 ## Installation steps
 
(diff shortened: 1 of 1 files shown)


tkeffer:
I restructured it slightly. See if I got it right: https://weewx.com/v5-docs/quickstarts/pip/#run-as-a-daemon

vinceskahan:
Lets go with it as-is now and see what other testers report if anything...

Just for future reference, I started basically there and later decided otherwise.
* if they did a root install (not --user), chcon is not needed
* in that case they are installing to system dirs and $HOME/.local/bin might not even exist
* and even then, selinux might be manually set enabled yet permissive, in which case chcon again is not needed

You only need the chcon if selinux is enabled+enforcing (the os default) and if files are copied into unexpected locations (ala, a --user installation of things running from systemd out of a $HOME)

selinux is painful on the best of days.

That said, your instructions above 'do' say they should use --user so it's likely ok as edited, but I wanted to fill in the gory details on why my original PR was structured as it was.  No action required.

Issue #733, opened 2022-01-22
Consider allowing `wee_import` to update existing archive records with imported data


gjr80:
At present `wee_import` only supports saving imported data as new archive records. If an imported record already exists in the archive (ie the `dateTime` fields are equal) the imported record is discarded.

The ability to update existing records would support the use case where a user wishes to amalgamate historical data from another station or source, perhaps with different types of sensors, with existing WeeWX data.

[This weewx-user thread](https://groups.google.com/g/weewx-user/c/afPbV04SDjs) refers.

PR #839, opened 2023-01-29
Add option --update to wee_import

jocelynj:
When specified, data in database with the same dateTime is updated with new values from import. Option can also be used when importing a different column of database from a different csv file.

Linked to #733.


tkeffer:
Gary: the CI test failed because of a problem in the test suite, not the PR. It was fixed in branch `V5`, but not in `development`.

jocelynj:
I'm using this patch locally, and it works fine for data directly coming from the CSV import.

I've seen that computed fields, like barometer from pressure, are not recomputed. Should they be recomputed?

tkeffer:
No longer necessary. PR #930 covers it.

PR #843, opened 2023-02-04
deltatime of `None` is raising an error instead of outputting "N/A" in `long_form()`

roe-dl:
If the deltatime value is `None`, the new funktion `long_form()` is raising an error. To output "N/A" (or what is configured to be used in this case) I re-used the first lines of `format()` and added the parameter `None_string`.

### DIFF
--- a/bin/weewx/units.py
+++ b/bin/weewx/units.py
@@ -791,7 +791,7 @@ class Formatter(object):
         _sector = int(_degree / _sector_size)
         return self.ordinate_names[_sector]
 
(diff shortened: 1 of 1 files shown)


tkeffer:
Can you give me an example of where `long_form` raises an error if given a value of `None`? There is a unit test for this, and it seems to be working fine. 

See https://github.com/weewx/weewx/blob/master/bin/weewx/tests/test_units.py#L203

tkeffer:
My apologies. Now I see the problem.

tkeffer:
Thanks, Karen!

PR #850, opened 2023-03-02
Update xtypes.py

mKainzbauer:
This change will enable calculation avg_ge for obs_types with percent unit

### DIFF
--- a/bin/weewx/xtypes.py
+++ b/bin/weewx/xtypes.py
@@ -535,6 +535,8 @@ class DailySummaries(XType):
                     val += ("group_temperature",)
                 elif val[1] in ['inch', 'mm', 'cm']:
                     val += ("group_rain",)
+                elif val[1] in ['percent']:
+                    val += ("group_percent",)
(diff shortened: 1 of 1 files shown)


tkeffer:
The `if`/`else` hack that you are patching is strictly for backwards compatibility of very old skins (before about 2010). Back then, `ValueTuple`s had only two member (value, unit), now they have three (value, unit, unit group).

So, if you form the `ValueTuple` correctly with 3 values there should be no need for this hack.

I'm certainly not an expert on your skin, but taking a quick look, I see on line [276](https://github.com/brewster76/fuzzy-archer/blob/master/bin/user/historygenerator.py#L276) of `historygenerator.py` that `ValueTuple`s are being formed with only 2 members. Start forming `ValueTuple`s with three members (as they should be) and the problem will go away.

I imagine the only reason your skin has worked so far is because you've restricted its use to `group_temperature` and `group_rain`, which are covered by the hack. However, when your user added `group_percent` it failed.

mKainzbauer:
Thanks for clarification, I'll find the issue in the skin and get it done the right way.

PR #872, opened 2023-06-11
RedHat / Fedora RPM packaging, systemd fixes

HQJaTu:
Changes:
- Made `setup.py` run on Python 3
- Added missing skin files for HTML-report
- Added RedHat 9 into RPM `.spec`
- Separated systemd `.service` Suse and RedHat to remove need of altering the file
- Fixed RPM build to produce a working `.rpm` for RedHat 9 and Fedora 38


tkeffer:
Jari,

Thanks for the PR. You've clearly dug into the code base, and that's good!

Here are my thoughts:

1. `setup.py` has worked with Python 3 for years. I think you worked off an old version. Not sure where you got it from.
2. In any case, with version 5, we are moving away from `setup.py` and towards `pip`, so this will soon be obsolete.
3. The place for `CONF_FILE`, `WEEWX_BINDIR`, etc., is in `/etc/default/weewx`, not the systemd unit files. 
4. I'm not keen on having separate systemd unit files for each operating system. What problem are you trying to fix here that isn't already done by the default file?

Thanks again.

-tk

HQJaTu:
>     1. `setup.py` has worked with Python 3 for years. I think you worked off an old version. Not sure where you got it from.

This must be on me. I have an old fork, sometimes rebasing doesn't go so well.

>     2. In any case, with version 5, we are moving away from `setup.py` and towards `pip`, so this will soon be obsolete.

Yes, `pyproject.toml` seems to be nearly mandatory nowadays.

>     3. The place for `CONF_FILE`, `WEEWX_BINDIR`, etc., is in `/etc/default/weewx`, not the systemd unit files.

In my Fedora, there is nothing in `/etc/default/`.
See: https://pagure.io/packaging-committee/issue/152

>     4. I'm not keen on having separate systemd unit files for each operating system. What problem are you trying to fix here that isn't already done by the default file?

As I run my Weewx on a Fedora, I package and run the software on that. The `sed`-thingie is bit kludgy and really doesn't offer the possibility of supporting Systemd on a chosen distro.

Here are some issues I plan on working in the future:
* Running as non-_root_
  * Editing `.service`-file isn't especially good approach
* As RHEL / Fedora deploys to `/usr/share/`, having `RequiresMountsFor=/home` isn't needed
  * Actually, security hardening would add `ProtectHome=true`, which is complete opposite
* Added capabilities with `CapabilityBoundingSet`

Remember, I do run this on my machine, the plan is to squeeze the permissions to bare needed. Hardening Systemd and SElinux is something I''ve done on other projects. Any of those changes/additions are likely to break on some other distro and `sed` simply won't cut it.

matthewwall:
hi jari,

we do not want a separate systemd unit file for each operating system, and we are not going to do anything in the weewx code base that would cause weewx to depend on systemd. the unit file should contain only the minimal settings required, and those should work on every system that has systemd.

the mechanism we use to customize the sysV init scripts is /etc/default/weewx. we intend to use that for customization of the systemd unit files as well.  i know that redhat uses /etc/sysconfig in addition to /etc/default, so we might use /etc/sysconfig/weewx instead of /etc/default/weewx on redhat systems. but the point is that we do *not* want users to have to modify the unit file.

some of these items to customize include:

- which user runs weewxd, e.g., root or weewx
- which python to use
- which weewx config file to use

by default, weewx v4 runs as the root user. in weewx v5 we intend to make weewx run as the 'weewx' user, at least on systems where weewx is installed using apt, yum, or zypper. i am not yet certain whether this will be possible, since permissions are often a problem with hardware devices.

the `RequiresMountsFor=/home` does not belong in the weewx unit file on a redhat install (or debian or suse).

we have not supported selinux, nor do we specify any systemd-specific process limitations.  in fact, most of the time i suggest that users disable selinux completely.  it just causes too many problems when working with hardware devices, at least for the target audience for weewx.  if you would like to write a wiki page that explains how to make selinux not get in the way of running a weather station, that would be wonderful!  or perhaps a wiki page that explains all of the CAP settings for CapabilityBoundingSet that one must specify for various weather station hardware?

i'm not sure what you mean by 'that sed thingy'. are you saying that sed should not be used when building a redhat rpm? are you saying that you want to be able to build redhat packages for weewx on a system that is running selinux?

m

HQJaTu:
Well ... in my own projects, I do tailor SElinux policies for the stuff I create and run. If possible, the services are always non-_root_ and have Linux capability-restrictions in effect to maximize security.

Without adjectives:
I know this stuff and happily accept the challenge to make my own Weewx run as non-root. However, to achieve such goal specific actions are needed both in `.spec` and `.service` -files. These changes will not work on any other distro, that's for sure.

It's obvious, my goal for security conflicts with Weewx-project's current design philosophy where aim is to keep things as simplistic as possible. To me Weewx is a valuable piece of software, which I've been running for years and plan on running for years, I'm simply offering the changes I made to be available for others to use.

PR #874, opened 2023-07-07
Make barometer algorithms available to the user

roe-dl:
This refers to https://groups.google.com/g/weewx-user/c/ViQusNGKCIA

WeeWX includes various sealevel pressure calculation algorithms in `wxforumulas.py` and `uwxutils.py`, but the user cannot choose between them. IMHO it would be interesting to see the different results and compare them. 

`get_scalar` knows the parameter `option_dict`, but it is not used. So no thing like `$current.barometer(algorithm='vaManBar')` exists. Therefore separate observation types for all the algorithms are used:

* `barometer`: sent by the hardware or calculated using the formula in `wxformulas.py` as before (no change)
* `barometerUnivie`: from `uwxutils.py` University of Vienna algorithm
* `barometerDavisVp`: from `uwxutils.py` Davis Vantage Pro algorithm
* `barometerManBar`: from `uwxutils.py` WMO algorithm
* `barometerDWD`: from `uwxutils.py` German Weater Service DWD algorithm

### DIFF
--- a/bin/weewx/tests/test_wxxtypes.py
+++ b/bin/weewx/tests/test_wxxtypes.py
@@ -155,6 +155,7 @@ class TestPressureCooker(unittest.TestCase):
             mock_mgr.assert_called_once_with(self.record['dateTime'] - 12 * 3600, max_delta=1800)
             self.assertEqual(t, (80.3, 'degree_F', 'group_temperature'))
 
+        pc = weewx.wxxtypes.PressureCooker(altitude_vt)
         # Mock a database in METRICWX units
         with mock.patch.object(db_manager, 'getRecord',
                                return_value={'usUnits': weewx.METRICWX,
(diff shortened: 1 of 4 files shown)


roe-dl:
Example diagram

image

tkeffer:
Nice to make these other algorithms accessible to the user, but introducing new types is not the way to do it. 

What was the problem with using `option_dict`? That's what it's there for.

roe-dl:
> What was the problem with using option_dict? That's what it's there for.

* The parameter `option_dict` exists, but `get_scalar` is called without that parameter all the way in WeeWX. 
* `CurrentObj` has `__getattr__` only, no `__call__`.
* `TimespanBinder` has `__call__`, but it is used to change the timespan, not to alter the observation type.
* `ObservationBinder` does not hold information about an option_dict to be applied.
*  So actually observation types cannot have parameters like `$current.barometer(algorithm='paManBar')`. WeeWX actually looks at observation types as pure readings that cannot have parameters.
* `get_aggregate()`'s `option_dict` is actually connected to the aggregation type only. So what, if both the observation type and the aggregation type have attributes?

I guess my knowledge is not deep enough to solve all these issues.

tkeffer:
I am not keen on introducing a new observation type for each algorithm. It solves a very narrow problem and could lead to a proliferation of types.

This seems like a perfect opportunity for a bunch of xtypes.

roe-dl:
I thought further. You are right. This makes no sense. If you start with algorithms, the next thing one may want is using a different altitude in calculation. It cannot be solved by introducing more observation types. The only way is to fully implement `option_dict` for `get_scalar`.

Meanwhile I provided an XType extension for those who want to experiment, as you suggested. You find it [there](https://github.com/roe-dl/weewx-GTS/blob/master/bin/user/barometer.py).

PR #879, opened 2023-07-16
Fix bug that broke stale image generation.

bigwave:
Image generation is skipped if only the "stale age" condition is met.


tkeffer:
Thanks so much for this PR! It's amazing this bug survived this long!

PR #889, opened 2023-09-13
add ability to set debug level via weectl

vinceskahan:
This PR lets the user also set their debug=N level via weectl CLI, following the precedent for some other items already configurable that way.

### DIFF
--- a/bin/weecfg/station_config.py
+++ b/bin/weecfg/station_config.py
@@ -32,6 +32,7 @@ def station_create(config_path, *args,
                    dist_config_path=None,
                    weewx_root=None,
                    docs_root=None,
+                   debug=None,
                    examples_root=None,
                    user_root=None,
                    dry_run=False,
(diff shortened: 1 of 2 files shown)


tkeffer:
Why `debug` and not, say, `log_success` or `log_failure`? Or, set up the Weather Underground uploader? Where do we draw the line?

It's not really part of a station configuration.

vinceskahan:
Personally I think that 'ideally' anything in core weewx should be (re)configurable via CLI including many things not currently configurable that way....but I'm a bit of an absolutist on CLI interfaces admittedly.

In this case, we almost always ask new weewx users to set their debug level up as they run into issues getting their feet wet, and new unix users also frequently have issues even editing a simple text file to do so.

This PR gives them a one-liner way to alter their debug level and removes a possible confusion/error vector for this kinds of users so theoretically we'd see less "how do I edit a file" type new users questions.

Your call.  You can of course always close it as 'will not implement' if you feel that way.

tkeffer:
Sorry, buddy, but I'm not seeing a compelling need.

PR #893, opened 2023-10-04
Implement `Cumulative` Xtype

gjr80:
This PR implements the `Cumulative` Xtype to allow the `cumulative` aggregate to allow resetting the cumulative total at time(s) specified by the `reset` option. This in turn allows users to specify a 'reset' time for cumulative plots at which time the cumulative plot value will be reset to zero.

The `reset` option is a date-time specification for times within the overall `cumulative` aggregate time span at which time the cumulative value will be reset to zero. The `reset` option is an optional string in the format `[mm-][dd][T]HH:MM`. A number of keyword values are also supported:
- `midnight` - reset at midnight each day
- `midday` - reset at midday each day
- `day` - reset at midnight each day
- `month` - reset at midnight at the end of each month (aka midnight at the start of the 1st of each month)
- `year` - reset at midnight at the end of each year (aka midnight at the start of 1 January each year)

The default is no reset.

The `Cumulative` Xtype is fully backwards compatible with the existing `cumulative` aggregate. Whilst the initial version of this PR includes a fully functioning replacement for the `cumulative` aggregate, there are one or two additional features still to be added. The purpose of submitting this PR now is for review of the `Cumulative` Xtype implementation/structure.

### DIFF
--- a/bin/weewx/tests/test_series.py
+++ b/bin/weewx/tests/test_series.py
@@ -34,7 +34,6 @@ month_stop_tt = (2010, 4, 1, 0, 0, 0, 0, 0, -1)
 start_ts = time.mktime(month_start_tt)
 stop_ts = time.mktime(month_stop_tt)
 
-
 class VaporPressure(weewx.xtypes.XType):
     """Calculate VaporPressure. Used to test generating series of a user-defined type."""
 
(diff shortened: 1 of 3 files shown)


tkeffer (commented):
Will also need to put something in the documentation.

tkeffer:
Just had a thought. What if you calculated the series using `sum`, then patched it up later? This way, you could do it in one pass, and you wouldn't have to replicate so much code.

gjr80:
Inadvertently closed this PR when I deleted the source fork (whoops!). I will get back to this IDC.

PR #896, opened 2023-10-28
handling UnknownType in property `raw`

roe-dl:
Sometimes webpage creation fails with the following messages:
```
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: Evaluation of template ... failed with exception '<class 'TypeError'>'
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: **** Ignoring template ...
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: **** Reason: 'UnknownType' object is not subscriptable
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: ****  Traceback (most recent call last):
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: ****    File "/usr/share/weewx/weewx/cheetahgenerator.py", line 348, in generate
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: ****      unicode_string = compiled_template.respond()
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: ****    File "...", line 753, in respond
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: ****    File "/usr/share/weewx/weewx/units.py", line 1094, in raw
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: ****      return self.value_t[0]
Oct 28 13:01:05 XXX weewx[382617] ERROR weewx.cheetahgenerator: ****  TypeError: 'UnknownType' object is not subscriptable
```

I thought it would be an idea to return `None` in such cases.

### DIFF
--- a/bin/weewx/units.py
+++ b/bin/weewx/units.py
@@ -1091,7 +1091,10 @@ class ValueHelper(object):
     @property
     def raw(self):
         """Returns just the data part, without any formatting."""
-        return self.value_t[0]
+        try:
+            return self.value_t[0]
+        except (TypeError,LookupError):
+            return None
 
     def convert(self, target_unit):
         """Return a ValueHelper in a new target unit.


tkeffer:
One could argue that it is an error to perform any operation on an instance of `weewx.units.UnknownType`, other than ask for a string representation. Indeed, that's what the [documentation for `ValueHelper`](https://github.com/weewx/weewx/blob/V5/src/weewx/units.py#L997) says.

What should the semantics be for the following? `None`, or an error?

```
$current.foobar.raw
```

roe-dl:
The error message `'UnknownType' object is not subscriptable` and `return self.value_t[0]` provides no hint to the user, where to look for the problem. `None` is often the value used if no value is present. So this seemed to me a possibility.

If raising an exception is the better solution, it could be something like this:
```
    @property
    def raw(self):
        """Returns just the data part, without any formatting."""
        if isinstance(self.value_t, UnknownType):
            raise TypeError(self.value_t.obs_type)
        return self.value_t[0]
```

tkeffer:
I dunno. There are other places where `value_t` is subscripted. For example, `round()`, `has_data()`.

AFAIK, this issue has never come up. I'm reluctant to clutter to improve the readability of an error that doesn't seem to happen very often.

roe-dl:
Several times I spent hours to find the reason for this exception. I ended up enclosing all the `.raw` in `#try` - `#end try`. It is annoying to do this in a large template, but it is possible. So I will go and do so.

Issue #867, opened 2023-05-05
Set locale when using internationalized skins


tkeffer:
Right now, we depend on the user setting the `LOCALE` environment variable before running reports. This works, but it means all reports use the same locale.

Alternatively, we could optionally set the locale within the program before running each report.

``` python
import locale
...
locale.setlocale(locale.LC_ALL, 'es_ES')
```

Unfortunately, the locale used in `setlocale()` requires a country code as well as a language code, so we can't use option `lang` in `skin.conf` (which is only a language code). So, we need to either add a country code to `lang`, or we need another option `locale`.

tkeffer:
Implemented in commit https://github.com/weewx/weewx/commit/bbaf514fe8c3d1692cc759620e7ea00d66ef0225

mKainzbauer:
Yes! 😀

tkeffer:
Glad you like it!

See [_Changing languages_](https://github.com/weewx/weewx/blob/development/docs_src/custom/custom-reports.md#changing-languages) for how it works.

PR #897, opened 2023-11-11
Set locale when using internationalized skins

mKainzbauer:
Set locale when using internationalized skins

See https://github.com/weewx/weewx/issues/867

This commit enables a 'locale = xx_XX' configuration for a report. If the locale is set and available, it will be used for the report, formatting will be done according to this locale.
If the locale is set and not available, an error message will be logged. If the locale is set and no 'lang = xx' config exists, 'lang' will be derived from the locale, even if the
locale is not available.
If the locale is set and a 'lang = xx' config exists, 'lang' be set to this value and not be derived from the locale.

### DIFF
--- a/src/weewx/reportengine.py
+++ b/src/weewx/reportengine.py
@@ -14,6 +14,7 @@ import os.path
 import threading
 import time
 import traceback
+import locale
 
 # 3rd party imports
 import configobj
@@ -142,6 +143,18 @@ class StdReportEngine(threading.Thread):
                 log.error("   ****       Report ignored")
                 continue
 
+            # Save the current locale
+            default_locale = locale.getlocale()
+
+            if 'locale' in skin_dict:
+                try:
+                    # Set locale if supported
+                    locale.setlocale(locale.LC_ALL, (skin_dict["locale"]))
+                except locale.Error as e:
+                    # Inform the requested locale couldn't be set
+                    log.error("Confguration error: %s", e)
+                    log.error("   ****       Couldn't set locale to '%s', previous setting '%s' wasn't changed.", skin_dict["locale"], default_locale)
+
             # Default action is to run the report. Only reason to not run it is
             # if we have a valid report report_timing, and it did not trigger.
             if self.record:
@@ -219,6 +232,9 @@ class StdReportEngine(threading.Thread):
             # Restore the current working directory
             os.chdir(cwd)
 
+             # Restore the previously set locale
+            locale.setlocale(locale.LC_ALL, default_locale)
+
 
 def _build_skin_dict(config_dict, report):
     """Find and build the skin_dict for the given report"""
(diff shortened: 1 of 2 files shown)


tkeffer:
Thanks! This is a very useful addition!

A few comments:

1. I wonder if we should deprecate the use of `lang` and just use `locale`? The language code would first search and read a language file specified by the first two characters of `locale`, then override this with the language and country code. That is, first `de` would be read, then `de_at`, allowing any Austrian variants to override whatever is found in `de`.
2. Thanks for including tests. However, I would add `locale` to the test template `StandardTest/metric`. Right now, the template tests for language, but setting `locale` would allow testing for the effects of setting a locale, including numbers formatting.
3. What is the relationship between your addition, and the call to `locale.setlocale()` found in `user/extensions.py`.
4. We will need some documentation.

Thanks again!

mKainzbauer:
Thank you for your comments, I'll give my best to contribute my thoughts on your points:

1. As you might know, I maintain the fuzzy-archer (a.k.a. Bootstrap) skin, which is using several JS libraries to do all this fancy live data stuff. One of them, Apache eCharts, makes use of a locale string for formatting numbers, times, dates, ..., so I need a locale. My solution was, to put a default locale for every lang in the {lang}.conf file, as for instance in https://github.com/brewster76/fuzzy-archer/blob/master/skins/Bootstrap/lang/de.conf#L16. The user then may or may not override this value. I am not sure, if deprecating `lang` is an entirely good idea. It might be, I just haven't thought about it and probably don't know every issues that could go along with it. The context for setting the locale with this code is a single report. Does it make sense to decouple the translation mechanism (`lang`) from the `locale`, which specifies things like decimal and thousands separators and the way, date and time is formatted? Currently the settings are decoupled, but only one locale can be used for a single weewxd. So the question is: do you want to tie `lang` and `locale` together?
2. Regarding the tests, my focus was only on testing, how my approach is affecting the `lang`, and not to break anything of the existing behavior. To be honest, I don't have an idea, how to test the effects of setting the `locale` in a unit test, since it depends on having a certain locale available on the system the code is running. So, a test against a preformatted de_AT-localized number would either fail, because the locale isn't available, or wouldn't run, if it it is set to run only, if the locale is available.
3.  I'd expect: any `locale` setting found in the configs for the current report, will override any other locale settings, when running the report. I lack knowledge, if this approach is thread safe and could have side effects, if another task is running parallel depending on a locale setting.
4. Definitely :)

bellrichm:
I think that supporting something like (1) should be strongly considered. A bit of background, WeeWX-JAS does all localization (formatting and translation) work in javascript. It grabs the list of languages the user has configured for the browser. It goes through the list to find the first one that is ‘supported’ by WeeWX-JAS. Since currently there is no way to specify a region in WeeWX, it just takes the language code and uses that. So there is no way to support something like this.

| en-US                                                        | en-GB                                                              |    
|----------------------------------------|-------------------------------------------|
| On 01/31/2023 the sky was the color red. | On 31/01/2023 the sky was the colour red. |

Full disclosure,  I don’t really expect anyone to be translating WeeWX-JAS and I know that WeeWX-JAS is pushing the bounds of WeeWX support.

PR #900, opened 2023-11-28
Fix comment about default value for trans_days.

cincuranet:


gjr80:
I note this comment was corrected in the `V5` branch some two months ago. It might be easier to reject this PR to prevent a conflict when `V5` is eventually merged into `master`.

cincuranet:
Fine with me.

PR #902, opened 2023-12-06
draft: use pytest to run unit tests

matthewwall:
the unit tests have been run by invoking python directly on each test_xxx.py file

this is an attempt to disentangle the tests so that they can be run by pytest

that will give us a more robust testing system, html reporting of test results, trivial test coverage reports, and much easier CI/CD


tkeffer:
609 files changed? Gulp.

Could you start afresh, but do it off of `V5`? It shouldn't take many changes that way.

matthewwall:
the merge request should have been made relative to V5, not master.  re-doing it.

PR #903, opened 2023-12-07
use pytest to run unit tests

matthewwall:
the unit tests have been run by invoking python directly on each test_xxx.py file

this is an attempt to disentangle the tests so that they can be run by pytest

that will give us a more robust testing system, html reporting of test results, trivial test coverage reports, and much easier CI/CD


tkeffer:
I seem to recall refactoring the SQLite / MySQL classes once, much as you have (that is, have the "common" class inherit from `object` instead of `unittest.TestCase`). 

I can't remember exactly what the issue was, but I think it was that the functions `.setUp()` and `.tearDown()` don't get called.

PR #905, opened 2023-12-12
adding new option time_offset

olwaldi:
This is my first try to create a pull request towards weewx.

### DIFF
--- a/bin/weeplot/utilities.py
+++ b/bin/weeplot/utilities.py
@@ -189,10 +189,11 @@ def scale(data_min, data_max, prescale=(None, None, None), nsteps=10):
     return minscale, maxscale, interval
 
 
-def scaletime(tmin_ts, tmax_ts):
+def scaletime(tmin_ts, tmax_ts, offset=0):
     """Picks a time scaling suitable for a time plot.
     
     tmin_ts, tmax_ts: The time stamps in epoch time around which the times will be picked.
+    offset: Move the timeaxis by the given value to the right.
     
     Returns a scaling 3-tuple. First element is the start time, second the stop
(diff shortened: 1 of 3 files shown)


tkeffer (commented):
This looks good! Thanks!

Two things:

1. Do the Pull Request relative to `V5`, not `master`. The latter is very far behind.
2. Please address the comment below and use function `weeutil.nominal_spans()`

olwaldi:
Thanks for your review. I updated imagegenerator.py to use weeutil.nominal_spans instead of int (BTW, thus I also updated your original use of int within calling scaletime) and moved my pull towards V5. I tested this change within my current weewx 4.10.2

Are there any final actions necessary from my side within git? Keep in mind that this is my first use of git (I used it 8 years ago two or three times, but I remember a little bit on its concepts).

tkeffer:
I'm not quite sure what happened, but this still is not based on branch `V5`. For example, in both `imagegenerator.py` and `utilities.py` there are many V4.10 features in the files.

Version 5 also uses Markdown. The HTML file you provided is no longer used. You want `./docs_src/reference/skin-options/imagegenerator.md`. 

The basic problem is that the branch you based everything on, `time_offset`, is based on `master`, not `V5`. 

Try this:
```shell
git fetch
git switch V5
```

Make your changes to the resulting branch.

```shell
git push
```

Then go to Github and make a PR for the now modified V5 branch.

Don't get discouraged!

olwaldi:
Sadly switching the branch fails within my fork - V5 is unknown.

Probabely I should change the source branch of my fork. Sadly this seems to be not supported. Even trying to re-create a new fork fails because I own already the old one.

I've googled a lot without a real solution.

PR #906, opened 2023-12-14
adding new option time_offset

olwaldi:


tkeffer:
Thanks again.

Looking at this, it appears that the option doesn't shift the time axis but, rather, extends it to the right by amount `time_offset`. The origin of the time axis remains unchanged.

As such, it is interacting with option `time_length`.

I would have it _shift_ the time axis. If you want it longer, then you should modify option `time_length` as well.

olwaldi:
That's correct. My intention is to extend the original time axis by the given offset into the future. Thus as a consequence a total period of time  of roughly time_lenght + time_offset is drawn.
I do interprete time_lenght as being the time left from current time, while time_offset is the time right to current time. Maybe I should rename time_offset into e.g. time_length_extension?

Originally I simply added my time_offset onto time_lenght within imagegenerator.py calling scaletime, as you partly suggest. But then your algorithm to select nice left/right borders (e.g. in 3h steps) will fail. Thus I'm forced to add my extra option offset within scaletime to use offset only when necessary.

tkeffer:
I'm sorry, but as it stands, there are just too many things that make me uncomfortable about this PR.

I'm not keen on how `time_length` is no longer the final arbitrator of how long the plot will be.

It also doesn't feel right how the changes in `scaletime()` (lines 299-301) are necessary to undo what the algorithm is trying to do for plots a day long.

PR #907, opened 2023-12-24
Adapt `lowBattery.py` to a new script that uses Pushover (pushover.net).

bdruth:
* largely reuses `lowBattery.py`
* replace email/smtp code & config with complementary Pushover code & config
* tested locally

image


tkeffer:
First I've heard of Pushover. Looks useful.

But it is a bit too specialized to be included in the distribution. 

It would make a perfect extension. Package it up, then include a link from the home page of the wiki.

PR #908, opened 2023-12-29
Update csv-example.conf / delimiter fix

c0by85:
delimiter must also be provided with ' ', otherwise there will be an error message.

TypeError: "delimiter" must be a 1-character string


gjr80:
Thanks very much for the PR, we picked this one up a month or so back and it was fixed at commit [a72ac47](https://github.com/weewx/weewx/commit/a72ac47b54db044dd94f1544598227e9060ec150) to appear in v5.

PR #924, opened 2024-02-04
catch the CannotCalculate and 'say' the observation does not have an data

bellrichm:
I added an xtype to the Seasons observations_stats variable. The xtype raised a weewx.CannotCalculate exception. It was not caught, causing the template generation to fail.

I ‘fixed’ it with this code. Not sure if there is a better way…

### DIFF
--- a/src/weewx/tags.py
+++ b/src/weewx/tags.py
@@ -386,7 +386,10 @@ class ObservationBinder(object):
                                                   'not_null', db_manager)[0])
         else:
             # Nope. Try the xtypes system.
-            val = weewx.xtypes.has_data(self.obs_type, self.timespan, db_manager)
+            try:
+                val = weewx.xtypes.has_data(self.obs_type, self.timespan, db_manager)
+            except weewx.CannotCalculate:
+                return False
         return val
 
     def series(self, aggregate_type=None,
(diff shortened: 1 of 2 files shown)


tkeffer:
If an xtype extension cannot calculate something (exception `CannotCalculate`) that does not mean another one can't.

Try including `weewx.CannotCalculate` in with the other types in the `except` clause:

```
    for xtype in xtypes:
        try:
            # Try this function. It will raise an exception if it doesn't know about the type of
            # aggregation.
            vt = xtype.get_aggregate(obs_type, timespan, 'not_null', db_manager)
            # Check to see if we found a non-null value. Otherwise, keep going.
            if vt[0]:
                return True
        except (weewx.UnknownType, weewx.UnknownAggregation, weewx.CannotCalculate):
            pass
```

Experiment a bit and see if that helps. 

Also, a regression test would be useful.

bellrichm:
I tried that, it didn’t work.  I’ll look to catch it up higher in the stack.

tkeffer:
Why didn't it work?

bellrichm:
> Why didn't it work?

The exception bubbled up to the template generation, causing it to fail

bellrichm:
> Also, a regression test would be useful.

If we find the appropriate place, I can work on a test, might take a while.  I’ll need some guidance, last time I tried and got lost in the weeds. I gave up when it seemed like I needed to instal MySQL.

bellrichm:
Yeah, that latest won’t work… because no other xtypes are tried
Back to the drawing board

bellrichm:
I’ll close this and dig into my xtype more.

PR #927, opened 2024-02-05
Catch the CannotCalculate exception, giving other xtypes a chance to perform the calculation

bellrichm:
As noted in this [pull request](https://github.com/weewx/weewx/pull/924#issuecomment-1925893127), this is the appropriate place. Now the `CannotCalculate` raised by `XtypeTable` is ‘ignored’ and my xtype gets a chance.
The reason I thought it was not working; because more of the `Seasons` can be processed, another `CannotCalculate` is now raised. Since they were both raised from the same line of code, I overlooked in the stack trace how that code was reached.

### DIFF
--- a/src/weewx/xtypes.py
+++ b/src/weewx/xtypes.py
@@ -157,7 +157,7 @@ def has_data(obs_type, timespan, db_manager):
             # Check to see if we found a non-null value. Otherwise, keep going.
             if vt[0]:
                 return True
-        except (weewx.UnknownType, weewx.UnknownAggregation):
+        except (weewx.UnknownType, weewx.UnknownAggregation, weewx.CannotCalculate):
             pass
     # Tried all the  get_aggregates() and didn't find a non-null value. Either it doesn't exist,
     # or doesn't have any data


tkeffer:
Rich, I think I may have led you astray. I didn't spend enough time studying what you were doing. Your [original commit](https://github.com/weewx/weewx/pull/924/commits/f381cd8e52e7e24e301fa37328346e1aa85aa449) is probably what we want.

The exception `weewx.CannotCalculate` should be raised when a type is known to the function, but all the information necessary to calculate it is not there. It's a definitive answer. So, for `has_data()`, it should just return `False`.

Sorry to make you do this, but I think your PR is important. Could we go back to the original PR?

bellrichm:
No problem, its a subtle difference but I think I follow. The other exceptions are ‘saying’ I don’t know about this, let’s see if another xtype does. The CannotCaluculate is ‘saying’, I know and ‘control’ this type and it ‘does not have a value’.

I probably should have started a development discussion, but I thought seeing the code would be clearer than my explanation….

Note, around line 136 is similar. I was going to create a separate pull request because ‘passing’ had some unwanted side effects. I think returning `False` will be clean. I’ll play around and see.

bellrichm:
I’ll see if I an add anything to the wiki.

PR #930, opened 2024-02-10
implement `weectl import --update` action

gjr80:
refer [PR839](https://github.com/weewx/weewx/pull/839).

PR839 sought to add the ability for `wee_import` to update existing database records with imported records whose timestamp already exists in the WeeWX archive. Current `wee_import`/`weectl import` behaviour is to ignore imported records whose timestamp already exists in the WeeWX archive. PR839 was raised against the WeeWX v4 code base, WeeWX v5 has since been released with `wee_import` being replaced with `weectl import`. Given the significant changes to the WeeWX code base in developing v5, PR839 will be discarded and the changes sought at PR839 will be applied via a new PR or via separate direct commits to the v5 code base.

This PR has been raised to clearly identify the changes to the v5 code base required to implement the intent of PR839. This PR includes the PR839 changes (to `import_cmd.py`, `weeimport.py` and `manager.py`) as well as a number of other necessary changes (to `csvimport.py`,`cumulusimport.py `,`wdimport.py `,`weathercatimport.py ` and `wuimport.py `).

This PR does not include documentation changes, documentation will be updated once the code is finalised.

### DIFF
--- a/src/weectllib/import_cmd.py
+++ b/src/weectllib/import_cmd.py
@@ -14,6 +14,7 @@ import_usage = f"""{bcolors.BOLD}weectl import --help
        weectl import --import-config=IMPORT_CONFIG_FILE
                      [--config=CONFIG_FILE]
                      [[--date=YYYY-mm-dd] | [[--from=YYYY-mm-dd[THH:MM]] [--to=YYYY-mm-dd[THH:MM]]]]
+                     [--update]
                      [--dry-run][--verbose]
                      [--no-prompt][--suppress-warnings]{bcolors.ENDC}
 """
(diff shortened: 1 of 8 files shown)


tkeffer:
I assume you've tested this.

As for `update`, I would prefer it was added to the end of the parameter list for `addRecord()` and `_addSingleRecord()`. It's entirely possible that some extension, somewhere, calls one of them with positional arguments.

gjr80:
I tested a small csv import into SQLite under 5.0.1. The test consisted of both source data that was complete (ie every import field was populated) and incomplete (some import fields were blank. Also tested both with and without `--update`. In all cases it did what it should. Whilst each import source will be capable of using `--update` I did not test each source as `--update` affects database operations sufficiently up the 'API tree' that the import source is irrelevant (though each '`weectl import` source' file needed a small change due to each source providing advice to the user on the options selected/import detail).

If you think there is a need for more testing let me know.

Signatures for affected `manager.py` methods have been altered as requested.

If you are happy with the code don't accept the PR, I will touch up the docs etc and do it all in one go later.

tkeffer:
Looks good!

PR #942, opened 2024-03-01
Pass the record retrieved from the DB into the XTypes system

bellrichm:
If necessary, the record is retrieved from the DB. But this is not the record passed into the XTypes system


tkeffer:
Funny how this little detail remained dormant for over 4 years!

Issue #962, opened 2024-06-07
shim deprecated wee_ commands for a while


vinceskahan:
It is remarkable how many "where did wee_xyz go" posts we see.  The move of everything to weectl is great, but users are clearly reading ancient notes and howtos and 'not' reading the upgrade and utilities guides that were so painstakingly written...

Suggest that the old wee_ commands be shimmed with something obvious for at least a while.  A bash example might look like something like the following....

```
#!/bin/sh
echo ""
echo "###################################################################"
echo "#                                                                 #"
echo "# ERROR - command replaced with 'weectl' as of weeWX version 5.0  #"
echo "#         please see the Upgrade and Utilities Guides for details #"
echo "#                                                                 #"
echo "###################################################################"
echo ""
exit 1
```

Suggest the same message be returned for the wee_config, wee_database, wee_debug, wee_device, wee_extension, wee_import, and wee_reports commands.

Unfortunately I cannot follow how exactly your existing weectl command is created through all the makefile and pip twisty little passages, so I'm at a loss regarding how to make this happen in terms of a PR, but basically suggest you add (add back) the old commands with a big honkin' unambigious message for the folks who simply will not RTFM.

duane-nepa:
That's a great idea.

So you have an idea how I got there. I was trying to add an extension. The git for the extension said to use wee_extension for the install. This didn't work. I google searched for wee_extension and I got to this guide. https://www.weewx.com/docs/4.10/utilities.htm

Even if there was a note on the top of that guide that said it was deprecated that would have been the clue I needed to look elsewhere. Maybe just the CSS could be edited. 

I don't often need to do anything to my setup. I was switching home automation software to Home Assistant and trying to get station data connected to that. And thought I was going to use MQTT.

But instead, I found an easy way to scrape the rss.xml file.

But now that I've seen the ver. 5 docs I'll know if I'm in the wrong place in the future.

Hope this helps.

roe-dl:
May be, it could be an idea, if someone checks all the commonly used extensions and creates an issue with them, if the documentation does not reflect the changed tools.

tkeffer:
I'm not sure how to package a shell script using Poetry, and then install it using `pip`. If there's a way to do that, I haven't found it.

It would have to be a bunch of small Python programs.

vinceskahan:
Python would be equally fine.  The bash snippet was just an example wording that exits non-zero for discussion purposes.

[...]

PR #964, opened 2024-06-19
add python shims for deprecated wee_commands for Issue #962

vinceskahan:
Here's a PR for starting to add those wee_xyz shims.

- I 'have' tested that poetry does the right thing and the shims seem to work ok for me (macos)
- I added edits to add them to the dpkg and rpm packages but have 'not' tested those additions

I didn't change any changelogs or software versions.  This PR is just the meat, so to speak.

Let me know if you want me to try to build/test the packages......

### DIFF
--- a/bin/wee_config
+++ b/bin/wee_config
@@ -1,0 +1,9 @@
+#!/bin/sh
+app=wee_config.py
+
+# Get the weewx location and interpreter.  Default to something sane, but
+# look for overrides from the system defaults.
+WEEWX_BINDIR=$(dirname "$0")/../src
+WEEWX_PYTHON=python3
+[ -r /etc/default/weewx ] && . /etc/default/weewx
+exec "$WEEWX_PYTHON" $WEEWX_PYTHON_ARGS "$WEEWX_BINDIR/$app" "$@"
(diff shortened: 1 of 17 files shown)


tkeffer:
Unfortunately, this won't work for pip installs. For pip executables, you specify an entry point in the `pyproject.toml` file, then pip manufacturers a shim that calls an appropriate Python program. Both `weewxd` and `weectl` work this way. You can not just include an executable.

I don't see an easy way to make a single version that works for both the package installers and pip.

Frankly, I don't this added complexity is worth it, but maybe you can figure something simple out.

vinceskahan:
Take another look at the diff. I added python scripts for each legacy
command, added them as entry points, and poetry did exactly what you said.
I also tested that the resulting shims worked.

-----  ***@***.*** ----

tkeffer:
My apologies. I missed that completely. Moving too fast, too early in the morning!

vinceskahan:
Minor additions to get the .deb package to generate the wrapped shims too.  Debian package looks ok for clean install.

vinceskahan:
Tom at this point I *think* this one is good to go forward on.  Tests ok for pip, deb, rpm installations.   I did not test upgrades as I trust Matthew's upgrade wizardry logic.

If 5.next is imminent I'd suggest you slip this one in if possible.

tkeffer:
This problem seems to have gone away as people have adapted to `weectl`.

Issue #981, opened 2024-11-17
Pyephem is deprecated


roe-dl:
*PyEphem* is deprecated as described on the [*PyEphem* website](https://rhodesmill.org/pyephem/). The author replaced it by the [*Skyfield* module](https://rhodesmill.org/skyfield/).

I looked into the source code of Pyephem's underlying C library and found out that it contains a hard coded table of base values that ends in the year of 2018. After that the values are extrapolated. So there is an accuracy issue there. In contrary the *Skyfield* module downloads that table from the NASA website to always get an actual version of it.

There is also another Python astronomy project called *Astropy*. A first glance I did not see which one is better.

tkeffer:
This is something I looked into a few years ago, and reached a similar conclusion. We will need to move to something else.

At the time, I didn't pursue Skyfield because it depends on NumPy, which would add a lot of overhead to WeeWX.

First I've heard of _Astropy_. Looks like it would also requires NumPy.

Because the extended almanac is an optional feature, maybe it's not so bad if it pulls in NumPy.

roe-dl:
* [Skyfield Elegant Astronomy for Python](https://rhodesmill.org/skyfield/)
* [The Astropy Project](https://www.astropy.org)

As you say, **Skyfield** depends on **NumPy**, but **AstroPy** is not better. It reads

> The first step to installing astropy is to ensure that you have a Python environment which is **isolated** from your system Python installation. This is important because astropy has many dependencies, and you do not want to accidentally break your system by installing incompatible versions of these dependencies.

According to the documentation, **AstroPy** depends on **NumPy**, too, and additionally on [PyERFA](http://pyerfa.readthedocs.org/), [PyYAML](https://pyyaml.org/), and [packaging](https://packaging.pypa.io/). And there are a lot more optional dependencies.

roe-dl:
I guess I will try to do some coding. I would prefer Skyfield, as it has one dependency only, compared to AstroPy which has many.

I would check for the presence of Skyfield and NumPy by trying to import them. If they are there Skyfield is used. If not, the program falls back to the current behavior which is to check for PyEphem etc.

Unlike PyEphem, Skyfield downloads actual timescale and ephemerides data from IERS and USNO. So if they are updated, Skyfield can use them. On the other hand, Skyfield needs a directory to store files. I would use `SQLITE_ROOT` for that, as it points to a directory of continuously updated data.

[...]

PR #988, opened 2025-01-04
Make almanac extensible

roe-dl:
See issue "[Pyephem is deprecated](https://github.com/weewx/weewx/issues/981)" for background.

The approach is very similar to the XType extensions. There is a list of almanacs that are tried in order until one of them provides a result for the given attribute. An almanac is a subclass of `weewx.almanac.AlmanacType`. 

Built-in almanacs of core WeeWX are the weeutil formulae and the PyEphem wrapper. PyEphem is used automatically if it is available. Otherwise the weeutil formulae are used. This behavior is as before. Calls to PyEphem or weeutil formulae in `__getattr__` I moved to separate classes that seamlessly fit into the new concept of an extensible almanacs list. So no special condition statements were needed.

If a developer wants to provide an almanac extension they have to define a WeeWX service that instantiates a subclass of `weewx.almanac.AlmanacType` and inserts it into the almanacs list `weewx.almanac.almanacs`, just like they would do with XType extensions.

I ran `test_almanac.py`. The result was OK. I checked the output of the Seasons skin, both with and without PyEphem, and the results were reasonable to me. Additionally I checked to output of `$almanac.sunrise`, `$almanac.sunset`, `$almanac.moon_fullness`, and `$almanac.separation()` again with and without PyEphem.

Whether `AlmanacType` should include a precalculation method, could be discussed. It would make a difference if the user  re-uses `$almanac` (for example by `#set $alm = $almanac` and then `$alm.sun.rise`, `$alm.sun.set` etc.).

I added documentation to the file in the hope it will make it easier to read.

### DIFF
--- a/docs_src/custom/cheetah-generator.md
+++ b/docs_src/custom/cheetah-generator.md
@@ -1629,7 +1629,8 @@ for each context. For example, the Thai language file would include:
 
 ## Almanac
 
-If module [`ephem`](https://rhodesmill.org/pyephem) has been installed, then
+If module [`ephem`](https://rhodesmill.org/pyephem) or an appropriate
+almanac extension has been installed, then
 WeeWX can generate extensive almanac information for the Sun, Moon, Venus,
 Mars, Jupiter, and other heavenly bodies, including their rise, transit and
 set times, as well as their azimuth and altitude. Other information is also
@@ -1651,7 +1652,7 @@ Current time is $current.dateTime
 #end if
 ```
 
-If `ephem` is installed this would result in:
+If an almanac module like `ephem` is installed this would result in:
 
 <div class="example_output">
 Current time is 03-Sep-2010 11:00
@@ -1983,7 +1984,12 @@ The separation between Venus and Mars is 55:55:31.8
 ### Adding new bodies
 
 It is possible to extend the WeeWX almanac, adding new bodies that it was not
-previously aware of. For example, say we wanted to add
+previously aware of. The following description explains how to do it for 
+PyEphem, which is supported by core WeeWX. If you use an alternative almanac 
+extension, refer to the documentation of that extension for how to add new 
+bodies.
+
+For example, say we wanted to add
 [*433 Eros*](https://en.wikipedia.org/wiki/433_Eros), the first asteroid
 visited by a spacecraft. Here is the process:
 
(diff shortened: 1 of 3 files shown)


tkeffer:
Sorry for taking so long to get back to you. I think you're on the right track.

Question: by making the `weetutil` Sun and Moon routines just another `AlmanacType`, do you think we still need `_precalc()`? Things like `moon_phase` could be supplied at evaluation time by `WeeutilAlmanacType`.

roe-dl:
First I wondered why there is a `_precalc()` method at all. Then I came to the conclusion that it may come in useful if you re-use the almanac instance by assigning it to a variable like `#set $alm = $almanac`. If you do so, you can use those pre-calculated values for all the almanac data you want to show. I don't know whether this was the original intension for the `_precalc()` method, but that was the reason I let it there.

And I wondered if I should include the pre-calculation method into the API in order to make it possible for other almanacs to do some pre-calculation, too.

tkeffer:
Generally, stateless designs are a better idea. If I were to do it again, I'd find a way around using something like `_precalc()`. 

Better to wait until evaluation time to decide which algorithm to use. All those conditionals based on `hasExtras` are a bad idea!

roe-dl:
I guess, sometimes you may decide against your principles due to speed reasons. But no problem, I will try to remove `_precalc()` and have a look at `hasExtras`.

roe-dl:
@tkeffer Do we need `_moon_fullness` (with underscore at the beginning)? It is included in the test.

tkeffer:
I don't think we need `_moon_fullness`.

Of the two values you give, I have no idea which is correct.

roe-dl:
> Of the two values you give, I have no idea which is correct.

`test_almanac.py` says:
* for the `weeutil formula`:
  ```Python
  self.assertAlmostEqual(self.almanac.moon_fullness, 3, 2)
  ```
* for the PyEphem module:
  ```Python
  self.assertAlmostEqual(self.almanac.moon.moon_fullness, 1.70, 2)
  ```

You could look at [mooncalc.org](https://www.mooncalc.org/#/46,-122,10/2009.03.27/12:00/1/3), which says 1.7% (which is rounded 2%).

tkeffer:
I think you're on the right track. I've created a new branch [`roe-dl-almanac_extensions`](https://github.com/weewx/weewx/tree/roe-dl-almanac_extensions) that is actually a branch off of `development`, instead of `master`. The branch `master` should be limited to bug fixes, while `development` is for new features.

Could you continue your work off of the new branch? Once you're got a proof-of-concept using SkyField (or another, more modern, almanac), we can merge it into `development`. 

Thanks, Karen.

roe-dl:
If you really mean "proof-of-concept" and do not expect all the attributes already working, then I can present an early version of [weewx-skyfield-almanac](https://github.com/roe-dl/weewx-skyfield-almanac). A lot of things already work like seasons, moonphases, sunrise, sunset, moonrise and moonset. And a lot of attributes of heavenly bodies are still missing. And of course, I still have to check the results. 

And there is a little [example](https://github.com/roe-dl/weewx-easter-almanac-example) that shows the general structure of an almanac extension.

tkeffer:
Those sound like perfectly sensible names. 

Which brings me to another topic: documentation. The official WeeWX documentation does not need to document Skyfield-specific extensions, but it should say something about Skyfield being available.

I'm hoping that you're willing to act as the author and maintainer of the extension. If you wish, we could host its repository in the [Github area for WeeWX](https://github.com/weewx), much like the WMR200 driver. Or, can continue to host it in your area.

roe-dl:
> Which brings me to another topic: documentation. 

I added some documentation in the CheetahGenerator description regarding almanac extensions. I wrote, that there is no need to move away from PyEphem unless problems suggest to do so. For Skyfield I mentioned pros and cons.

> I'm hoping that you're willing to act as the author and maintainer of the extension. If you wish, we could host its repository in the [Github area for WeeWX](https://github.com/weewx), much like the WMR200 driver.

Yes, we can do so if you want.

I realized that you referenced me by Johanna this time while you did so in the past by Karen, which is my nickname at Google (and indeed my middle name). I do not know whether this is important to you. To me it is not.

tkeffer:
> I realized that you referenced me by Johanna this time while you did so in the past by Karen, which is my nickname at Google (and indeed my middle name). I do not know whether this is important to you. To me it is not.

I'd always thought your name was Karen, but then I saw Johanna in the copyright and thought I'd gotten it wrong. I'm relieved to hear that I did not!

tkeffer:
I've merged the branch into `development`.

Thanks, Karen, for another valuable contribution!

PR #990, opened 2025-01-16
Ensure QC does not crash with bad data

mdornseif:
On my installation I was experiencing crashes - propably due to the schemaless nature of sqlite.

### DIFF
--- a/src/weewx/qc.py
+++ b/src/weewx/qc.py
@@ -54,19 +54,28 @@ class QC:
 
         for obs_type in self.mm_dict:
             if obs_type in data_dict and data_dict[obs_type] is not None:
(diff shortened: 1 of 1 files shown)


tkeffer:
Thanks for your PR. Two comments:

1. I understand your motivation here, but your PR would allow whatever caused the exception to be put into the database, where it could cause even more mischief. Better to set the value to `None`.
2. Could you give us some history on what is causing the problem? Is it a value that is presented as a string? Something else? It would be good to fix the problem at its source, instead of catching it downstream.

Thanks again.

tkeffer:
@mdornseif Anything new on this?

tkeffer:
Inactive.

Issue #991, opened 2025-01-29
inconsistency between wview_extended.py and units.py on luminosity vs illuminance


toast14:
In src/schemas/wview_extended.py, a schema field is defined for "luminosity".  In src/weewx/units.py, the value "illuminance" is in "group_illuminance".  Ideally, these two should match so the value is properly enumerated with units.

tkeffer:
Thank you for bringing this up! There have been several email and issue discussions about this in the past, going back to issue #685 and beyond. Unfortunately, none of them have been driven to conclusion.

First, note that despite their similar names, luminosity and illuminance are not the same. 

__Luminosity__ is the amount of _perceived_ light emitted by a source, typically measured in _lumens_. It is distinct from radiative power (measured in watts) in that it is weighted by the sensitivity of the human eye. For example, visible light is weighted more than UV or infrared.

__Illuminance__ is the amount of perceived light that falls on a surface area, typically measured in `lux`, or 1 lumen / m&sup2;.

In WeeWX, we have a unit for `illuminance` (unit `lux`), but no unit for `luminosity`.

I see that you have submitted PR #992 as an attempt to fix this. Unfortunately, I think it would make things worse because a unit intended for illuminance, would be used for luminosity.

Three things need to be done to set this all right:

1. We need to add a column `illuminance` to the `wview_extended` schema.
2. We need to add a new unit group `group_luminosity` to the set of unit groups in `units.py`.
3. The default unit of measure for `group_luminosity` should be set to `lumen`. 

The downside of this approach is that several extensions have been using `luminosity`, when they really should have been using `illuminance` (which didn't exist). It's unclear to me whether these extensions have been assuming that type `luminosity` is measured in `lux`. If so, these changes won't break anything, but extension authors should really fix things to use the right unit.

If you want, you could submit a PR that does these 3 things. Or modify the one you did submit.

Thanks again!

toast14:
I have modified the code as above (and on the pull request).

gdt:
As I said in the PR, I suspect all uses of luminosity are likely erroneous.   Thus I think the best course is to add illuminance and lux, and not to change luminosity right now.   Then, to address each use of luminosity, decide if it is correct (my guess is 99% not correct) and fix it.

I suspect we are where we are because various weather station manuals use terms incorrectly, and they've been adopted.

[...]

PR #992, opened 2025-01-30
#991:  change illuminance in units.py to luminosity to match schema

toast14:
Changed illuminance in units.py to luminosity to match provided extended schema.


tkeffer:
Sorry. I forgot another step. Because the unit `lumen` does not exist yet in WeeWX, formatting and labeling information also has to be added to `defaults.py`. Follow the pattern set by `lux`.

toast14:
Lumen units added.

gdt:
I would like to see the PR title updated and the commits rebased to include only what belongs here; bringing in commits about ubuntu and sqlite is confusing (even if they are on the branch this would be merged into).  It's also confusing to have things changed and changed back.   (I realize Tom might squash-merge, but were I maintainer I would want clean PRs to review.)

Also, as someone who has read about photometry, i find the use of the word luminosity very confusing for two reasons:
   - As I read wikipedia, it's a radiometric unit (watts), not a photometric unit (lumens)
   - It measures total light output from a body, and thus has nothing to do with weather

I agree we should clean up the schema so we have a column for illuminance, with units lux (= lm/m^2).    And then driver maintainers can clean up drivers.

I see the existence of luminosity in wview schema as an error.   I don't understand what it's supposed to mean, and I suspect various things are put in that column.     My approach would be to leave it be, and have this PR simply add illuminance.   Then, ever use of luminosity can be investigated, starting with a presumption that it is erroneous, and evaluated and addressed.

Because of that, I think we should not be adding "lumens" as a unit, at least at this time.

toast14:
I looked through some more of the documentation and code this morning.  According to the documentation, the FineOffset driver uses luminosity instead of illuminance; however, when I go to the code, it's using illuminance.  Also, I missed the localization within in the Skins directory for illuminance.

I can cancel this pull request and just rework everything to just illuminance and remove the errant uses of luminosity.

gdt:
Great, thanks.    Happy to review, as the unofficial chief photometry ranter for weewx :-)    (My own station measures irradiance, and not any kind of photometric unit.)

tkeffer:
Compelling arguments, Greg. 

I've trimmed down the PR to just add `illuminance` to the schema.

Issue #997, opened 2025-02-22
Solar Time


roe-dl:
@tkeffer I have a question.

The daily course of the weather is heavily influenced by the position of the sun in the sky, which is expressed by solar time. Clock time can be far away from that because of time zones and daylight saving switching. I therefore find it interesting to also display the solar time. 

Solar time can be calculated by the almanac, but actually it cannot be displayed in an easy way, because it is already local time and therefore cannot be converted by the `time.localtime()` statement in the formatter. To add support for displaying the solar time or other local time I would like to add some little 10 lines to the formatter in `units.py`. 

Would you consider accepting a PR for that?

tkeffer:
Sounds useful, but I'm not sure what you have in mind. Tags such as `$almanac.sun.rise.solar`?

roe-dl:
I thought about something like `$almanac.solar_datetime.format("%H:%M:%S")`. Your example sounds interesting, too.

roe-dl:
As this is continued in the PR #998, I guess, this issue can be closed now.

PR #998, opened 2025-03-04
almanac related units, almanac locale

roe-dl:
* When the almanac will be extensible, there are more language dependent words than just moon phases and compass directions. So pass the complete `[Almanac]` section from the language definition file to the almanac for the use in extensions. An example for those language dependent words not available so far are the names of the planets.
* There are 2 undocumented but used PyEphem almanac attributes that return a value in Astronomical Units. Therefore I added that unit to the conversion dictionary.
* To display solar date and time a small extension to the `ValueHelper` class is necessary as described in #997.

### DIFF
--- a/src/weewx/tests/test_almanac.py
+++ b/src/weewx/tests/test_almanac.py
@@ -172,7 +172,7 @@ class AlmanacTest(unittest.TestCase):
     @unittest.skipIf(not pyephem_installed, "Skipping test_exceptions_pyephem: no pyephem")
     def test_exceptions_pyephem(self):
         # Try a nonsense body
-        with self.assertRaises(KeyError):
+        with self.assertRaises(AttributeError):
             self.almanac.bar.rise
         # Try a nonsense tag
         with self.assertRaises(AttributeError):


tkeffer:
Not quite understanding the purpose of `local_djd`. A few questions:

1. What would this be used for? Solar time? Something else? 
2. How does it differ from the existing `dublin_jd`?
3. Why make it a new unit group, rather than a new unit type? Isn't it just another way of measuring time?

roe-dl:
Unknown heavenly bodies like `$almanac.bar.rise` raised an exception that resulted in a complaint to the logs and no page created instead of writing the name of the tag into the created page, as it happens elsewhere. Assuming that was not intentional I changed the exception to `AttributeError` which results in the common behavior.

tkeffer:
My apologies for not having the time to review this. I'm in the middle of a 3 month marine voyage. Hopefully, I'll have some time later.

roe-dl:
No problem. I used the time to further work on the project.

roe-dl:
Passing the complete `[Almanac]` section to the almanac also makes it possible to get localized constellation names. The the Skyfield almanac extension comes with localized names for some common languages, and the installer merges them into all language definition files it can find. Then a tag like `$almanac.sun.constellation` would return the actual constellation the Sun is in. The answer is astronomical, not astrological. For the stars you also get the constellation they belong to.

src/weewx/almanac.py
```diff
@@ -398,6 +400,10 @@ def get_almanac_data(self, almanac_obj, attr):
                                                context = 'ephem_day',
                                                formatter=almanac_obj.formatter,
                                                converter=almanac_obj.converter)
+        elif attr.startswith('previous_') or attr.startswith('next_'):
+            # Prevent availability checks from being misinterpreted as
+            # heavenly bodies.
+            raise AttributeError(attr)
```
tkeffer, line 406: The `elif` farther up the file (see line 371 or thereabouts) should have caught this. It checks for specific names (such as `previous_equinox`). If that's not good enough, it could do your check `attr.startswith('previous_')...` etc.

tkeffer:
I made a few inline comments.

Where will `local_djd` come from?

src/weewx/almanac.py
roe-dl: This is to prevent tags like `$almanac.previous_aphelion` from being interpreted as a heavenly body's name. The PyEphem almanac does not cover that event, but almanac extensions may add more `previous_...` and `next_...` events. If skin authors use those additional events without checking for the availability of the respective extension this code produces a reasonable error message.

Issue #1013, opened 2025-09-12
Feature request: "WOW-BE reboot" as StdRESTful service


mrtnrey:
UK Met Office has recently announced that they intend to stop managing the central data hub of WOW in the course of 2026. The Royal Meteorological Institute of Belgium has decided to re-launch the platform under the name "WOW-BE reboot" with a new back-end. See https://wow.meteo.be/en/ for more background.

The upload protocol, however, stays the same as WOW-UK. Only the endpoint changes from `http://wow.metoffice.gov.uk/automaticreading?` to `http://wow.meteo.be/api/v2/send?`. Station owners have to make a new account on https://wow.meteo.be/web/login and re-register their station (they receive a new UUID).

Documentation on the send protocol is here https://wow.meteo.be/docs/api/

It would be nice to have this new platform integrated into WeeWX as an extra service in StdRESTful. BTW: we intend to open up all received data as open data. The back-end itself is open source: https://github.com/rmibelgium/wowbe under GPL.

lizdodd:
The WOW-BE service can accept solar radiation however Weewx does not yet send this. This would be a useful addition to the Weewx code.

tkeffer:
Thanks, everyone, for tracking this.

Could someone put together a PR on the `master` branch?

lizdodd:
I have a question before working on this.
Do you want WOW and WOW-BE available in parallel or change to WOW-BE?

evilbunny2008:
To do both do I just copy class StdWOW(StdRESTful): and rename it something like class StdWOWBE(StdRESTful): and leave protocol set at "WOW" ?

evilbunny2008:
Copying the class section and modifying it slightly worked, but I have no idea what changes are needed to send different barometer readings to the different WOW's as per @lizdodd comment above.

Looks like WOW-BE accept 2 different barometer readings:

 "absbaromin": 0,
  "baromin": 0,

Looks like they also accept "dewptf" which isn't listed in the restx.py file

The next question is a slight variation on the above one, how to also send solar information?

"solarradiation"

https://wow.meteo.be/docs/api/#/operations/send.wow

glennmckechnie:
To use the variables changes supplied by @lizdodd, I changed / added them to the (new) class WOWBEThread entry as per the following diff.

absbaromin is now working correctly - for me
solarradiation isn't, but I'm sending a very high value so it is probably being dropped.
dewptf is there already and is being uploaded to the daily observations table, for the 3 of us.

[...]

PR #1014, opened 2025-09-16
Updated restx.py to send data to WOW-BE

evilbunny2008:
Credit for the code changes goes to @glennmckechnie, I simply adapted what he posted to issue #1013 

This pull request includes changes to restx.py to enable uploading to WOW-BE.

Other changes I made to restx.py make it easier to inherit common code from existing classes.

I can almost guarantee that my code will not be 100% pythonic and welcome feedback to improve it to achieve the same outcome.

I also removed a fair bit of white space on empty lines that showed up in nano.

Code changes were tested in weeWX 5.1.0 using the Debian package install method, but I don't think the changes will negatively impact other install methods.

### DIFF
--- a/src/weewx/restx.py
+++ b/src/weewx/restx.py
@@ -1109,7 +1109,7 @@ class StdWOW(StdRESTful):
     """Upload using the UK Met Office's WOW protocol.
 
     For details of the WOW upload protocol, see
-    http://wow.metoffice.gov.uk/support/dataformats#dataFileUpload
+    https://wow.metoffice.gov.uk/support/dataformats#dataFileUpload
     """
 
     # The URL used by WOW:
--- a/src/weewx/restx.py
+++ b/src/weewx/restx.py
@@ -1150,7 +1150,7 @@ class StdWOWBE(StdWOW):
     """
 
     # The URL used by WOW:
-    archive_url = "https://wow.meteo.be/docs/api/#/operations/send.wow"
+    archive_url = "http://wow.meteo.be/api/v2/send"
     protocol_name = "WOWBE"
     Thread = WOWBEThread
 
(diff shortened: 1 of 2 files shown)


evilbunny2008:
@glennmckechnie Good catch, I just updated the pull request to have the sample weewx.conf with a WOWBE section and the class listed in restful_services line.

Thanks again for your help.

tkeffer:
Thank you for the PR! I appreciate the time that it took to put it together.

A few suggestions:

1. A request: don't introduce formatting changes that are not essential to the PR. They make it hard to see what has changed. Or, do them in a separate commit. See [code style](https://github.com/weewx/weewx/blob/master/docs_src/devnotes.md#code-style).
2. Along the same line, I did notice that the lines (starting around line 551)
    ```python
          if weewx.debug >= 2:
            log.debug("%s url: '%s'", self.protocol_name, request.get_full_url())
    ```
    seem to have disappeared.
3. In `StdWOW.__init__` there is a test 

                if self.protocol_name == StdWOW.protocol_name:
    This is presumably so that the contained code is not run when the subclass `StdWOWBE` is initialized. This is not terribly Pythonic. Another approach is to use class instances for `StdWOW` and `StdWOWBE` for the class-specific things. For example:
    ```python
    class StdWOW:
        Thread = WOWThread
        def __init__(self):
            ...
            self.archive_thread = self.Thread(self.archive_queue, _manager_dict,
                                              protocol_name=self.protocol_name,
                                              **_ambient_dict)
            

    class StdWOWBE(StdWOW):
        Thread = WOWBEThread

    ```
    You actually did something similar with `protocol_name` --- I'm just extending it to classes.

    If you prefer, this could also be done using property methods. Either way, once you're done, you should not even need an initializer for `StdWOWBE`.

tkeffer:
I made some changes to what you did, then merged them into `development` (not `master`). I also added documentation and a change list entry. Commit https://github.com/weewx/weewx/commit/7715977b429dfac78d601750baf46442adfb11da

Take a close look at the documentation part. I'm not using WOW-BE, so I had to guess about how to do the registration. Make sure I did it right. 

Please take it for a test spin!

PR #1015, opened 2025-09-16
Support For Gill MetConnect

Cosmospacedog:
MetConnect-Social_-272591656

I recently set up this custom driver for receiving and processing data from [Gill Metconnect](https://gillinstruments.com/compare-weather-stations/metconnect-weather-station/) systems and thought it was worth contributing back to weewx.

Gill instruments provide weather stations capable of providing diverse weather and gps data, this driver covers their MetConnect line - having been tested with the MX400,MX500,MX550 and MX600 models.

These devices are set up ahead of transmit via the [Gill MetSet](https://gillinstruments.com/downloads/metset-download-form/) software and cannot have their configs changed on the fly. Thus, the driver makes use of the .mcf config files provided on setup to infer the correct packet data format.

No additional dependencies are required to implement the driver as metConnect systems transmit standard UTF-8 serial data.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -18,6 +18,10 @@ Fix problem that prevented cached values of first and last timestamp from
 being set. [PR #999](https://github.com/weewx/weewx/pull/999). Thanks to user
 Rich Bell!
 
+Fix problem in `Standard` skin if `Select Month` was literally selected in drop
+down list. [PR #1002](https://github.com/weewx/weewx/pull/1002). Thanks to user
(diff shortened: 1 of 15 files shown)


evilbunny2008:
While I can't speak authoritatively, it's my guess that this will probably be rejected and suggested that you distribute it as a driver extension package from your own GitHub releases section.

This is the standard for the majority of hardware drivers for weeWX, then just post a link to this [wiki page](https://github.com/weewx/weewx/wiki).

PR #1016, opened 2025-09-18
Tweaked WOW-BE information in stdrestful.md

evilbunny2008:
Tweaked WOW-BE information in stdrestful.md, it slightly differs from WOW in that you can use a password instead of a 6 digit PIN.

### DIFF
--- a/docs_src/reference/weewx-options/stdrestful.md
+++ b/docs_src/reference/weewx-options/stdrestful.md
@@ -296,10 +296,13 @@ every archive record to be posted. Optional. Default is zero.
 
 ## [[WOW-BE]] 
 
-WeeWX can send your current data to [WOW-BE](https://wow.meteo.be), a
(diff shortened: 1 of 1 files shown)


tkeffer:
Thanks. Did you get a chance to try the `development` branch?

evilbunny2008:
I'll build it into a deb package and try it over the weekend then.

I was just checking to see if there was anything major like the 5.x versions that no longer run as root and caused permission issue.

evilbunny2008:
Regarding the schema change, did I get it right thinking it's just a python change and not a database change?

I rarely build packages, but I have in the past to back port programs in Debian testing so I avoid getting a broken system by mixing and matching stable and testing system packages, well old stable at the moment as I haven't taken the plunge to Trixie yet.

I'm playing it safe and letting others do the beta testing. *grin*

PR #1017, opened 2025-09-19
Updated debian/rules file so package would build successfully

evilbunny2008:
## Preface...

I'll preface this by saying I'm no expert when it comes to producing .deb files and I have no idea how @matthewwall is able to build WeeWX Debian packages with the current debian/rules file. I had to modify it to make a .deb file.

Using the modified debian/rules file attached to this PR I was successful in building a deb package using the steps outlined at the bottom.

I used pandoc to convert from .md to .txt instead of just copying, I haven't explored if keeping a .md and .txt in sync in some other way might be better.

It would also be better to have the etc directory in the root weewx directory and not in the pkg directory, I'm not sure why etc is in the pkg directory as the Debian package tools will fail.

The Redhat package files in pkg directory also caused problems, I have no idea how to overcome this without deleting them.

## How to make a WeeWX Debian package from the development branch...
 * Install all the necessary packages needed to download and build the .deb file: `apt -y install devscripts debhelper dh-python python3-all python3-setuptools pandoc git`
 * `cd /usr/src`
 * Clone the repository: `git clone -b development https://github.com/weewx/weewx.git`
 * `cd weewx/pkg`
 * Move the etc folder to the root directory: `mv etc ..`
 * Remove the Redhat files in the pkg directory, a warning about the debian directory is fine: `rm *`
 * Copy a modified copy of the debian/rules file: `cp ~/rules debian/rules`
 * Update the debian/changelog using dch: `EMAIL="user@example.com" DEBEMAIL="$EMAIL" dch -v 5.1.99~dev1-1 "WeeWX 5.2 development branch build"`
 * `cd ..`
 * Build a source tarball: `git archive --prefix=weewx-5.1.99~dev1/ HEAD | xz -9 > weewx_5.1.99~dev1.orig.tar.xz`
 * `cd pkg`
 * Build the package: `debuild -us -uc`

## After all that you should have a shiny new Debian package...

You can install it using dpkg:
 * `dpkg -i /usr/src/weewx/weewx_5.1.99~dev1-1_all.deb`

### DIFF
--- a/DEV_NOTES.txt
+++ b/DEV_NOTES.txt
@@ -8,7 +8,7 @@ checklist for doing a release:
 3. Make sure all changes have been logged
      1. doc_src/changes.md
      2. doc_src/upgrade.md
-     3. make deb-changelog
+     3. make debian-changelog
      4. make redhat-changelog
      5. make suse-changelog
 4. Build the documentation
@@ -113,7 +113,7 @@ how to build wheel and source tarball:
   make pypi-package
 
 how to build debian package:
-  make deb-changelog
+  make debian-changelog
   emacs pkg/debian/changelog    # add package-specific changes, if any
   git commit -m "update deb changelog" pkg/debian/changelog
   make debian-package
@@ -371,7 +371,7 @@ that it finds.  that might not be what you want.
 
 the debian changelog *must* have a version number that matches the app version.
 the redhat package will build if the version numbers do not match.  use the
-rpm-changelog and deb-changelog targets to ensure that changelog versions match
+rpm-changelog and debian-changelog targets to ensure that changelog versions match
 the application version for a release.
 
 there are many ways to build a debian package.  first tried dpkg (uses DEBIAN


matthewwall:
hi @evilbunny2008 !  

the file DEV_NOTES.txt contains details about how to build a deb package, as well as rpm for redhat or suse. it also includes the pre-requisites for doing a package build. the `pkg` directory contains everything needed to build a package, whether that is for redhat, debian, suse, bsd, solaris, ...  

almost all of the artifacts have been moved from `pkg/etc` to the directory `src/weewx_data/util` - this is so that when the weewx configuration directory (e.g., /etc/weewx) is deployed, we put all of the ancillary files into that directory.  this makes it easier for someone to do os-integration.  for example, if you do a pip install on an os with systemd, you will have a functional unit file to install. or if you do a git 'install' on a bsd system, there will be rc files for you to use so you don't have to create it from scratch.

the files in `pkg/etc` are slightly different from those in `src/weewx_data/util`, because they are used to build the package, not used to show someone how to do what the package would do.  for example, when a package installs a systemd unit file, it puts it in /lib/systemd/system, but when a user installs a systemd unit file, s/he is supposed to put it in /etc/systemd/system.

note that as of weewx 5, the `pkg/etc` directory contains *only* systemd and udev resources, and these are written generically so that they will work on redhat, debian, or suse distributions.

matthewwall:
another thing to note is that any build of any package happens in the weewx source tree or optionally a writeable directory if the source tree is read-only.  this is by design, and goes way back to early gnu conventions for cross-compiling.  when you build a package (deb, rpm, whatever) it goes into a directory called `build` in the source tree.

matthewwall:
the two changes in the merge request have been incorporated into master:
1) fix the old 'deb-changelog' references in DEV_NOTES.txt
2) add option to mkchangelog.pl so that signing and locating USER/EMAIL from gpg info can be conditional
i used different semantics for the conditional, and instead of `SIGN` i used a more specific `GPG_KEYID` to control the signing.

i think we still need to document the steps for creating and optionally signing your own deb (or rpm) packages from the development branch.  that would be a good thing to add to DEV_NOTES.txt. lets do that in a new merge request.

but thank you @evilbunny2008 for testing/refining the packaging process.  your contributions have not only helped the process, but also led us to automated much more of it using vagrant.  your changes and the vagrant generalizations were added to the codebase in weewx 5.2

PR #1018, opened 2025-09-25
Urlencode authentication key

boosterl:
The WOW-BE service (see #1013 ) also allows full passwords as site authentication keys. If this password contains any special characters like `&` or `=`, this can mess up the query-string sent to the service.

This shouldn't cause any regressions for the WOW-service also, because the PIN codes used there aren't altered by URL-encoding them.

### DIFF
--- a/src/weewx/restx.py
+++ b/src/weewx/restx.py
@@ -1038,7 +1038,7 @@ class WOWThread(AmbientThread):
 
         _liststr = ["action=updateraw",
                     "siteid=%s" % self.station,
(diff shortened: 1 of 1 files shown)


evilbunny2008:
I didn't include any characters that need encoding in the WOW-BE station password, so it never occurred to me that this edge case could be a problem, so good catch @boosterl 

@tkeffer 
I've tested the change and WeeWX is happily posting to both WOW and WOW-BE, even with characters in the WOW-BE password like % & = etc

PR #1030, opened 2025-11-10
Update weewx.service

evilbunny2008:
weeWX was starting on my system before mariadb had started causing weeWX to exit on boot up

### DIFF
--- a/pkg/etc/systemd/system/weewx.service
+++ b/pkg/etc/systemd/system/weewx.service
@@ -8,6 +8,9 @@ After=time-sync.target
 Wants=network-online.target
 After=network-online.target
 
+After=mariadb.service mysqld.service
+Wants=mariadb.service mysqld.service
+
 [Service]
 ExecStart=weewxd /etc/weewx/weewx.conf
 StandardOutput=null


tkeffer:
I did not know that. Thanks!

Reopened.

mymaestro:
Just because I have mariadb on my server doesn't mean I want weewx to wait for it. I'm using sqlite3 for weewx and mariadb for other things.
Perhaps an install option like **weectl database reconfigure** would be more appropriate, so that every weewx installation doesn't have mariadb/mysql mentions clogging up the support lines; only the ones who need it.

evilbunny2008:
weeWX only supports MySQL, MariaDB and SQLite, so there is nothing else that will extend it unless Tom feels like supporting something else entirely such as PostgreSQL which doesn't remotely seem likely, but who knows what some bored geek with an itch may come up with in future...

For those using SQLite but still have MariaDB installed, what's the actual time penalty we're talking about, is it more than a fraction of a second?

SystemD was designed to start as much as it can at the same time, however for those of us using MariaDB this is absolutely needed otherwise weeWX exits and doesn't bother to retry and this is a far more eloquent solution compared to introducing code or other gymnastics in weeWX to achieve the same thing.

tkeffer:
Done with commit https://github.com/weewx/weewx/commit/4d470f34050cfe80da11b4da7dfe60efdf5486a8

I also uploaded the [revised docs](https://www.weewx.com/docs/5.2/usersguide/mysql-mariadb/#5-make-sure-mysqlmariadb-starts-before-weewx).

evilbunny2008:
I'm yet to get any feedback to my question here, or on the dev list, about how much delay there is if weeWX has to wait for a database to start. I don't expect it to be even a second to be honest.

What does weeWX really do in such a tiny time frame that would be at all detrimental?

During an archive event it can take quite a while depending on the CPU and templates needing to be processed. Far far far longer than a insignificant delay during boot-up, and boot-ups can also be pretty rare too, once a distro's kernel stabilises it's usually in the order of months.

Where as for people that do need such a delay, it's detrimental not having it, because weeWX exits, it doesn't bother retrying at all when the database isn't started.

As you said anyone with MySQL or MariaDB installed already know what they're doing, I expect the number of people having such software installed needing a delay would vastly out number the very small number of people that have a SQL DB installed but not using it for weeWX.

This change will have absolutely no effect on the vast majority of people just using the SQLite backend, because SystemD was designed specifically with this sort of thing in mind. Apart from the 2 commentators above, how many others have a SQL DB installed but not using it for weeWX?

Using a SQL DB comes into it's own when needing to handle queries from many things at the same time.

I'm not sure how many still use systems dedicated to weeWX, but I don't expect the percentage to be very high because most seem to have multi-use servers, or virtualised such as docker containers or even VPS.

So tell me how saving a precious second during a boot-up would have any real world benefit people's lives, compared to weeWX starting up and not finding the database and then exiting?

Not everyone has a UPS and the grids these days aren't as reliable as they should be given their price tags, which Davis owners may eventually find out when their memory corrupts, but then weeWX won't restart in such situations either, how many minutes or hours (on slow arm based CPUs) are then wasted waiting while weeWX catches up and handles past archive records once restarted manually?

evilbunny2008:
> Agree. Also agree with Tom's documentation changes. Please don't change the 'code' part of the service files.

Why not when it will benefit more users than it will cause a very minor inconvenience?

> We can't presume that a system with a db existing on it has that db present for weewx and not for other reasons. You also can't presume that a system with multiple weewx instances and multiple .conf files all use the same db.

Multiple instances is kinda besides the point, if only 1 needs it that's more than enough reason for it.

> I could see making the setup-daemon.sh script perhaps smarter and maybe even ask a question or two (or have optional --require_XYZ kinds of options), but this PR does not do attempt to do that.

Why do that when a simple change to the service file will fix everything? No code changes needed what so ever.

What specifically is there of any real world and actually noticeable and/or detrimental downside to this?

So far 2 people have raised an issue, without any hard data or facts to back their claim, that this would be a noticeable problem for them.

All I have seen so far is thought bubbles and emotive responses and nothing that this change would be actually noticeable at all.

I have lived experience that by not having those 2 lines there is a detrimental issue with weeWX because it doesn't wait or retry at all, it just exits.

If the power goes out overnight for whatever reason, or it's at a remote location, weeWX has to be then started manually and it can take, depending on hardware etc, a long time to catch up, and that's if it can catch up because not all stations use batteries or store past records.

> Personally I'd suggest perhaps adding commentary to the existing .service files saying 'uncomment if you use xyz' or the like as a reasonable middle ground. I know the o.p. will disagree.

Why? Why make everyone that needs this, something that weeWX promotes that it supports out of the box, have to go out of their way because 2 people think, not know, there may be some kind of inconvenience?

> I don't know how you'd support different requires in the [weewx@.service](mailto:weewx@.service) file specifically for the case where one instance requires db xyz and another doesn't or requires db foo. You'd almost need unique per-instance service files for that case meaning hand-editing.

Anything that complex is so far beyond the scope of this issue it's not funny, please backup your statements with hard data to show how a very very small delay during boot-up will actually have any real world noticeable effect, because I very much doubt anyone can, or they would have done so by now.

I thought the weeWX project would put the needs of the vast majority of users that do use a SQL DB over a couple of people with a weird setup, why is the "wants" of the few trumping the needs of the many?

To reiterate, the 2 lines have zero effect on those that don't have a SQL DB installed, so won't effect them in any case.

PR #1037, opened 2025-11-17
Update de.conf

mulecram:
Various translation adjustments DE


roe-dl:
I would suggest to check if the longer words fit into the respective fields of the web page and if the they make sense in the context on the web page. So, for example, if there is a head line saying "Sonne" (Sun), then "Aufgang" is enough  to label sunrise, and the full word "Sonnenaufgang" may seem too much.

PR #1040, opened 2025-11-26
Update control since python3-ephem is no longer updated

evilbunny2008:
Since the python module python3-ephem has had no updates since 2018 and is no longer maintained it makes no sense to have the weeWX Debian package depend on it.

### DIFF
--- a/pkg/debian/control
+++ b/pkg/debian/control
@@ -18,7 +18,6 @@ Depends:
  python3-pil,
  python3-serial,
  python3-usb,
- python3-ephem
 Suggests:
  sqlite,
  rsync,


roe-dl:
Please distinguish the kind of updates. PyEphem gets updates to adapt to new Python versions and in case of bugs. PyEphem gets no updates to its underlying ephemerides.

evilbunny2008:
@roe-dl any idea how much of a difference there is in accuracy since PyEphem last got a data update?

Are we talking a little bit or there are now major differences to what the skyfield extension outputs?

I assume different heavenly bodies probably drift at different rates...

PR #1055, opened 2025-12-25
Issue1053 - Add Uploaders section to User's Guide

vinceskahan:
Here's a PR to add some brief Uploaders documentation as well as linking it into the pullout menu on the left.  I *think* the links are ok, but did not rebuild all the docs via poetry to verify.

### DIFF
--- a/docs_src/usersguide/uploading.md
+++ b/docs_src/usersguide/uploading.md
@@ -22,10 +22,10 @@ These are located in the `[StdRESTful]` section of `weewx.conf`.
 Adds your system to the public weewx map of registered sites at
 [https://weewx.com/stations.html](https://weewx.com/stations.html).
 
-If enabled, the station will periodically (re)register with the
+If enabled, the station will periodically register with the
 weewx station registry to keep its registration active.  Stale
 registrations are automatically removed from the registry and map
-after 
+eventually if they are not reregistered.
 
 This is disabled by default.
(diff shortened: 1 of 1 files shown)


vinceskahan:
Ok thanks. Enjoy the feast !

-----  ***@***.*** ----

evilbunny2008:
One comment about the following:
`This legacy uploader supports the Met Office Weather Observations Website (WOW) which is being decommissioned beginning in January 2026.`

They're not actually decommissioning it, they're just no longer interested in getting the data directly from hobbyists any more, even though they rave on how useful the data is/was, going forward they're only interested in getting the data from companies making weather stations.

Which seems a little short sighted in my humble opinion, as advanced hobbyists such as those running weeWX, are more likely to care about the quality of data and best placement of sensors compared to some little old lady getting an xmas present and putting it up where it's the easiest place to put it.

vinceskahan:
@evilbunny2008 - I used the WOW wording, for better or worse.  I would prefer to not get into political and free/not-free discussions of any kind in any forum.

@tkeffer - thanks for the instructions for making the venv for poetry. Very cool.  Links all work.  Backticks applied. Reworded the registry sentences.  FWIW - rather than repeat the already fine howto information from the Reference, I chose to point to the specific section for each item.  I think it reads ok to my eyes.

PR #1062, opened 2026-01-18
update minimum python version required to 3.7

mrsnootles:
Am guessing the python 3.7 requirement under v5.2.0 in changes.md is correct and the docs were overlooked...

### DIFF
--- a/docs_src/usersguide/introduction.md
+++ b/docs_src/usersguide/introduction.md
@@ -7,7 +7,7 @@ WeeWX.
 
 ### Python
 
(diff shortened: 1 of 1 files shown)


tkeffer:
Nope. V3.6 or later will work. 

The information is buried away, but a backport needs to be installed to use V3.6. See [Install prerequisites](https://www.weewx.com/docs/5.2/quickstarts/git/?h=3.6#install-pre-requisites), item 5.

However, WeeWX v5.3.0 will require Python 3.7, mostly because some of the prerequisites require it.

PR #1076, opened 2026-03-31
Modified ReportTiming to check if files don't exist but would have been created otherwise

evilbunny2008:
I need to periodically generate reports but because the report_timing setting didn't match report generation was skipped even though data exists for the period in the database but files that would be generated don't exist.

A solution was to make minimal changes to the ReportTiming class to check if files the report would create exist and if the files don't exist allow report generation regardless of the report_timing setting.

Existing behaviour won't be altered by this PR, instead the proposed code checks for 'CreateIfMissing' appended to the report_timing string, eg @yearlyCreateIfMissing to enable the new functionality.

Distributing an extension just to extend the ReportTiming class doesn't seem like the best idea, and I thought this change would benefit others with the same need.

### DIFF
--- a/docs_src/custom/cheetah-generator.md
+++ b/docs_src/custom/cheetah-generator.md
@@ -434,7 +434,7 @@ There are several _aggregation periods_ that can be used:
     </tr>
     <tr>
         <td class="first_col code">$week</td>
-        <td>This week. The start of the week is set by option <a href="../../reference/weewx-options/stations/#week_start"><span
+        <td>This week. The start of the week is set by option <a href="../reference/weewx-options/stations/#week_start"><span
             class="code">week_start</span></a>.
         </td>
         <td class="code">$week.outTemp.max</td>
@@ -455,7 +455,7 @@ There are several _aggregation periods_ that can be used:
     <tr>
         <td class="first_col code">$rainyear</td>
         <td>This rain year. The start of the rain year is set by option <a
-            href="../../reference/weewx-options/stations/#rain_year_start"><span class="code">rain_year_start</span></a>.
+            href="../reference/weewx-options/stations/#rain_year_start"><span class="code">rain_year_start</span></a>.
         </td>
         <td class="code">$rainyear.rain.sum</td>
         <td>The total rainfall for this rain year. </td>
@@ -605,7 +605,7 @@ This table summarizes the options:
         <td class="code text_highlight">.ordinal_compass</td>
         <td>Format the value as a compass ordinals (<i>e.g.</i>, "SW"), useful
 	for wind directions. The ordinal abbreviations are set by option
-	<a href="../../reference/skin-options/units/#directions"><span class="code">directions</span></a>
+	<a href="../reference/skin-options/units/#directions"><span class="code">directions</span></a>
 	in the skin configuration file.
         </td>
     </tr>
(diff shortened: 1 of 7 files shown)


tkeffer:
The report timing functionality was Gary's thing --- I've never looked at it closely, but I'll do my best.

I get the directive `@yearlyCreateIfMissing`. But if the intention is to always trigger report generation even if just one file is missing, why does it have to be attached to a timing? Why not just `@createIfMissing`? So it would look like

    report_timing = @yearly, @createIfMissing

Or am I missing something?

The new function `dict_search()` looks useful. Move it to `weeutil/weeutil.py`. Document behavior. Include a unit test in `weeutil/tests/test_weeutil.py`. 

The file `reportengine.py` uses `pathlib`, instead of `os.path`. It's not a hard and fast rule, but it's nice to be consistent. If you want to stick with `os.path`, be sure to import it.

Document the new feature in the [Scheduling report generation](https://www.weewx.com/docs/5.3/custom/report-scheduling/) section.

Mention it in the change log.

evilbunny2008:
> Or am I missing something?

Not missing anything, I just didn't consider implementing it that way.

> Move it to `weeutil/weeutil.py`

Done

> The file `reportengine.py` uses `pathlib`, instead of `os.path`.

os was already imported for the set_cwd() function, I haven't used the pathlib module, but can re-code if you'd prefer that to be used.

> Document the new feature in the [Scheduling report generation](https://www.weewx.com/docs/5.3/custom/report-scheduling/) section.

Done

> Mention it in the change log.

Done.

After filing the PR it occurred to me that users will expect reports to run if template files have been modified or the extension updated, so the updated code checks the mtime of template files and generated files.

Added a checks to detect if html_dest_dir exists, otherwise the report is allowed in the case the user deletes it, which will prevent an Exception.

tkeffer:
Getting close. Just need unit tests for `dict_search()`.

evilbunny2008:
> Getting close. Just need unit tests for `dict_search()`.

Sorry overlooked that, and I wasn't feeling particularly inspired so the test data is based on a partial skin dict.

tkeffer:
It just occurred to me: what happens when the report includes `[[SummaryByMonth]]` and `[[SummaryByYear]]` sections? They use a `strftime()` template name. For example,

```
    [[SummaryByMonth]]
        # Reports that summarize "by month"
        [[[NOAA_month]]]
            encoding = normalized_ascii
            template = NOAA/NOAA-%Y-%m.txt.tmpl
```
This can resolve into dozens of files. Do all of them get checked?

evilbunny2008:
I feel manually running reports would be more appropriate for back filling and I assume the `weectl report` code already exists to do that.

Also I just noticed running `weectl` with no arguments causes an AttributeError.

For the current report, I'm going to go searching for the function that turns filename date strings into filenames. I will amend the PR once I find that function.

tkeffer:
This project feels like it needs a refactoring.

This implementation is not very modular. For example, class `ReportTiming` now needs knowledge of the details of how templates are specified and where their results go, something that is really in the domain of the `CheetahGenerator`. That creates brittleness.

Upon reflection, it also seems artificial that templates are subject to `@createIfMissing`, but not plots. 

There may be far simpler ways of solving the problem. For example, how about the first time the report generator gets run, it optionally ignores `report_timing` and generates everything. After that first run, `report_timing` comes into play. With this strategy, nothing would be missing. 

Or `report_timing` applies on an individual file level, instead of for a whole report (somewhat like `stale_age`)? This way you could selectively generate individual files or plots on a schedule. The go/no-go decision would be made deep within the loops inside the Cheetah and image generators, so the report engine would not need any information about how they specify their products. Indeed, the entire class `ReportTiming` would disappear.

I guess I'd like to understand your use case better. How does a file come to be missing?

Other comments, although not as important:

- I'm not keen on implementing your own version of `strftime()`. It's a lot of extra code and could easily miss nuances. 

- Function `getFileName()` does not really belong in `weeutil/weeutil.py`. Its use case is very specific to the Cheetah generator.

evilbunny2008:
I didn't implement anything in the latest push, I just moved code about.

PR #1080, opened 2026-04-12
Udev rules fix for acurite stations on newer kernels

zerog2k:
Added a workaround to unbind the hid-generic driver for Acurite devices on newer kernels, due to buggy Acurite hid implementation. I believe this should be backwards compatible with all modern kernels (i.e. 5+)

Issue widely reported with newer kernels on weewx-user list, e.g. https://groups.google.com/g/weewx-user/c/0Jiye8CmtLE/m/Bob4QRxzAgAJ

Submitting PR to master, as this is a fix.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -18,6 +18,7 @@ Use better measure of system uptime.
 Fix problem that prevented table headers from appearing in the Seasons's 
 statistics page.
 
+Fix problem with Acurite usb failures in newer kernels. [PR #1080](https://github.com/weewx/weewx/pull/1080)
 
 ### 5.3.1 03/03/2026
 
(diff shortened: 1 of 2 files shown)


tkeffer:
Has this been tested on both bookworm and trixie?

Also, could you add an entry to the change log?

zerog2k:
Thanks @tkeffer
I think I updated the right changelog spot?
I only have Ubuntu 24 and a Debian Bullseye system to test this with currently. I have put out the call for more folks to help verify with Bookworm & Trixie on the thread. (I saw a few other related threads, when searching for 'acurite' in the past month or two - not sure if it's worth poking on those threads as well?)

tkeffer:
Looks good. Thanks.

Gribnif:
In case it helps anyone, I can confirm that this fixed a problem I was having with CentOS 5.14.0-708.el9.x86_64.

PR #1086, opened 2026-04-19
fix(pkg): read HTML_ROOT from config for chown/chmod during upgrade

### LINKED ISSUE #1046: use HTML_ROOT to determine where to chown/chmod
during an upgrade (and new install?) use HTML_ROOT to determine where to chown/chmod

CodeLine9:
During upgrade, \`setup_reporting_dir\` hardcoded the HTML_ROOT path instead of reading it from the configuration file. If a user had customized HTML_ROOT in weewx.conf, permissions would be applied to the default location, leaving the actual reporting directory with wrong ownership.

Add \`get_html_root()\` helper to both RPM spec and Debian postinst that reads HTML_ROOT from weewx.conf, falling back to the default when the config file does not exist (fresh install).

Fixes #1046

### DIFF
--- a/pkg/debian/postinst
+++ b/pkg/debian/postinst
@@ -433,11 +433,24 @@ setup_database_dir() {
     chown -R $WEEWX_USER:$WEEWX_GROUP $WEEWX_HOME
 }
 
+# get HTML_ROOT from the configuration file
+get_html_root() {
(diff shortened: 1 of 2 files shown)


CodeLine9:
Closing — no longer pursuing contributions here.

PR #1088, opened 2026-04-24
Add option record_stats to rsyncupload to record upload stats in archive

n7qnm:
Patch to rsyncupload and reportengine to record rsync statistics (rsyncFiles, rsyncBytes and rsyncDuration) to archive.

These can be used to track upload speed over time.

To use, see https://github.com/n7qnm/weewx-rsync-metrics, install rysnc_mettrics.py and set "record_stats = true" in the weewx.conf [rsync] stanza.

If record_stats is not set, the changes are innocuous.

### DIFF
--- a/src/weeutil/rsyncupload.py
+++ b/src/weeutil/rsyncupload.py
@@ -4,10 +4,11 @@
 #
 #    3-Jan-2021. Refactored by tk
 #    3-Apr-2024. Added option rsync_options.
+#    17-Apr-2026. Added option record_stats to capture upload metrics.
 #
 #    See the file LICENSE.txt for your full rights.
 #
-"""For uploading files to a remove server via Rsync"""
+"""For uploading files to a remote server via Rsync"""
 
 import errno
 import logging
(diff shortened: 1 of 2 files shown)


tkeffer:
I get that you are trying to intercept the statistics created by `RsyncUpload`, then send them upstream where they can be put in the database. But I think there are ways of doing this that are not so intrusive to the code base.

The simplest would be to modify `RsyncUpload.run` to simply return the values that you need. Then write an extension that subclasses `RsyncUpload`, which receives the values. That would be the place to put your thread locking mechanisms. This way the only change in the WeeWX code base is a simple `return` statement.

Another approach that would require zero changes in the code base would be to write a specialized logging handler. It would intercept any messages from `weeutil.rsyncupload` and extract the statistics you need.

n7qnm:
Thanks for the reply - good points.   Perhaps a service that could be added, and would write messages to MQTT or someplace else based on a config file entry.

Let me think about that.

Clay

________________________________

andylittle:
Just a thought, does this belong in [weewx-cmon](https://github.com/matthewwall/weewx-cmon)

Andy

n7qnm:
Hmmm - perhaps.

Given the subclass, cmon could probably reach out and grab the data.   My intent was a little more "generic", because I use Zabbix to monitor my weather server and a whole lot more.

I'll ping Mr. Wall

Clay

________________________________

n7qnm:
Closing request - will rework as suggested by Tom and provide a service w/no core mods required

PR #1095, opened 2026-06-05
Seasons tags

roe-dl:
In Meteorology seasons statistics are also very common. This PR adds seasons timespans to the list of available aggregation periods. It refers to the thread "[Seasons Aggregations](https://groups.google.com/g/weewx-user/c/Av_osAxSbbM)" at the Google user group.

Considerations:

Like `$month` for the month and `$year` for the year, the tag `$season` provides the timespan of the actual season. And like the other tags the parameter `seasons_ago` can be used to get past seasons. Additionally the parameter `years_ago` can be used to get the same season as the current season but within another year.

`$seasonsyear` provides the timespan of a whole year starting with spring and ending with winter. Hemisphere is observed. It can be used together with the `.seasons` iterator to write tables of seasons statistics.

There are two common types of seasons: meteorological and astronomical. Both are widely used. The differences are as follows:
* **Meteorological seasons** always start at the 1st of a month, that is, at March 1st, June 1st, September 1st, and December 1st.
* **Astronomical seasons** start at the day of the equinox or solstice.

The following table shows it for the northern hemisphere. For the southern hemisphere exchange spring with autumn and summer with winter:

Season | Meteorological season | Astronomical season
--------|------------------------|--------------------
Spring | March, April, May | March equinox to June solstice
Summer | June, July, August | June solstice to September equinox
Autumn | September, October, November | September equinox to December solstice
Winter | December, January, February | December solstice to March equinox of the next year

As default I set the meteorological seasons. There is no special reason for that choice besides the fact, that we do meteorology with WeeWX. To get the astronomical seasons use the parameter `season_type` with `season_type='astronomical'`. 

The iterator `.seasons` can be used with both `$seasonsyear` and `$year`. The parameter `season_type` is available for both. 

Why let the astronomical seasons start at the **day** of the event and not at the moment of the event itself? The answer is, that WeeWX has pre-calculated daily summaries, that can only be used if the timespan starts and ends at a day boundary according to local time. And meteorological observations show a daily course. So it seams to me no use to start midday.


tkeffer:
Thanks for this!

I made a few very minor changes.

PR #1097, opened 2026-06-12
Modified ReportTiming to check if files don't exist but would have been created otherwise

evilbunny2008:
I need to periodically generate reports but because the report_timing setting didn't match report generation was skipped even though data exists for the period in the database but files that would be generated don't exist.

A solution was to make minimal changes to the ReportTiming class to check if files the report would create exist and if the files don't exist allow report generation regardless of the report_timing setting.

Existing behaviour won't be altered by this PR, instead the proposed code checks for 'CreateIfMissing' appended to the report_timing string, eg @yearlyCreateIfMissing to enable the new functionality.

Distributing an extension just to extend the ReportTiming class doesn't seem like the best idea, and I thought this change would benefit others with the same need.

I still need this functionality but #1076 was closed when I shifted my changes to a different branch so I could submit another PR without having changes from this PR

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -32,7 +32,7 @@ report to identify bottlenecks.
 Added @createIfMissing option for report_timing handling so that report
 generation will be allowed if there are missing report files or if the
 modified times of templates are newer than generated files.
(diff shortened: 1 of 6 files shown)


tkeffer:
Squash merged with commit 13d23a6b0525922f3879f5592110e6fb05f24872

Issue #1094, opened 2026-06-04
weeWX won't keep trying if MariaDB/MySQL goes away after connected to


evilbunny2008:
I have `loop_on_init = True` in weewx.conf but this only helps during start up

Unless I've overlooked something, there doesn't seem to be an option to keep weeWX from exiting when MariaDB/MySQL goes away after a connection has been made

Server log:
```
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__: Caught unrecoverable exception:
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****  query: b'SELECT MAX(dateTime) FROM archive'
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****  Traceback (most recent call last):
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****    File "/usr/share/weewx/weewx/engine.py", line 210, in run
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****      self.dispatchEvent(weewx.Event(weewx.CHECK_LOOP, packet=packet))
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****      ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****    File "/usr/share/weewx/weewx/engine.py", line 241, in dispatchEvent
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****      callback(event)
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****      ~~~~~~~~^^^^^^^
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****    File "/usr/share/weewx/weewx/engine.py", line 662, in check_loop
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****      raise BreakLoop
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****  weewx.engine.BreakLoop
Jun 04 09:50:16 weewx-server weewxd[1128304]: CRITICAL __main__:     ****
[...]

PR #1098, opened 2026-06-12
MySQLdb has 2 sets of Exceptions

evilbunny2008:
importing from MySQLdb doesn't import the same base exception classes as those defined in MySQLdb._exceptions, something to do with C verse python exceptions

### DIFF
--- a/src/weedb/mysql.py
+++ b/src/weedb/mysql.py
@@ -18,7 +18,7 @@ except ImportError:
                          InterfaceError as MySQLInterfaceError)
 else:
     try:
-        from MySQLdb import DatabaseError as MySQLDatabaseError
+        from MySQLdb._exceptions import DatabaseError as MySQLDatabaseError
     except ImportError:
         from _mysql_exceptions import (DatabaseError as MySQLDatabaseError,
                                        InterfaceError as MySQLInterfaceError)
@@ -66,6 +66,8 @@ def guard(fn):
             raise klass(e)
         except MySQLInterfaceError as e:
             raise weedb.DisconnectError(e)
+        except MySQLOperationalError as e:
+            raise weedb.DatabaseError("test")
 
     return guarded_fn
 
--- a/src/weedb/mysql.py
+++ b/src/weedb/mysql.py
@@ -66,8 +66,6 @@ def guard(fn):
             raise klass(e)
         except MySQLInterfaceError as e:
             raise weedb.DisconnectError(e)
-        except MySQLOperationalError as e:
-            raise weedb.DatabaseError("test")
 
     return guarded_fn
 


tkeffer:
I'm not sure how we're improving things here. Indeed, what's broken?

The stated intention of the `MySQLdb` implementation is to follow [PEP 0249](https://peps.python.org/pep-0249/#exceptions), which pretty clearly states that `DatabaseError` should be imported directly from the module. There is no mention of `MySQLdb._exceptions`.

evilbunny2008:
Catching DatabaseError imported from `__init__.py` should have handled an OperationalError as it's a derived class, but `__init__.py` exposes them by republishing the exception classes and in the process the classes are no longer derived.

This change uses the original classes rather than copies so that DatabaseError will properly catch exceptions raised.

tkeffer:
I'm not following. My version works fine. What is actually broken?

Can you give me a short snippet that breaks? Also, what version of `mysqlclient` are you using, etc? I'd like to reproduce whatever problem this is.

In using the exceptions from `MySQLdb._exception` there is a risk that we could break some other implementation out there.

evilbunny2008:
As shown below the version in Debian stable is getting a bit long in the tooth, and there is no newer versions packaged, not even in unstable:
```
--- usr/lib/python3/dist-packages/MySQLdb/connections.orig.py	2019-11-19 19:28:43.000000000 +1100
+++ usr/lib/python3/dist-packages/MySQLdb/connections.py	2026-05-11 03:22:56.032260529 +1000
@@ -236,7 +236,11 @@
         # Since _mysql releases GIL while querying, we need immutable buffer.
         if isinstance(query, bytearray):
             query = bytes(query)
-        _mysql.connection.query(self, query)
+
+        try:
+            _mysql.connection.query(self, query)
+        except Exception as e:
+            raise OperationalError(f"query: {query}")
 
     def _bytes_literal(self, bs):
         assert isinstance(bs, (bytes, bytearray))
```

> Also, what version of mysqlclient are you using, etc? I'd like to reproduce whatever problem this is.

The Debian package is named python3-mysqldb, probably because that's what the author named the directory on GitHub
```
# apt-cache search python3|grep mysql
python3-aiomysql - library for accessing a MySQL using asyncio (Python 3)
python3-mysqldb - Python interface to MySQL
python3-pymysql - Pure-Python MySQL Driver - Python 3.x
```

Doing a search the other way round doesn't produce anything more relevant
```
# apt-cache search mysql|grep -i python
python-aiomysql-doc - library for accessing a MySQL database from the asyncio (common documentation)
python3-aiomysql - library for accessing a MySQL using asyncio (Python 3)
python3-mariadb-connector - Enables Python programs to access MariaDB and MySQL database
python3-peewee - Simple ORM for PostgreSQL, MySQL and SQLite (Python 3)
python3-asyncmy - Fast asyncio MySQL/MariaDB driver (Python 3)
python3-mysqldb - Python interface to MySQL
python-pymysql-doc - Pure-Python MySQL driver - doc
python3-pymysql - Pure-Python MySQL Driver - Python 3.x
python3-typeshed - collection of library stubs for Python, with static types
```

As I said, the Debian stable package is an old version

```
# dpkg --list|grep python3-mysqldb
ii  python3-mysqldb                   1.4.6-2+b6                           arm64        Python interface to MySQL
```

> In using the exceptions from MySQLdb._exception there is a risk that we could break some other implementation out there.

This occurred to me as well, but when I checked the GitHub repo, exceptions are still defined in [_exceptions.py](https://github.com/PyMySQL/mysqlclient/blob/main/src/MySQLdb/_exceptions.py) but [`__init__.py`](https://github.com/PyMySQL/mysqlclient/blob/v1.4.6/MySQLdb/__init__.py) changed how exceptions are republished since v1.4.6

tkeffer:
Doing a little research, Debian 13 uses (as you note) V1.4.6. In this version `DatabaseError` is imported from `MySQLdb._mysql`.

Pypi.org has version v2.2.8. In this version, `DatabaseError` is also imported from `MySQLb._mysql`. So no differences there.

I don't think the version number is involved.

This discussion started with Issue #1094, where you thought that WeeWX would not recover the server went away. If I understand correctly, you closed that issue because you had made changes that you had forgotten about.

So, what exactly are we trying to solve here? I need something I can reproduce.

evilbunny2008:
> If I understand correctly, you closed that issue because you had made changes that you had forgotten about.
>
> So, what exactly are we trying to solve here? I need something I can reproduce.

Yes, the changes I made are in the diff in my previous comment. However while looking into this issue further I found out that the OperationalError exception should have been caught by the code handling DatabaseError exceptions but because the way the python3-mysqldb package republishes exceptions, it isn't caught and weeWX will exit as seen in #1094.

This PR aims to address this short coming in the most minimal way possible.

tkeffer:
Sorry if I'm being obtuse, but I'm still not seeing what we're trying to fix.

You made a change to the codebase of a very stable underlying library, which resulted in unexpected behavior. I'm not seeing the reason why we should be changing the WeeWX code to accomodate this change. 

What was the original reason for your change? What were you trying to fix?

One thing I suppose we could do is catch both versions of the exceptions, but I'd still like to know why.

evilbunny2008:
> Sorry if I'm being obtuse, but I'm still not seeing what we're trying to fix.

An un-handled exception leading to weeWX exiting.

> You made a change to the codebase of a very stable underlying library, which resulted in unexpected behavior.

The change should have still been caught and handled and weeWX continued to retry connecting to the database which had been restarted.

However the exception was a corrupted copy so weeWX didn't handle it and weeWX exited. The fact that I made a change isn't the point and in fact other parts of the same code will raise the same exception that I did, the point is it wasn't handled and so weeWX stopped working for 8 hours until I noticed it wasn't running.

> What was the original reason for your change? What were you trying to fix?

I was trying to find out what SQL query was causing me problems, but that's not the point, weeWX shouldn't have exited like that for something it should be catching but didn't because the exception raised wasn't derived like your code expects it to be because MySQLdb doesn't expose the original defined exceptions properly, only copies, so the exception isn't handled like you intend.

> One thing I suppose we could do is catch both versions of the exceptions, but I'd still like to know why.

Because weeWX will stop retrying and exit if you don't.

tkeffer:
Sorry, but I just don't see it that way. The Python DBAPI has a contract that if the server goes away, it will raise an exception derived from `DatabaseError`, which is defined in the module itself. Code that uses the API have a right to expect that any conformant raised exception is also rooted in `DatabaseError`. They should not be expected to catch any exception raised in `MySQLdb._exception`.

In particular, the leading underscore in `_exception` implies that it is an internal implementation detail, and should not be relied on. By contrast, the version in `MySQLdb` is a public facing version.

If you want to raise an exception in the `MySQLdb` implementation, I would suggest raising `MySQLdb.DatabaseError`, and not `DatabaseError`. It's the public, stable interface.

evilbunny2008:
> The Python DBAPI has a contract that if the server goes away, it will raise an exception derived from `DatabaseError`

That's the crux of the problem, in v1.4.6 it's not derived it's a copy.

> which is defined in the module itself.

No, it's defined in _exceptions.py

> They should not be expected to catch any exception raised in `MySQLdb._exception`.

No they shouldn't.

> By contrast, the version in `MySQLdb` is a public facing version.

But that's not where they're actually defined, everything is defined in _exceptions.py, for both the current version and v1.4.6

```
_exceptions.py:class DatabaseError(Error):
_exceptions.py:class DataError(DatabaseError):
_exceptions.py:class OperationalError(DatabaseError):
_exceptions.py:class IntegrityError(DatabaseError):
_exceptions.py:class InternalError(DatabaseError):
_exceptions.py:class ProgrammingError(DatabaseError):
_exceptions.py:class NotSupportedError(DatabaseError):
```

> If you want to raise an exception in the `MySQLdb` implementation, I would suggest raising `MySQLdb.DatabaseError`, and not `DatabaseError`. It's the public, stable interface.

I checked what exceptions are commonly raised in MySQLdb and ProgrammingError is one example in cursors.py. So I reverted changes and changed the exception raised and it's not caught either. The import code in cursors.py and other files uses:

```
from ._exceptions import (
    Warning, Error, InterfaceError, DataError,
    DatabaseError, OperationalError, IntegrityError, InternalError,
    NotSupportedError, ProgrammingError)
```

evilbunny2008:
> If you want to raise an exception in the `MySQLdb` implementation, I would suggest raising `MySQLdb.DatabaseError`

That doesn't work, `MySQLdb._exceptions.DatabaseError` still appears in the trace even though I raised a `MySQLdb.DatabaseError` exception.

Because connections.py doesn't normally import MySQLdb that initially threw a different exception, so it changed what was executed, and weeWX exited under both conditions.

PR #1100, opened 2026-06-13
feat: Introducing weectl rest -command

HQJaTu:
When sending data to 3rd party services / interfaces, this happens on Weewx default scheduler bi-hourly.

During setup or troubleshooting getting a data-packet every 30minutes is waaaaay too slow. A data packet should be sent at admin's command. Unfortunately `weectl` lacks such subcommand handling data sending.

This PR introduces a new command. Usage:
```text
usage: weectl rest list
            [--config=FILENAME]

       weectl rest run [NAME ...]
            [--config=FILENAME]


List RESTful services, or force an upload

options:
  -h, --help  show this help message and exit

Which action to take:
  {list,run}
    list      List all configured RESTful services
    run       Force an upload to RESTful services
```

Functionality follows other `weectl` subcommands. Admin can get a list of integrations or additionally trigger a data send on all or a selected one. This fills the need for being able to send data at admin's request.

### DIFF
--- a/docs_src/utilities/weectl-rest.md
+++ b/docs_src/utilities/weectl-rest.md
@@ -1,0 +1,75 @@
+# weectl rest
+
+Use the `weectl` subcommand `rest` to list the configured RESTful upload
+services, and to force an upload on demand.
+
+RESTful services are the uploaders that publish your data to online weather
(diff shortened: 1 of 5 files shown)


tkeffer:
Merged into the `development` branch.

Nice work! Very complete. The only thing I added is a change log entry.

PR #1103, opened 2026-06-24
Update de.conf

mrclgtb:
a few adjustments (de)


src/weewx_data/skins/Seasons/lang/de.conf
```diff
@@ -40,9 +40,9 @@ unit_system = metricwx
         barometer              = Luftdruck         # QFF
         barometerRate          = Luftdruckänderung
         cloudbase              = Wolkenuntergrenze
-        dateTime               = "Datum/Zeit"
+        dateTime               = Datum/Zeit
         dewpoint               = Taupunkt
-        ET                     = Evapotranspiration
+        ET                     = Evapotranspiration (Verdunstung)
```
tkeffer: I worry that this line could be very long for labeling purposes. Can it be shortened to simply "Verdunstung"?

tkeffer (approved):

src/weewx_data/skins/Seasons/lang/de.conf
mrclgtb: Hi @tkeffer  thank you very much for pointing that out. Of course. I don't even understand myself why I included “ET” and the translation or term associated with it. I've corrected both to german "Verdunstung".

roe-dl:
@tkeffer Unfortunately I was on vacation the last weeks. So I could not protest in time.

German "Evapotranspiration" (English evapotranspiration) is not the same as "Verdunstung" (English evaporation). This is from Wikipedia:

German:
> Evapotranspiration bezeichnet in der [Meteorologie](https://de.wikipedia.org/wiki/Meteorologie) die Summe aus [Transpiration](https://de.wikipedia.org/wiki/Transpiration) und [Evaporation](https://de.wikipedia.org/wiki/Evaporation)

English:
> Evapotranspiration (ET) refers to the combined processes that moves water from the Earth's surface (open water and ice surfaces, bare soil and [vegetation](https://en.wikipedia.org/wiki/Vegetation)) into the [atmosphere](https://en.wikipedia.org/wiki/Atmosphere_of_Earth) including both [evaporation](https://en.wikipedia.org/wiki/Evaporation) and [transpiration](https://en.wikipedia.org/wiki/Transpiration).

And in contrary bare "Verdunstung":

German:
> Bei einer Verdunstung geht ein [Stoff](https://de.wikipedia.org/wiki/Stoff_(Chemie)) vom [flüssigen](https://de.wikipedia.org/wiki/Fl%C3%BCssigkeit) in den [gasförmigen](https://de.wikipedia.org/wiki/Gasf%C3%B6rmig) Zustand über, ohne dabei die [Siedetemperatur](https://de.wikipedia.org/wiki/Siedepunkt) zu erreichen.

English:
> Evaporation is a type of [vaporization](https://en.wikipedia.org/wiki/Vaporization) that occurs on the [surface](https://en.wikipedia.org/wiki/Interface_(chemistry)) of a [liquid](https://en.wikipedia.org/wiki/Liquid) as it changes into the gas phase.

So you cannot translate the English "Evapotranspiration" to german "Verdunstung" in labels.

I understand, if someone has difficulties remembering latin or greek terms in science. But I do not see that that is a matter of labels. I would suggest that the user makes the label a link to some kind of explanation or description instead of changing the label to a word with another meaning, only to have an original german word.

tkeffer:
OK, so as I understand it, just leaving it at a simple "Evapotranspiration" is the better choice.

roe-dl:
> OK, so as I understand it, just leaving it at a simple "Evapotranspiration" is the better choice.

I think so.

tkeffer:
Changed in commit https://github.com/weewx/weewx/commit/c568bfc7273a3ba213fa6d2d50e5380298a00f1e

mrclgtb:
@roe-dl 

> @tkeffer Unfortunately I was on vacation the last weeks. So I could not protest in time.

This is, after all, a matter of translation. In this case, however, “Evapotranspiration” could be translated as “Gesamtverdunstung.” But yes, here too, one can debate what is meant by it, just as with the term “Evapotranspiration” itself.
I asked also various LLMs about this, and they all returned “Gesamtverdunstung” as the German term here. Of course, every user should examine the term more closely to understand its exact meaning.

roe-dl:
>  ... I asked also various LLMs about this, ...

I would not rely on LLMs. They provide a lot of stupid answers. You can ask Wikipedia, which will often provide a good result (but not always). You can look into the [Wetter- und Klimalexikon](https://www.dwd.de/DE/service/lexikon/lexikon_node.html) of the German Weather Agency (Deutscher Wetterdienst - DWD). You can ask a German meteorologist. Or you can consult well-known scientific literature. The word "Gesamtverdunstung" I never heard before. For WeeWX I would always use words, that are common in scientific use. 

A LLM may be useful to get the sources to consult. But the answer itself is of no use, as long as you have not checked it independently.

mrclgtb:
> I would not rely on LLMs. They provide a lot of stupid answers.

Of course, but not always. Still, it's a good idea to check. The term “Gesamtverdunstung” can be found, for example, on the DWD website -> https://www.dwd.de/DE/fachnutzer/landwirtschaft/dokumentationen/allgemein/basis_verdunstung_doku.html?lsbId=510228

"Die Evapotranspiration ist die Gesamtverdunstung..."

LLMs do need a source for this, after all ;-)

But okay, I don’t want to start an discussion here. Everyone can adjust their translation within their weewx instances however they like. So everything’s fine.

> For WeeWX I would always use words, that are common in scientific use.

That makes sense!

roe-dl:
> "Die Evapotranspiration ist die Gesamtverdunstung..."

Das besagt aber nicht, daß die Begriffe äquivalent sind.

That does not say, both words have the same meaning.

> LLMs do need a source for this, after all ;-)

They are known to invent sources sometimes.

> But okay, I don’t want to start an discussion here.  ...

I think, it is ok to discuss how to express something properly. But the use cases are different. So we will not always find a solution that fits all needs.

PR #1110, opened 2026-08-22
Bash completion for weectl

evilbunny2008:
There is no bash completion for weectl.

For install and uninstall, I included references everywhere that weewx.rules appears, but I mostly only saw references in the Debian packaging files.


tkeffer:
Great addition! I was hoping someone would do this.

How about something for the pip users? The pip daemon install process already has users running

    sudo sh ~/weewx-data/scripts/setup-daemon.sh

We could include installing your bash completion as part of this, but I think it would be better to keep the two steps separate.

This would mean supply a dedicated install script under `src/weewx_data/scripts` along with instructions on how to use it.

evilbunny2008:
Nothing needs to be added to the bash completion directory as far as I know, people just need to `source file`

tkeffer:
I'm not an expert, but as I understand it, a user either has to source the completion script from their `.bashrc`, or it needs to be included in the bash completion directory.

Either way, I would suggest including instructions for the pip user.

hilman2:
@evilbunny2008 I put two commits up against your `feature-branch-3` —
evilbunny2008/weewx#1 — so they arrive here as part of your PR rather than as a
competing one. Yours to take, change, or ignore.

**One is a bug in the packaging as it stands.** `DST_CFGDIR` is `/etc/weewx`, so the
file lands in `/etc/weewx/bash_completion.d/weectl`, which nothing reads, and `postrm`
removes `/etc/bash_completion.d/weectl`, which was never written. The pattern in this
repo is to stage files under `/etc/weewx` and let `postinst` copy them into place — that
is what `setup_udev` and `setup_init` do — and that step is missing for the completion.
So the completion does not currently take effect from a `.deb` or `.rpm`.

**The other is the pip half @tkeffer asked for.** `pkg/` is not in the wheel:
`pyproject.toml` packages the Python packages and `weewx_data` and nothing else, so a
pip user never receives the file at all. Moving it to
`src/weewx_data/util/bash_completion.d/` gets it copied to `~/weewx-data/util` by
`weectl station create`, the same treatment `util/import` and `util/init.d` already get,
and the Debian rules and RPM spec take it from there — one copy of the file instead of
two. Then a `setup-bash-completion.sh` next to `setup-daemon.sh`, and a section in the
pip quickstart.

On whether it needs sourcing: there is a third option, which is what these commits use.
Install the file under the **name of the command** in a completion directory, and
bash-completion loads it on demand the first time someone types `weectl` — nothing to
source, nothing in `.bashrc`, and no root for the per-user case. Ubuntu 24.04,
bash-completion 2.11, shell started with `--norc`, no `.bashrc` read:

```
before: bash: complete: weectl: no completion specification
after:  complete -F _weectl weectl
weectl station -> create reconfigure upgrade list-drivers -h --help
```

Nice piece of work, by the way — 287 lines of it, down to the per-subcommand options.

---

Testing and the rest of the detail are in the linked PR. One thing worth knowing if you
develop on Windows: git hands you the completion file with CRLF endings there, and
`complete -F _weectl weectl\r` then registers a completion for `weectl\r`, so nothing
works and nothing says why. On Linux it is LF and fine.

evilbunny2008:
> Nice piece of work, by the way — 287 lines of it, down to the per-subcommand options.

I wish I could take the credit, but I pointed Claude at a zip file of the current WeeWX source code and it's getting pretty darn good with far fewer mistakes than it used to, but you still need to check the output.

tkeffer:
Thanks. A great addition!

PR #1112, opened 2026-08-24
Use `importlib` to load an extension's `install` module

### LINKED ISSUE #1109: Use importlib instead of __import__
The WeeWX engine uses the legacy `__import__` function. Nowadays, `importlib` is universally available. We should use it.

hilman2:
Closes #1109. Against `master`, as this is a fix rather than a feature.

Do you want a change log entry? Your own commit for the `engine.py` half did not get
one, so I left it out.

## What

3a7223de took care of `engine.py` on the day the issue was opened. The one in
`weecfg.get_extension_installer()` was the last `__import__` in the tree; there are now
none.

`importlib.import_module()` returns the module, so the `sys.modules['install']` lookup
that followed the call goes away -- the same shape you gave `weedb` in e129a14d.

What has to stay is the `sys.modules.pop()`. Every extension calls its installer module
`install`, so leaving it in the cache hands the next extension installed by the same
process the previous one's installer. The comment there said only "Get rid of the
module", which does not say why it matters. It does now.

`get_extension_installer()` had no tests of its own. There are three now: an ordinary
load, a directory with no `install.py`, and two extensions in a row. Removing the
`pop()` fails two of them -- the second extension comes back holding the first one's
installer.

The second commit is cosmetic. 3a7223de replaced `import sys` with `import importlib`
where it stood, which left it between `socket` and `threading` in an otherwise
alphabetical block. Drop that commit if you would rather not have it.

Nothing else changes: same exception on a missing `install.py`, same return value, no
new dependency.

## Testing

Full suite, one file per run the way the makefile does it, against `origin/master` and
against this branch. Ubuntu 24.04, Python 3.12, MariaDB 10.11, `make test-setup-ci`.

|                 | passed | failed |
|-----------------|--------|--------|
| `origin/master` | 374    | 0      |
| this branch     | 377    | 0      |

The three extra are the new tests. The output is otherwise identical line for line.

`vermin --target=3.7` reports the same minimum before and after (3.6).

### DIFF
--- a/src/weecfg/__init__.py
+++ b/src/weecfg/__init__.py
@@ -715,12 +715,13 @@ def get_extension_installer(extension_installer_dir):
         sys.path.insert(0, extension_installer_dir)
         try:
             # Now I can import the extension's 'install' module:
-            __import__('install')
+            install_module = importlib.import_module('install')
         except ImportError:
             raise ExtensionError("Cannot find 'install' module in %s" % extension_installer_dir)
-        install_module = sys.modules['install']
         loader = getattr(install_module, 'loader')
-        # Get rid of the module:
+        # Every extension names its installer module 'install', so it must not be left in
+        # the module cache: the next extension installed by this process would get this
+        # one's installer back instead of its own.
         sys.modules.pop('install', None)
         installer = loader()
     finally:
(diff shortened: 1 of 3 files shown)


hilman2:
Closing in favour of #1122, which now does this as well. Both touched the same fifteen
lines of `get_extension_installer()`, so one PR is easier to review than two.

The tests and the second commit went with it.

PR #1113, opened 2026-08-24
Document group `[DisplayOptions]`

### LINKED ISSUE #977: Document group [DisplayOptions]

hilman2:
Closes #977. Against `master`, where the other documentation commits are.

Writing this turned up three places where the comments in the *Seasons* `skin.conf` do
not match what the templates do (below). The page follows the templates and leaves the
comments alone. Want them fixed too, here or separately?

## What

The section is not documented. Neither `docs_src` nor the wiki mentions
`[DisplayOptions]` or the tag `$DisplayOptions`, beyond one `skin.conf` excerpt in
[v5 performance troubleshooting](https://github.com/weewx/weewx/wiki/v5-performance-troubleshooting).

The page is modelled on [`[Extras]`](https://weewx.com/docs/5.5/reference/skin-options/extras/),
since it is the same kind of thing: a section WeeWX passes through to the
templates without interpreting it. So it says what the mechanism is, documents
the options *Seasons* recognizes, and ends with how to add your own.

Each option gives the default the template falls back on when the option is
absent, which is not the value that ships in `skin.conf` -- the shipped lists
are much longer. It also covers `season_type`, which `statistics.inc` reads but
`skin.conf` does not set.

Placed after `[Extras]` in the nav for the same reason.

## The three mismatches

I started from the comments in `skin.conf` and checked each one against the
templates:

- There is no "Statistical Summary" section anywhere in the skin.
  `observations_stats` feeds the "Statistics" widget on the front page and the
  table on `statistics.html`.
- The sensor lists are not a page of their own. They appear under
  "Connectivity", "Battery Status" and "Voltage" inside the "Sensor Status"
  widget, which is on both the front page and the telemetry page.
- `show_reports` controls the "Monthly Reports" drop-down, not a link.

## Testing

`python -m zensical build` reports no issues. In the built HTML the page renders
and all four of its links resolve: `../extras/`, `../imagegenerator/`,
`../../../custom/cheetah-generator/#aggregation-periods` and
`.../#helper-functions` -- the last two against anchors that are present in the
target. The link warnings elsewhere in the docs are unchanged by this commit.

### DIFF
--- a/docs_src/reference/skin-options/displayoptions.md
+++ b/docs_src/reference/skin-options/displayoptions.md
@@ -1,0 +1,138 @@
+# [DisplayOptions]
+
+This section holds options that control what a skin displays, and in what
+order. Like [`[Extras]`](extras.md), the options are not interpreted by WeeWX
+itself. The section is made available to the templates as the tag
+`$DisplayOptions`, and what to do with it is up to the skin.
+
+The *Seasons* skin uses it to decide which observation types, plots, and
+sections appear on its pages. The options below are the ones *Seasons*
+recognizes. Your own skin is free to use others.
+
+The templates read each option with a fallback, so removing one does not cause
+an error: the template's own default is used instead. Those defaults are much
+shorter than the lists that ship in `skin.conf`, and are given for each option
+below.
+
+!!! Note
+    Naming an observation type does not guarantee that it will appear. The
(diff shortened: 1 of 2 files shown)


tkeffer:
This issue has been hanging around for a long time. Thanks for fixing it!

PR #1115, opened 2026-08-24
Say so when a Vantage logger's memory is corrupt

### LINKED ISSUE #1105: Better error diagnostic for corrupt Vantage logger memory
Right now it takes an expert to diagnose a logger with corrupt memory.

The idea is to calculate how many records to expect to download, then decrement as each record is processed. If the counter is non-zero when exiting the loop the memory must be corrupt. Log the problem, along with a link to the fix.

hilman2:
Closes #1105. Against `master`, alongside `314882d6` ("Better error message when unable
to determine hardware type"), which is the same kind of change.

**One question, on the threshold.** The issue says "if the counter is non-zero when
exiting the loop the memory must be corrupt". Taken literally, that fires on healthy
stations. The normal end of a download is also a declining timestamp: the memory wraps
around inside the final page, so up to five of the promised records are never sent. I
therefore log only when more than a page is still outstanding. That is precisely when
the console described pages it could not fill.

## What it does

The console says up front how many records it is about to send, i.e. `_npages * 5 -
_start_index`. The loop now counts them and compares on the way out. When the count
falls short by more than a page:

```
Logger promised 224 archive records, but returned 0 before its timestamps went backwards.
This is what corrupt logger memory looks like. For how to fix it, see
https://github.com/weewx/weewx/wiki/Troubleshooting-the-Davis-Vantage-station#corrupt-station-memory
```

The link goes to the wiki because the fix is several steps, and it is already written up
there: pull the power and batteries, else `weectl device --dump` then `--clear-memory`,
else replace the logger or fall back to `record_generation = software`.

The other early exit, on an unwritten `0xff` record, is deliberately left alone. That is
a freshly cleared logger. Warning there would send someone chasing a fix they had just
applied.

## Testing

No hardware needed. `src/weewx/drivers/tests/test_vantage.py` puts a fake port in front
of the driver and feeds it canned DMPAFT replies and pages. The failure reproduces from
a checkout:

| case | result |
|---|---|
| Two full pages, ten records | all ten returned, nothing logged |
| Wrap inside the final page | seven returned, nothing logged |
| 45 pages promised, first record already stale | nothing returned, message logged with the link |
| Just-cleared logger, `0xff` records | two returned, nothing logged |

The third case is the log from the wiki page, in the same shape. `Retrieving 45 page(s);
starting index= 1`, followed straight away by a timestamp from two weeks earlier.

Both mutations of the check fail the suite, so the tests are not decorative:

- threshold `> 0` instead of `> 5`: the wrap-around case fails, i.e. the false alarm
  described above
- check removed: the corruption case fails

Full suite: 378 passed, no failures. That is 374 before, plus these four.
`vermin --target=3.7` reports the same minimum as before.

There is no precedent for a URL in a log message anywhere in `src/`. Should the message
name the wiki page without the URL instead?

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -3,6 +3,10 @@ WeeWX change history
 
 ### 5.5.n n-Month-2026
 
+Say so in the log when a Vantage logger returns far fewer archive records than it
+said it would, which is what corrupt logger memory looks like, and link to the fix.
+Fixes [Issue #1105](https://github.com/weewx/weewx/issues/1105).
+
 Saving the configuration file no longer changes its mode or ownership. Under a
 package installation, `weectl extension install`, `weectl extension uninstall`,
 `weectl station reconfigure` and `weectl station upgrade` were leaving
--- a/src/weewx/drivers/tests/test_vantage.py
+++ b/src/weewx/drivers/tests/test_vantage.py
@@ -1,0 +1,196 @@
+#
+#    Copyright (c) 2026 the WeeWX contributors
+#
+#    See the file LICENSE.txt for your full rights.
+#
+"""Test how the Vantage driver downloads archive records.
+
+These tests stand a fake serial port in front of the driver, so they need no
+hardware. What they are mostly about is telling apart the two reasons the
+download loop can stop early: reaching the end of the data, which is normal, and
+a logger whose memory has gone bad, which is not.
+"""
+import struct
+import time
+import unittest
+from unittest.mock import patch
+
+import weewx.drivers.vantage
+from weewx.drivers.vantage import Vantage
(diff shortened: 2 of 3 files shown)


tkeffer:
It would be nice to test this on a "real" instance of the Vantage logger, but as hard as I tried, I could not corrupt my logger.

Eventually, some user will stumble across it.

I think the hardwired URL is fine. That page hasn't moved in 6 years.

vinceskahan:
This is really a great addition.  You might want to put a note at the top of the wiki page mentioning that it shouldn't be renamed or moved and mention the Vantage driver references the current wiki location by url.

Issue #1051, opened 2025-12-23
write up how to write an extension


matthewwall:
document extra arguments feature

hilman2:
This went stale this morning, so a note on what is actually left in it.

The body — the extra arguments feature — you documented in
[`05a0b18`](https://github.com/weewx/weewx/commit/05a0b18cfb9cf1e9f3c16d7bb0146b6ca0c9759a)
in February. The section `Passing arguments on to your installer` is in the docs today.

What is still open is @tkeffer's
[comment above](https://github.com/weewx/weewx/issues/1051#issuecomment-3868445526):
extensions supplying their own translations to a skin. That one really is undocumented —
the recursive `lang` search is in the 5.3.0 change log, but there is nothing about it in
`docs_src` on either branch, nor in the wiki, nor in the published 5.5 docs. An
extension author has no way to find it, and the alternative they will reach for is
editing the skin's own `de.conf`, which the next upgrade takes away.

#1116 adds a section to `custom/extensions.md` covering it, using the `weewx-climate`
layout from that comment. If it lands, this issue is done.

PR #1116, opened 2026-08-24
Document how an extension adds translations to a skin

hilman2:
Covers what is left of #1051. Against `master`, where the other documentation commits
are. One new section in `custom/extensions.md`, nothing else touched.

**On the issue itself:** the body asked for the extra arguments feature, and you
documented that in `05a0b18` back in February. What is still outstanding is the
[comment you added on 8-Feb](https://github.com/weewx/weewx/issues/1051#issuecomment-3868445526),
i.e. extensions supplying their own translations to a skin. This PR is that part, so the
issue can be closed once it lands. It went stale this morning, which is what prompted me
to look.

## Why it needs saying

The recursive search of a skin's `lang` directory arrived in 5.3.0 and is in the change
log, but appears nowhere in the documentation. I checked `docs_src` on both branches,
the wiki, and the published 5.5 docs. An extension author has no way to find it.

The alternative they will reach for instead is editing the skin's own `de.conf`, which
is lost on the next upgrade. That is the mistake the feature exists to prevent, so it
seems worth a section rather than a change log line.

## What it says

Where the translations go, using your `weewx-climate` layout as the example, and four
things that are not obvious from the outside. I checked each against
`reportengine.get_lang_dict()` rather than reading the source and guessing:

| | |
|---|---|
| Subdirectory name must **begin** with `lang` | `lang-climate` found, `climate-lang` not |
| Several extensions at once | each in its own subdirectory, all merged |
| Extension and skin define the same phrase | the extension wins |
| Extension ships `de.conf` but no `th.conf` | `lang=th` shows the phrase untranslated, no error |

The third one surprised me, and the section says plainly that it is best avoided.

A note records that this needs 5.3 or later, so an author knows what they are asking of
their users.

`python -m zensical build` reports no issues, and the link to *Localization* resolves in
the built HTML.

### DIFF
--- a/docs_src/custom/extensions.md
+++ b/docs_src/custom/extensions.md
@@ -113,6 +113,57 @@ Once an extension has been packaged, it can be installed using `weectl`:
 
     weectl extension install EXTENSION-LOCATION
 
+## Adding translations to someone else's skin
+
+An extension that adds a widget to an existing skin brings text of its own. Do
+not put the translations in the skin's own language files: those edits are lost
+the next time the user upgrades WeeWX or reinstalls the skin.
+
+Instead, put the translations in a subdirectory of the skin's `lang` directory.
+Any subdirectory whose name starts with `lang` is searched, and what it holds
+is merged into the language the report is being run in. So an extension called
+`climate`, which adds a widget to the *Seasons* skin, would install
+
+```
+skins
+└── Seasons
+    └── lang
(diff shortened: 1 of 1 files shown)


tkeffer:
Uploaded to https://www.weewx.com/docs/5.5/custom/extensions/#adding-translations-to-someone-elses-skin

PR #1117, opened 2026-08-24
Stop using the datetime calls Python 3.12 deprecates

### LINKED ISSUE #1052: make fousb driver work with python 12

hilman2:
Closes #1052. Against `master`, as this is a fix.

The FineOffset driver calls `datetime.utcfromtimestamp()` four times and
`datetime.utcnow()` once. Both are deprecated as of 3.12, and a grep over `src/`
finds them nowhere else. Nothing else in the driver trips 3.12 up: it compiles
with no warning, and the PyUSB 0.4 style calls it makes, i.e. `usb.busses()`,
`dev.deviceClass` and `devh.controlMsg()`, are all still present in PyUSB 1.3.1.

**This was tried once before and reverted.** `2d789431` replaced the four
`utcfromtimestamp()` calls with `fromtimestamp(x, timezone.utc)`. You reverted it
in `2202248b`, because it produced "can't subtract offset-naive and offset-aware
datetimes". The reason is in that diff: those four became aware, while
`datetime.min`, `datetime.max` and `utcnow()` in the same function stayed naive,
and `sync()` subtracts and compares across all of them.

This PR makes the rest aware as well. No datetime leaves the driver, i.e. loop
packets are rebuilt by `pywws2weewx()` with a `dateTime` of its own, and
`genArchiveRecords()` converts back to epoch seconds before yielding. So the
representation is the driver's to choose, and aware UTC is the one that cannot be
mixed up.

Two of the changes are more than a change of representation:

- The station clock comparison in `sync()` was subtracting a naive local datetime
  from a naive UTC one. It now subtracts two UTC values. What it computes is
  unchanged, because only `err.seconds % 60` is read and every zone offset is a
  whole number of minutes.
- `genArchiveRecords()` drops its hand-rolled epoch arithmetic, and with it the
  daylight saving FIXME beside it. `timestamp()` on an aware datetime does no
  local conversion.

`set_clock()` and the two `now()` calls in the configurator are unchanged. The
station keeps wall clock local time.

## Visible differences

Four places print a whole datetime and gain a `+00:00` suffix: the "get %d
records since" debug line, the invalid-data error in `get_records()`, the
"synchronised to %s for ptr" debug line at the end of `sync()`, and the
`weectl device --history` output. That matches what the docstrings say those
values are. The `%H:%M:%S` log lines are unchanged.

## Testing

`src/weewx/drivers/tests/test_fousb.py` covers the epoch conversion in
`genArchiveRecords()`, including the local hour that comes round twice when
daylight saving ends, and the estimate `sync()` arrives at over both bounds.
Making the bounds naive again fails with "can't compare offset-naive and
offset-aware datetimes", i.e. the 2023 error.

Full suite: 389 passed on 3.7 and on 3.14, no failures. `vermin --target=3.7`
reports the same minimum as before.

I have no FineOffset hardware. `live_data()` is stubbed in the sync test, so what
it reads off the USB device is not exercised.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -13,6 +13,11 @@ Say so in the log when a Vantage logger returns far fewer archive records than i
 said it would, which is what corrupt logger memory looks like, and link to the fix.
 Fixes [Issue #1105](https://github.com/weewx/weewx/issues/1105).
 
+The FineOffset USB driver no longer calls `datetime.utcnow()` or
+`datetime.utcfromtimestamp()`, which Python 3.12 deprecates. Its datetimes are now
+aware UTC. Fixes [Issue #1052](https://github.com/weewx/weewx/issues/1052).
+[PR #1117](https://github.com/weewx/weewx/pull/1117).
+
 Saving the configuration file no longer changes its mode or ownership. Under a
 package installation, `weectl extension install`, `weectl extension uninstall`,
 `weectl station reconfigure` and `weectl station upgrade` were leaving
--- a/src/weewx/drivers/fousb.py
+++ b/src/weewx/drivers/fousb.py
@@ -1,4 +1,4 @@
-# Copyright 2012-2024 Matthew Wall
+# Copyright 2012-2026 Matthew Wall
 # See the file LICENSE.txt for your full rights.
 #
 # Thanks to Jim Easterbrook for pywws.  This implementation includes
@@ -226,15 +226,18 @@ import weewx.wxformulas
 log = logging.getLogger(__name__)
 
 DRIVER_NAME = 'FineOffsetUSB'
-DRIVER_VERSION = '1.3'
+DRIVER_VERSION = '1.4'
 
-# Every datetime in this driver is aware UTC. None of them leave it: loop packets are
-# rebuilt by pywws2weewx() with a dateTime of its own, and genArchiveRecords() converts
-# back to epoch seconds before yielding. So the representation is the driver's to
-# choose, and the one that carries its time zone is the one that cannot be mixed up:
-# sync() compares and subtracts the two bounds below against times read from the
-# station, and doing that across a naive and an aware value raises "can't subtract
-# offset-naive and offset-aware datetimes".
+# Every datetime that carries an observation time is aware UTC. None of them leave the
+# driver: loop packets are rebuilt by pywws2weewx() with a dateTime of its own, and
+# genArchiveRecords() converts back to epoch seconds before yielding. So the
+# representation is the driver's to choose, and the one that carries its time zone is
+# the one that cannot be mixed up: sync() compares and subtracts the two bounds below
+# against times read from the station, and doing that across a naive and an aware value
+# raises "can't subtract offset-naive and offset-aware datetimes".
+#
+# The station's own clock is local wall clock time, so set_clock() and the configurator
+# keep naive local datetimes on purpose.
 UTC = datetime.timezone.utc
 DT_MIN = datetime.datetime.min.replace(tzinfo=UTC)
 DT_MAX = datetime.datetime.max.replace(tzinfo=UTC)
--- a/src/weewx/drivers/fousb.py
+++ b/src/weewx/drivers/fousb.py
@@ -228,6 +228,17 @@ log = logging.getLogger(__name__)
 DRIVER_NAME = 'FineOffsetUSB'
 DRIVER_VERSION = '1.3'
 
+# Every datetime in this driver is aware UTC. None of them leave it: loop packets are
+# rebuilt by pywws2weewx() with a dateTime of its own, and genArchiveRecords() converts
+# back to epoch seconds before yielding. So the representation is the driver's to
+# choose, and the one that carries its time zone is the one that cannot be mixed up:
+# sync() compares and subtracts the two bounds below against times read from the
+# station, and doing that across a naive and an aware value raises "can't subtract
+# offset-naive and offset-aware datetimes".
+UTC = datetime.timezone.utc
+DT_MIN = datetime.datetime.min.replace(tzinfo=UTC)
+DT_MAX = datetime.datetime.max.replace(tzinfo=UTC)
+
 def loader(config_dict, engine):
     return FineOffsetUSB(**config_dict[DRIVER_NAME])
 
@@ -1093,11 +1104,8 @@ class FineOffsetUSB(weewx.drivers.AbstractDevice):
         """
         records = self.get_records(since_ts)
         log.debug('found %d archive records' % len(records))
-        epoch = datetime.datetime.utcfromtimestamp(0)
         for r in records:
-            delta = r['datetime'] - epoch
-            # FIXME: deal with daylight saving corner case
-            ts = delta.days * 86400 + delta.seconds
+            ts = int(r['datetime'].timestamp())
             data = pywws2weewx(r['data'], ts,
                                self._last_rain_arc, self._last_rain_ts_arc,
                                self.max_rain_rate)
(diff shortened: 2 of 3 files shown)


chaunceygardiner:
Thanks for this, and for writing down why the 2023 attempt failed.  The direction is right, and the test is a welcome addition.  Two things before it can go in, and one suggestion that I think makes it a better change.

**It needs a rebase.**  `docs_src/changes.md` conflicts with master now that the #1105 entry occupies the same spot at the top of 5.5.n.

**The docstring's reason for staying naive is not right.**  `datetime.min` and `datetime.max` can carry a time zone: `datetime.max.replace(tzinfo=timezone.utc)` works, and it compares and subtracts like any other aware datetime.  2d789431 broke because it made four values aware and left `utcnow()` and the two bounds naive, not because the bounds could not be made aware.  The helper docstring, the test module docstring and the PR description all repeat the claim, so a future reader would be sent down the wrong road.

**Suggestion: go aware instead.**  Nothing in the driver lets a datetime escape.  Loop packets are rebuilt by `pywws2weewx` with a `dateTime` from `time.time()`, and `genArchiveRecords` converts back to epoch seconds before yielding.  So the driver is free to pick its representation, and aware UTC is the one that cannot be mixed up.  It is no bigger than what you have:

```python
UTC = datetime.timezone.utc
DT_MIN = datetime.datetime.min.replace(tzinfo=UTC)
DT_MAX = datetime.datetime.max.replace(tzinfo=UTC)
```

and then, in the four methods you already touch:

```python
# get_records
dt = datetime.datetime.fromtimestamp(since_ts, UTC)
dt = DT_MIN

# sync
range_hi = DT_MAX
range_lo = DT_MIN
prev_date = DT_MIN
prev_date = datetime.datetime.now(UTC)
err = last_date - datetime.datetime.fromtimestamp(self._station_clock, UTC)

# live_data, both places
result['idx'] = datetime.datetime.fromtimestamp(int(next_live), UTC)
result['idx'] = datetime.datetime.fromtimestamp(int(next_log), UTC)

# genArchiveRecords: the epoch, the delta arithmetic and the FIXME all go
ts = int(r['datetime'].timestamp())
```

Two places come out better than equivalent.  The `_station_clock` line in `sync()` today subtracts a naive local datetime from a naive UTC one, and only gets away with it because the result is taken modulo 60 seconds.  And `genArchiveRecords` can drop its hand-rolled epoch arithmetic along with the daylight-saving FIXME beside it, since an aware `timestamp()` does no local conversion.

`set_clock()` and the two `now()` calls in the configurator should stay as they are.  The station keeps wall-clock local time, and none of those calls is deprecated.

Visible differences: the `%H:%M:%S` log lines are unchanged.  The three places that print a whole datetime, which are the "get %d records since" debug line, the invalid-data error in `get_records`, and the `weectl device --history` output, gain a `+00:00` suffix.  That matches what the docstrings already say they are.

The test would shrink to match: the naive-ness assertions go, and a round trip through `fromtimestamp(ts, UTC).timestamp()` for `genArchiveRecords` would be worth adding in their place.  I have no FineOffset hardware either, but the failure mode here is loud: a missed spot raises `TypeError` at the first comparison in `sync()` rather than drifting silently.

hilman2:
Rebased, and gone aware. You are right about the bounds: the claim that they
cannot carry a time zone was wrong, and it is out of the docstring, the test
module and the PR description.

Two notes on your list. The `_station_clock` subtraction changes what it means but
not what it computes, because only `err.seconds % 60` is read and every zone
offset is a whole number of minutes. And there is a fourth place that gains
`+00:00`: the "synchronised to %s for ptr" debug line at the end of `sync()`.

The test drives `sync()` with a stubbed `live_data()`, so the comparisons against
both bounds actually run, and covers the epoch conversion under
`TZ=America/Los_Angeles` across the hour that repeats when daylight saving ends.
Making either side naive again gives the 2023 error.

chaunceygardiner:
This is exactly it.  Thanks for taking the suggestion, and for finding the fourth `+00:00` site that I missed.

I checked it over rather than taking it on trust.  It rebases clean, the new tests pass warning-free, and the full suite is unchanged apart from the MySQL cases this machine cannot run.  Two things I confirmed:

- Reverting either bound to naive fails a test with the 2023 error, so these are real regression tests rather than assertions that cannot fail.
- The daylight saving pair really does straddle the repeated hour.  Those two timestamps render as 01:30 PDT and 01:30 PST, the same wall clock an hour apart, which is the case the deleted FIXME named.

Your note on `err.seconds % 60` is right, and for a slightly better reason than you give.  The change lets `err` go negative for the first time, and Python normalizes a negative timedelta into a large positive `.seconds`.  It still comes out the same because 60 divides 86400, so `.seconds % 60` equals `total_seconds % 60` for either sign.  Worth half a sentence in the commit message if you think a future reader would wonder.

One change before this goes in.

**The new module comment overstates, and the overstatement is a trap.**  It opens "Every datetime in this driver is aware UTC", but five naive local ones remain, in `set_clock()` and in the configurator.  Those are naive deliberately: `set_clock()` BCD-encodes `now.hour` and `now.minute` straight into the station's registers, and `get_clock()` reads them back through `time.mktime()`, so the station keeps local wall clock time.  Someone who believed that sentence and made `set_clock()` consistent with it would set every FineOffset station to UTC.  Scoping the claim is enough:

```python
# Every datetime that carries an observation time is aware UTC. None of them leave the
# driver: loop packets are rebuilt by pywws2weewx() with a dateTime of its own, and
# genArchiveRecords() converts back to epoch seconds before yielding. So the
# representation is the driver's to choose, and the one that carries its time zone is
# the one that cannot be mixed up: sync() compares and subtracts the two bounds below
# against times read from the station, and doing that across a naive and an aware value
# raises "can't subtract offset-naive and offset-aware datetimes".
#
# The station's own clock is local wall clock time, so set_clock() and the configurator
# keep naive local datetimes on purpose.
```

Two small ones while you are in there:

- The file header still reads `Copyright 2012-2024`.  Extend it to 2026; this is a substantive edit.
- Bump `DRIVER_VERSION` to `1.4`.

hilman2:
Scoped, in 9fb5a0dd. You are right that the sentence was a trap: `set_clock()`
BCD-encodes `now.hour` into the registers and `get_clock()` reads them back through
`time.mktime()`, so anything that made it consistent would set every station to UTC.
Copyright year and `DRIVER_VERSION` are in the same commit.

One correction to the sign argument, which does not change the conclusion. `err` could
already go negative before this change, anywhere east of UTC: naive UTC minus naive
local is -2h plus the skew in CET, i.e. `days=-1, seconds=79200`. West of UTC it was
positive. What is new is that the sign follows the actual clock skew rather than the
zone offset. The rest holds, and 60 dividing 86400 is why the result never changed. It
is in the commit message.

chaunceygardiner:
You are right, and my claim was wrong.  `err` could already go negative east of
UTC: naive UTC minus naive local is the zone offset with its sign flipped, so it
was negative in CET before this change as much as after.  I ran it to be sure --
Berlin comes out at `-1 day, 21:59:59` and Los Angeles at `6:59:59`.  Your framing
is the better one: what is new is that the sign follows the actual clock skew
rather than the zone offset.  Good that it is in the commit message.

The scoping reads right.  I checked the rest of it as well: five naive local
datetimes remain, all of them in `set_clock()` and the configurator, and
everything that carries an observation time is aware UTC.  After this lands there
is no `utcnow()` or `utcfromtimestamp()` left anywhere in `src/`.  The tests pass
warning-free under `-W error::DeprecationWarning`, and it merges clean onto master.

Copyright year and `DRIVER_VERSION` are both in.  Merging this.  Thanks for the
care you put into it, and for correcting me where I had it wrong.

Issue #1050, opened 2025-12-23
build suite of system tests using vagrant


matthewwall:
automate install/upgrade tests using vagrant

vinceskahan:
Verifying systemd setup and perhaps udev rules would be a couple things this would help with if you wanted to test across a variety of os+version combinations.  Maybe python version differences as well.

matthewwall:
this is the list of tests, to be run on each operating system:

1) install latest weewx, verify:
  - simulator runs
  - reports generated
  - permissions ok (weewx user can modify weewx.conf)
  - start/stop weewx
  - install an extension
  - remove an extension
 
2) upgrade to latest weewx, verify:
  - simulator runs
  - permissions ok (weewx user can modify weewx.conf)
  - start/stop weewx
  - changes to weewx.conf are not obliterated
  - changes to skin are not obliterated
  - previous extension(s) respected
  - install an extension
  - remove an extension
 
there are many more tests that could be added, but it would be nice to have a basic suite of tests to run automatically when one generates deb/rpm packages for each platform.

hilman2:
I had a go at @matthewwall's list, in containers rather than Vagrant guests. It found a
real bug on the first run. Sketch here:
[hilman2/weewx@system-test-spike](https://github.com/hilman2/weewx/tree/system-test-spike/dev/system-tests).

**Containers or VMs?** Most of the list needs neither. Installing the `.deb`,
permissions, the simulator and the extension cycle all work in a plain container. Those
could run in the CI that already exists, on every push. Only `systemd` start and stop
needs a VM, or a privileged systemd container.

`check-install.sh` holds the checks and knows nothing about how the machine was made. It
runs unchanged in a container, a Vagrant guest, or on real hardware. `run-docker.sh`
builds the `.deb` from a checkout, installs it in a clean `debian:12` and runs them. Two
minutes, both containers removed afterwards.

```
== checks ==
== package ==
  ok    package is installed
  ok    weewxd on PATH
== user and permissions ==
  ok    weewx user can write /etc/weewx/weewx.conf
  ok    /etc/weewx/weewx.conf is not world readable (660)
  ok    /etc/weewx owned by weewx
== operating system integration ==
  ok    udev rules installed
  skip  systemd unit installed (postinst only installs it under systemd)
== the simulator produces data ==
  ok    archive records written (2)
  ok    reports generated (5 html files)
  ok    plots generated (60 images)
  ok    no critical errors in the log
== extensions ==
  ok    extension installs
[...]

PR #1118, opened 2026-08-24
Keep the configuration file's mode and ownership when saving it

hilman2:
Against `master`, as this is a fix. Found while looking at #1050. The first container
test I wrote for it tripped over this immediately.

## What happens

`weecfg.save()` with `backup=True` moves the old file aside and copies a new one into
place. A copy is a new file. So it takes its mode from the caller's umask, and its
ownership from whoever is running, i.e. root for anything invoked with `sudo`.

The Debian `postinst` deliberately sets `/etc/weewx/weewx.conf` to `weewx:weewx 0660`,
and says why:

```sh
# ensure that the configuration files are not world-readable so that
# we protect any passwords and tokens
chmod 660 /etc/weewx/*.conf*
```

That is undone by `weectl extension install`, `weectl extension uninstall`,
`weectl station reconfigure` and `weectl station upgrade`. All four save with a backup.
Measured in a Debian 12 container, installing the `.deb` built from this tree:

|  | mode and owner |
|---|---|
| fresh install | `-rw-rw---- weewx:weewx` |
| after `extension install` | `-rw-r--r-- root:weewx` |
| after `extension uninstall` | `-rw-r--r-- root:weewx` |
| after `station reconfigure` | `-rw-r--r-- root:weewx` |

With this commit all four stay `-rw-rw---- weewx:weewx`.

## What follows from it

`weewxd` only reads that file, never writes it, so nothing stops working. That is
presumably why this has gone unnoticed. Two things do change, and the packaging sets out
to prevent both:

- The `weewx` user can no longer edit its own configuration. That is what the `0660` and
  the setgid directories are for.
- `weewx.conf` becomes readable by every local account. It holds FTP, RESTful and
  database passwords. That is the case the comment above is explicitly guarding.

A pip installation is unaffected. Everything already belongs to the user running, so the
`chown` is a no-op and the umask gives the same mode back.

## The fix

Note the mode and ownership before the move, put them back after the copy. A
`PermissionError` from `chown` is ignored, since not running as root means the file is
already ours. The no-backup branch never had the problem: it
writes through the existing file, which keeps both.

## Testing

Two unit tests in `test_config.py`, so this is covered without a container. Removing the
`os.chmod` fails the first one.

Verified end to end as above. The `.deb` was built from this tree and installed in a
clean Debian 12 container, the four commands run in turn, `stat` after each.

Full suite: 376 passed, no failures.

### DIFF
--- a/src/weecfg/__init__.py
+++ b/src/weecfg/__init__.py
@@ -10,6 +10,7 @@ import importlib
 import os.path
 import pkgutil
 import shutil
+import stat
 import sys
 import tempfile
 
@@ -224,6 +225,13 @@ def save(config_dict, config_path, backup=False):
     # Check to see if the file exists, and we are supposed to make a backup:
     if os.path.exists(config_path) and backup:
 
+        # Note how the file looks before it is moved aside. The copy further down
+        # creates a new file, which would otherwise take its mode from the caller's
+        # umask and its ownership from whoever is running. After a package install
+        # that means root:root 0644 in place of weewx:weewx 0660, and the 'weewx'
+        # user can no longer edit its own configuration.
+        old_stat = os.stat(config_path)
+
         # Yes. We'll have to back it up.
         backup_path = weeutil.weeutil.move_with_timestamp(config_path)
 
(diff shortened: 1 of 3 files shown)


src/weecfg/__init__.py
```diff
@@ -236,6 +244,15 @@ def save(config_dict, config_path, backup=False):
             # Now move the temporary file into the proper place:
             shutil.copyfile(tmpfile.name, config_path)
 
+        # Then put the old mode and ownership back.
+        os.chmod(config_path, stat.S_IMODE(old_stat.st_mode))
+        if hasattr(os, 'chown'):
+            try:
+                os.chown(config_path, old_stat.st_uid, old_stat.st_gid)
```
tkeffer: Any reason why `os.chown()` would not exist?

hilman2: Only Windows, where it does not exist. WeeWX does not run there, so the guard is dead code. Taken out in 5b0fa1c2.

PR #1119, opened 2026-08-24
Document group `[Accumulator]`

hilman2:
Against `master`, where the other documentation commits are. One new page, one nav line.

`[Accumulator]` is covered by [the wiki](https://github.com/weewx/weewx/wiki/Accumulators),
but nothing about it is in the Reference Guide, which is where someone reading up on a
`weewx.conf` section looks first. The page is a summary of the options and points at the
wiki for the rest.

It surfaces on weewx-user regularly. Two from the last few weeks:

- ["Is there any where a definitive and correct set of documents that discuss how to use
  deltas and accumulators and their formats?"](https://www.mail-archive.com/weewx-user@googlegroups.com/msg54490.html)
  The asker reports finding "multiple solutions which do not seem to be solutions".
- Someone [trying `extractor = lasttime`](https://www.mail-archive.com/weewx-user@googlegroups.com/msg54612.html)
  to get the time of a lightning strike, and getting a `KeyError`.

The second is worth a note of its own, which the page carries. An unrecognized name
raises a `KeyError` naming it, rather than saying what is wrong. It is not raised at
startup either, but the first time the type appears in a packet, or at the end of the
first archive interval. There is very little to work back from.

## What is on it

The four options, i.e. `accumulator`, `adder`, `merger` and `extractor`, with the values
each one takes. The lists come from `ACCUM_TYPES`, `ADD_FUNCTIONS`, `MERGE_FUNCTIONS`
and `EXTRACT_FUNCTIONS`, not from the defaults at the top of the file. So they are
complete.

Then a paragraph on how to choose an extractor, since that is the one people actually
need. Readings taken at a moment average. Things counted over the interval sum. Running
totals the hardware maintains take `last`.

Placed after `[StdArchive]` in the nav, which is where the records it fills come from.

Each claim was checked against the code. `merge_avg` really does keep the lowest minimum
and use the average as the maximum. `extract_sum` returns `None` rather than zero when
nothing arrived. `firstlast` is the only accumulator that will take strings.
`record_generation` decides whether the accumulator is asked for a record at all.

`python -m zensical build` reports no issues, and both links resolve in the built HTML.

No change log entry, since documentation commits here do not get one.

### DIFF
--- a/docs_src/reference/weewx-options/accumulator.md
+++ b/docs_src/reference/weewx-options/accumulator.md
@@ -1,0 +1,111 @@
+# [Accumulator]
+
+As LOOP packets arrive, WeeWX holds the values for each observation type in an
+*accumulator*. At the end of the archive interval the accumulator is asked for
+one value per type, and those values become the archive record.
+
+How that happens is settable per observation type. The defaults suit the types
+WeeWX knows about: temperatures are averaged over the interval, rain is
+totalled, `txBatteryStatus` reports its last reading. A type WeeWX has not seen
+before, one that your driver or a service adds, is averaged. That is wrong for
+anything that counts or accumulates.
+
+This section is not in the distributed `weewx.conf`. Add it only for the types
+whose treatment you need to change:
+
+``` ini
+[Accumulator]
+    [[lightning_strike_count]]
+        extractor = sum
+    [[lightning_distance]]
+        extractor = min
+```
+
+Each subsection is named after an observation type and takes any of four
+options. Nothing else needs to be given: what you leave out keeps its default.
+
+!!! Note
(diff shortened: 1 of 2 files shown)


tkeffer:
Thanks for your contribution. Accumulators are documented [here](https://github.com/weewx/weewx/wiki/Accumulators).

It's a good thing to have a summary in the Reference Guide, but it should also point to the more definitive Wiki document.

hilman2:
Thanks, I missed that one. It is linked now, right above the tables, and the page reads
as a summary that sends you there.

Reading it turned up a mistake of mine as well. I had `merger = avg` down as a property
of the type. It goes with `loop_hilo = False`. Fixed in 37e9e47.

PR #1120, opened 2026-08-24
Read HTML_ROOT from the config when setting up the reporting directory

### LINKED ISSUE #1046: use HTML_ROOT to determine where to chown/chmod

hilman2:
Closes #1046. Against `master`, as this is a fix.

`setup_reporting_dir()` used a hardcoded path. A station that had moved `HTML_ROOT` got
the ownership set on `/var/www/html/weewx`, while the directory it actually writes to
was left alone. If that directory belonged to root, which it does after a plain
`sudo mkdir`, `weewxd` could not write its reports there.

Measured in a Debian 12 container. `HTML_ROOT` moved to `/srv/weather`, directory
created by root, then the package reinstalled:

```
before   Configuring reporting directory /var/www/html/weewx
         /srv/weather  root:root 755      weewx cannot write

after    Configuring reporting directory /srv/weather
         /srv/weather  weewx:weewx 2775   weewx can write
```

The default case is unchanged: with `HTML_ROOT = /var/www/html/weewx` the message and
the result are the same as before.

## The three details that make it work

`get_html_root()` reads the first `HTML_ROOT` from the config file. That is the one in
`[StdReport]`; later ones belong to individual reports.

- The option is indented. A pattern anchored on `^HTML_ROOT` finds nothing and silently
  falls back to the default, which looks like it works and does not.
- Commented-out examples such as `#HTML_ROOT = ...` sit in the shipped config and must
  not match.
- A trailing comment on the line has to be stripped.

A relative `HTML_ROOT` is relative to `WEEWX_ROOT`, which the packages do not use. Those
fall back to the default rather than creating a directory in `$PWD`.

Eight cases checked against the function, including all of the above, plus a missing
config file and a config with no `HTML_ROOT` at all.

Same change in `weewx.spec.in`. `sh -n` passes on `postinst`.

#1086 went at this in April and was withdrawn by its author the same day, without
review.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -6,7 +6,9 @@ WeeWX change history
 Package upgrades now read `HTML_ROOT` from `weewx.conf` before setting ownership on
 the reporting directory. A station that had moved `HTML_ROOT` was left with the
 directory owned by whoever created it, and `weewxd` could not write its reports there.
+A relative `HTML_ROOT` is resolved against `WEEWX_ROOT`.
 Fixes [Issue #1046](https://github.com/weewx/weewx/issues/1046).
+[PR #1120](https://github.com/weewx/weewx/pull/1120).
 
 Saving the configuration file no longer changes its mode or ownership. Under a
 package installation, `weectl extension install`, `weectl extension uninstall`,
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -3,6 +3,11 @@ WeeWX change history
 
 ### 5.5.n n-Month-2026
 
+Package upgrades now read `HTML_ROOT` from `weewx.conf` before setting ownership on
+the reporting directory. A station that had moved `HTML_ROOT` was left with the
+directory owned by whoever created it, and `weewxd` could not write its reports there.
+Fixes [Issue #1046](https://github.com/weewx/weewx/issues/1046).
+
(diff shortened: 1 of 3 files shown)


vinceskahan:
How do you know which HTML_ROOT to use if you have a scenario where multiple skins each have their own (optional) settings which might (or might not) be absolute or relative paths (or some combination) in addition to the system-wide default ?

For example - I specify HTML_ROOT for each skin, currently I use paths relative to WEEWX_ROOT but I've used both absolute and relative paths in years past.

WEEWX_ROOT = /home/pi/weewx-data
HTML_ROOT = public_html/vp2

and for each skin similarly
HTML_ROOT = public_html/vp2/Standard
HTML_ROOT = public_html/vp2/Seasons

and so on....

hilman2:
The first one, i.e. the one in `[StdReport]`. Reports below it are covered too, since `set_permissions` is recursive.

Relative paths were the gap, thanks. 46ae6a0b reads `WEEWX_ROOT` and resolves against it, following `weecfg`. Your example now comes out as `/home/pi/weewx-data/public_html/vp2`.

A report whose `HTML_ROOT` sits outside the top one is still not touched. `weewxd` creates that directory itself, in `cheetahgenerator.py`.

PR #1122, opened 2026-08-24
Restore sys.path after loading an extension installer

### LINKED ISSUE #1109: Use importlib instead of __import__

hilman2:
Closes #1109. Against `master`, as this is a fix. Found while working on that issue.

`old_path = sys.path` takes a reference, not a copy. The next line mutates that same list
with `insert()`, so the assignment in the `finally` block puts the mutated list back. The
installer directory stays on the path.

Measured against `master`:

```
sys.path before 6, after 7
directory still on path: True
after four extensions: 10 entries, 4 added
```

Those directories are temporary and get removed afterwards, so what is left behind are
dead entries that every subsequent import walks through.

Now done with the `add_path()` context manager, and with `importlib.import_module()` in
place of `__import__`. The latter makes the `sys.modules['install']` lookup that followed
the call unnecessary, i.e. the same shape you gave `weedb` in e129a14d. That was #1112,
which is closed in favour of this one, since both touched the same fifteen lines.

What stays is the `sys.modules.pop()`. Every extension calls its installer module
`install`, so leaving it in the cache hands the next extension installed by the same
process the previous one's installer. The comment there said only "Get rid of the
module", which does not say why it matters.

A second commit puts `import importlib` back in alphabetical order in `engine.py`.
3a7223de replaced `import sys` with it where it stood, which left it between `socket` and
`threading`.

## Tests

`get_extension_installer()` had none. There are three now: an ordinary load, a directory
with no `install.py`, and two extensions in a row. Reverting either half of the function
breaks them:

| reverted to                          | fails                                            |
|--------------------------------------|--------------------------------------------------|
| `old_path = sys.path`                | `test_get_installer`, `..._missing`              |
| `sys.modules.pop()` deleted          | `test_get_installer`, `..._two_extensions`       |

Full suite, one file per run the way the makefile does it. Debian 12, Python 3.13,
MariaDB, non-root:

|                 | passed | failed |
|-----------------|--------|--------|
| `origin/master` | 374    | 0      |
| this branch     | 377    | 0      |

`vermin --target=3.7` reports the same minimum before and after.

### DIFF
--- a/src/weecfg/__init__.py
+++ b/src/weecfg/__init__.py
@@ -12,6 +12,7 @@ import pkgutil
 import shutil
 import sys
 import tempfile
+from contextlib import contextmanager
 
 import configobj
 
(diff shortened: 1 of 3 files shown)


tkeffer:
Good spotting!

For these kinds of temporary reassignment of system resources I've come to prefer the `@contextmanager` pattern. Something like

```
from contextlib import contextmanager

@contextmanager
def add_path(new_path):
    old_path = list(sys.path)
    try:
        sys.path.insert(0, new_path)
        yield sys.path
    finally:
        sys.path = old_path
```

Also, while you're in there, the function `get_extension_installer()` should probably be updated to use `importlib` instead of `__import__`.

hilman2:
Both done. `add_path()` is your sketch, with a line in the docstring saying why the copy
matters.

The `importlib` change was #1112, opened before this one. I have closed it and moved its
tests here, since both were rewriting the same fifteen lines. Its second commit came
along too: `import importlib` sits between `socket` and `threading` in `engine.py`, where
3a7223de left it.

PR #1125, opened 2026-08-25
Add a listener for hardware that pushes

### LINKED ISSUE #1124: Provide a core listener for push-based hardware
**TL;DR:** WeeWX can only pull. Every driver for hardware that pushes has to build
its own HTTP server first. I would like to put that server in the core, without
adding any hardware support to it.

Is that something you would take? If so, I will write it.

Open questions are at the bottom.

## The problem

A growing share of consumer hardware never answers a request. It posts, on its own
schedule, and there is nothing to poll. Ecowitt gateways and consoles, Acurite
bridges, WeatherFlow, the Fine Offset ObserverIP line. The driver has to be a
server, so today each one writes a server.

`weewx-interceptor` opens its own `TCPServer` and `BaseHTTPRequestHandler`, Python 2
fallback imports included. `ecowittcustom` inherited that code verbatim when it
forked. `weatherflow-udp` opens its own socket. Each one solves the same list
again, and differently:

- thread ownership, and whether requests are served serially or concurrently
- shutdown, i.e. `server.shutdown()`, `server_close()`, joining the thread
- `allow_reuse_address`, or a port that stays bound after a restart
- bind address and IPv6
- reading `Content-Length` and capping the body, so one request cannot exhaust memory
- request timeouts, so a console holding a keep-alive connection does not pin a thread
- the response body, which the device requires before it treats the upload as delivered
- `X-Forwarded-*` handling, because the wiki recipe puts nginx in front of it
- a debug switch that logs the raw body

That last one is not a nicety. It is the first thing anyone needs when a new sensor
shows up and its fields are missing.

## The core already has the other half

`weewx.restx.RESTThread` gives an uploader its thread, its queue and its retry
handling. `AmbientThread` and `WOWThread` build on it, and so do extensions outside
the tree. Nobody writes an upload loop by hand.

There is no inbound equivalent, so everybody writes the receive loop by hand.

## The pattern exists already

`interceptor.py` is worth reading here, because it made this split years ago. It has
one `Consumer` class that owns the socket, the handler and the queue. Per device it
has a `Parser` and a `get_response()`, and `DEVICE_TYPES` maps a config value to the
right pair. Eight device types share one server.

The proposal is to lift that split one level, out of a single driver and into the
core, so that it is shared across drivers instead of across device types within one.

## Sketch

```python
from weewx.listener import HTTPListener

class MyDriver(weewx.drivers.AbstractDevice):

    def __init__(self, **stn_dict):
        self.listener = HTTPListener(**stn_dict)   # address, port, path, max_body

    def genLoopPackets(self):
        for request in self.listener:
            packet = self.parse(request.body)
            if packet:
                yield packet
```

The listener owns the socket, the thread, the bounded queue and the shutdown. It
binds to `weewx.SHUTDOWN` from #1106, so a driver no longer has to leak a thread to
get a clean exit. The response is a driver concern, so it comes from the driver,
either as a callable or a subclass.

What stays outside: parsing, field maps, unit assignment, hardware. The core gains
no device it would have to own.

## What it opens up

**Writing a receive driver becomes writing a parser.** Today the first 200 lines are
boilerplate that has nothing to do with the hardware. After this, a new driver is one
function from bytes to a dict, plus a response string.

**One listener, several drivers.** Only one process can hold port 80. Anyone running
an Ecowitt gateway and an Acurite bridge today needs two ports and two device
configurations. A shared listener can dispatch by path, e.g. `/ecowitt` and
`/acurite`.

**Driver and service from the same building block.** Several of these extensions want
to run as a service, augmenting another driver's loop packets rather than being the
source. Each one implements that separately today.

**Tests.** A listener in the core is tested in the core. Driver tests then become
parser tests, i.e. a captured POST body in, an expected packet out, no socket
involved. None of the four Ecowitt drivers in the wiki has tests today, and the
socket setup is a large part of the reason.

**One place for fixes.** A bug in request handling is fixed once, not in six forks
that all descend from the same 2016 code.

**A diagnostic that users can run.** With the listener in the core, `weectl device`
can gain a mode that prints incoming raw bodies. A user with an unsupported sensor
could then produce the dump the wiki asks for, without patching a driver first. That
is how new sensors get supported at all, and right now it needs a developer at both
ends.

## Why this helps the people who write these drivers

They stop maintaining infrastructure they never wanted to own. They inherit fixes
instead of forking them. Their config options become the same across drivers, so
`port` and `address` mean the same thing everywhere, and a user moving between two
receive drivers keeps their setup.

And it lowers the entry cost for the next one. The Ecowitt situation is what happens
when that cost is high: several partial forks, each with one maintainer, and no
obvious one to point a newcomer at.

## Open questions

- HTTP only, or a UDP variant too? WeatherFlow and Atmocom broadcast.
- `weewx/listener.py`, or under `weeutil`?
- How should the response reach the listener, i.e. callable, subclass, or config?
- Python 3.7 stays the floor, so `http.server` and `socketserver` only, no new
  dependencies. Is that the right constraint?

## Scope

This is not a driver, and not a protocol. Nothing existing has to change, and
`weewx-interceptor` could adopt it later without its configuration moving.

hilman2:
Adds `weewx.listener`, so a driver for hardware that pushes does not have to write
its own server. Closes #1124.

Submitting to `development`, as this is a feature.

`HTTPListener` and `UDPListener` share a `Listener` that owns the queue and the
iteration. A driver is left with the part that differs between devices:

``` python
from weewx.listener import HTTPListener

class MyDriver(weewx.drivers.AbstractDevice):

    def __init__(self, **stn_dict):
        self.listener = HTTPListener(**stn_dict)

    def genLoopPackets(self):
        for request in self.listener:
            packet = self.parse(request.text)
            if packet:
                yield packet

    def closePort(self):
        self.listener.close()
```

The listener owns the socket, the thread, a bounded queue and the shutdown. It binds
in the constructor, so a port already in use is reported where the driver is built.
It raises `WeeWxIOError` once it has stopped listening, because from a driver that is
otherwise indistinguishable from a station that went quiet.

What stays with the driver: parsing, field maps, units, and the response, which is
part of the protocol.

No new dependencies: `hmac`, `logging`, `queue`, `socket`, `threading`,
`urllib.parse`, `http.server`, and `to_int`/`to_bool` from `weeutil`.

46 tests in `test_listener.py`, `--target=3.7` clean, full suite 420 passed.
Documented under *Porting to new hardware*, with a change log entry.

One thing I left out. With `MetaDriver`, two receiving drivers can now run in one
process, and they need two ports. A shared listener dispatching by path would fix
that, and `path` is already there for it, but it wants process-wide state and I would
rather add it when somebody has the setup than guess at it now.

## First consumer

<https://github.com/hilman2/weewx-ecowitt>, which I wrote on top of this. It has been
running against an HP2561AE console since today, using the listener from this branch
rather than its bundled copy.

It is also where the encapsulation pays off. Its protocol, catalog and mapping import
without WeeWX, so its tests run from captured payloads with no socket and no engine,
and `python -m user.ecowitt` runs the listener directly to show what a station sends
before anything is configured.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -3,9 +3,9 @@ WeeWX change history
 
 ### 5.6.0 MM/DD/YYYY
 
-Added `weewx.listener`, a listener for hardware that pushes its data rather than
-answering a poll. Drivers for such hardware no longer have to write their own HTTP
-server. See [Issue #1124](https://github.com/weewx/weewx/issues/1124).
+Added `weewx.listener`, for hardware that pushes its data rather than answering a
+poll. Drivers for such hardware no longer have to write their own HTTP server or UDP
+socket. See [Issue #1124](https://github.com/weewx/weewx/issues/1124).
 
 
 ### 5.5.0 6-Aug-2026
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -1,6 +1,13 @@
 WeeWX change history
 --------------------
 
+### 5.6.0 MM/DD/YYYY
+
+Added `weewx.listener`, a listener for hardware that pushes its data rather than
+answering a poll. Drivers for such hardware no longer have to write their own HTTP
+server. See [Issue #1124](https://github.com/weewx/weewx/issues/1124).
+
+
 ### 5.5.0 6-Aug-2026
 
 Added the ability for services to bind to a `SHUTDOWN` event. This allows
--- a/docs_src/custom/drivers.md
+++ b/docs_src/custom/drivers.md
@@ -308,6 +308,27 @@ What the listener does not check is whatever the device carries in its own paylo
 as the Ecowitt `PASSKEY` or a Weather Underground station ID. Those belong to the
 protocol, so they are the driver's to verify.
 
+### Hardware that broadcasts
+
+Some hardware does not post at all. It puts a datagram on the local network and moves
+on, e.g. WeatherFlow on port 50222, or a Davis WeatherLink Live on 22222. Use
+`weewx.listener.UDPListener` for those. Same queue, same iteration, no response to send:
+
+``` python
+from weewx.listener import UDPListener
+
+self.listener = UDPListener(port=50222)
+```
+
+It takes `port`, `address`, `max_body`, `allowed_hosts`, `log_raw`, and `queue_size`
+with the same meanings, plus `reuse_address`. That one defaults to `True`, so other
+programs on the machine can read the same broadcasts. Set it to `False` to have the port
+to yourself.
+
+Two differences are worth knowing. UDP has no way to refuse an oversized datagram, so
+`max_body` truncates rather than rejects. And a datagram carries no path, no headers and
+no token, so `allowed_hosts` is the only filter there is.
+
 ## Define the configuration
 
 You then include a new section in the configuration file `weewx.conf` that
--- a/docs_src/custom/drivers.md
+++ b/docs_src/custom/drivers.md
@@ -240,8 +240,10 @@ straight through:
   <tr>
     <td class="code">path</td>
     <td><i>every path</i></td>
-    <td>Accept this path only, <i>e.g.</i>
-        <span class="code">/data/report/</span>. Anything else gets a 404.</td>
+    <td>Which paths to answer for, <i>e.g.</i>
+        <span class="code">/data/report/</span>. A comma-separated list, or a
+        Python list, is several. A callable is passed the path and returns whether
+        to accept it. Anything else gets a 404.</td>
   </tr>
   <tr>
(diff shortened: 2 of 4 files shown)


hilman2:
Pushed one more commit: `path` now takes a list, or a comma-separated string, and a
callable.

A driver that answers to a console whose path you choose and to one whose path is
burned into its firmware could not say so with a single string. The callable is for
when the set is not known at startup, e.g. a path per station where a station can be
added while WeeWX is running. `response` already takes one for the same kind of reason.

The check still happens before the request is built, so a path nobody wants costs a
lookup and a 404 rather than a parse. Five tests, and `drivers.md` documents both forms.

PR #1126, opened 2026-08-25
Do not lose an archive record when two periods end in one loop

hilman2:
`StdArchive` loses an archive record when the main loop runs past the end of two
periods without being broken.

The loop breaks `archive_delay` seconds after a period ends. A packet arriving
before that does not break it, so a second period can end while the first one's
accumulator is still in `old_accumulator`, waiting for `post_loop()`. The next
`OutOfSpan` replaces it, and its record is never written.

```python
(self.old_accumulator, self.accumulator) = \
    (self.accumulator, self._new_accumulator(event.packet['dateTime']))
```

The shape that does it is one packet per archive period, arriving just after the
boundary. A driver polling an API on the archive interval does that. Twelve such
packets write seven records on master and twelve with the patch.

`old_accumulator` becomes a queue. `post_loop()` works through it oldest first, so
records still reach the database in order.

Submitting PR to master, as this is a fix.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -5,6 +5,7 @@ WeeWX change history
 
 Fix loss of an archive record when the main loop runs past the end of two archive
 periods without being broken.
+[PR #1126](https://github.com/weewx/weewx/pull/1126).
 
 Say so in the log when a Vantage logger returns far fewer archive records than it
 said it would, which is what corrupt logger memory looks like, and link to the fix.
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -3,6 +3,9 @@ WeeWX change history
 
 ### 5.5.n n-Month-2026
 
+Fix loss of an archive record when the main loop runs past the end of two archive
+periods without being broken.
+
 Say so in the log when a Vantage logger returns far fewer archive records than it
 said it would, which is what corrupt logger memory looks like, and link to the fix.
 Fixes [Issue #1105](https://github.com/weewx/weewx/issues/1105).
--- a/src/weewx/engine.py
+++ b/src/weewx/engine.py
@@ -556,9 +556,6 @@ class StdArchive(StdService):
         # than one can be waiting, because the loop is not broken until archive_delay
         # after a period ends, and a packet arriving before that does not break it.
         self.accumulators = collections.deque()
-        # The accumulator whose record is being made, so that new_archive_record can
-        # augment it. None the rest of the time.
-        self.old_accumulator = None
 
         if self.record_generation == 'software':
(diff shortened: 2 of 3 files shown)


tkeffer:
There was a (probably not very good) underlying assumption that LOOP packets come more frequently than once every archive interval. No reason to assume that.

Question: the division between `self.accumulator`, `self.old_accumulator`, and `self.old_accumulators` seems like an artificial one. Why not implement a FIFO queue? Add a new accumulator at the top and it becomes the currently active one. Older accumulators get pushed down, then get processed in order.

hilman2:
You are right, a FIFO is cleaner. I kept the split because `old_accumulator` is
read by `new_archive_record` and `_software_catchup`, and this is a fix for master.

#1127 proposes dropping the accumulator altogether. Every LOOP packet goes into a
table of its own, and a record is worked out from the packets of its period. The
shuffling, `OutOfSpan` and both accumulator attributes go with it, and so does this
bug.

Would you rather see the FIFO here, or keep this one small?

tkeffer:
> Would you rather see the FIFO here, or keep this one small?

The FIFO change should be a small one.

hilman2:
Done in 9f88782f. A `deque`: the one at the front is the period being filled,
anything behind it has ended and is waiting. New periods go on the front, finished
ones come off the back, oldest first.

`self.accumulator` goes with it. Nothing outside `StdArchive` reads it, here or in
the skins.

29 insertions, 17 deletions now.

tkeffer:
Do we still need `self.old_accumulator`? Can we just check if the queue is longer than 2, then use the last entry?

Double-check! It's certainly possible I'm missing something!!

hilman2:
You were right, it is gone in 06a017bf. The accumulator stays at the back of the
queue until its record has been dispatched, so `new_archive_record` reads it from
there.

Double-checking turned up something else, fixed in 10f12994. A packet older than the
one before it was pushed onto the front of the queue, which made the period that was
really running look like a finished one. Its record went out early and short, and
the packets that belonged in it then found the record already written. The queue is
now ordered by period rather than by arrival.

That is not a corner case. It happens whenever a second source feeds the loop, and
it is measurable:

| | `development` | this PR |
|---|---|---|
| every 2-20 s, in order | 98.2% | 99.4% |
| one packet per archive period | 78.9% | 99.4% |
| timestamps that jump backwards | 45.5% | 74.4% |
| two sources, one 40 s behind | 72.5% | 96.9% |

Percentage of loop packets that reach an archive record. Method is in
https://github.com/weewx/weewx/issues/1127.

Two tests added, both fail without the fix. 391 pass.

Issue #1127, opened 2026-08-25
An archive record is the only copy of its readings


hilman2:
An archive record is the only copy of the readings that went into it. The LOOP
packets are gone once `StdArchive` has accumulated them. Three things follow:

* A reading that arrives after its period has ended cannot be put where it belongs.
* Every restart loses the period it happened in, up to a full archive interval.
* Nothing can be checked afterwards. A record built from 29 packets instead of 38
  looks exactly like one built from 38.

I would like to put every LOOP packet into a table of its own and work each record
out from the packets of its period. The accumulator, the shuffling, `OutOfSpan` and
`old_accumulator` all go. A late packet then only has to reach the table.

**Is this a direction you want?** #901 was yours, so I suspect it may be. I have it
working and would rather ask than send 2,000 lines unannounced.

Measured against `development`, counting rain that reaches the database: a driver
polling an API on the archive interval loses 20 % of its readings today. Two
sources, one 40 s behind, lose 31 %. Numbers below.

---

## What happens now

```python
except weewx.accum.OutOfSpan:
    (self.old_accumulator, self.accumulator) = \
        (self.accumulator, self._new_accumulator(event.packet['dateTime']))
```

A packet from an earlier period displaces the period being accumulated. That was
safe for as long as drivers stamped packets when they read the hardware. Vantage,
WMR100 and Ultimeter all do. Push hardware, MQTT and several sources in one instance
do not.

## Loss

Every packet carries one tip of the bucket, so the rain in the database counts the
packets that survived. 30,000 packets per row.

| packet stream | reaches the database |
|---|---|
| dense, 2-20 s, in order | 86.9 % |
| one packet per archive period | 79.5 % |
| with backward jumps | 43.4 % |
| two sources, one 40 s behind | 69.2 % |

Row four is what @roe-dl described in #1124.

## Every restart

The accumulator lives in memory, so a restart discards whatever it held. Stop weewx
at 12:03 with a five-minute interval and the readings from 12:00 to 12:03 are gone.
No warning, no gap in the record numbering, nothing in the log.

That is the cost of an upgrade, a configuration change, a reboot, an OOM kill. It
applies to every installation, not only to the ones with late data.

With the packets kept, the period is worked out on the way back up. Nothing to
configure.

## What the packets make visible

Live station, two hours of an Ecowitt console:

```
8 s   942 x  ########################################
9-16 s 18 x
> 20 s  1 x  (72 s, nine uploads that never arrived)
```

[...]

PR #1129, opened 2026-08-26
Split StdArchive into a record generator and a record store

hilman2:
From your comment in #1127: "By separating archive record generation from database
storage we could support this as an extension."

`StdArchiveGenerator` makes archive records out of LOOP packets. It emits
`NEW_ARCHIVE_RECORD`. `StdArchiveStore` listens for that and writes to the database.
They share nothing else. Replace either one and the other keeps working. A second
store on another binding saves the same records twice.

`StdArchive` stays, and runs both. A configuration that names it is unaffected.

Two things move:

- The accumulator travels on the event. The store needs it for the daily summary's
  high/lows. It can no longer reach into the generator for it. A record covering
  some other period, which a hardware catchup brings, travels without one.
- Augmentation happens before the event is dispatched, not on the way into the
  database. Every listener now sees the same record. Before, it depended on service
  order.

14 tests in `src/weewx/tests/test_archive_services.py`. They cover either half alone,
both together, the combined service, and a replacement generator driving the stock
store. Four mutations to the new code each fail exactly one test.

End to end, `weewxd` against the simulator, same start, once each way:

| | |
|---|---|
| records compared | 2235 |
| fields compared | 227970 |
| fields differing | 0 |
| daily summary tables | 100, none differing |
| replacement generator, stock store | 2632 records |

Full suite 388 passed on 3.12 and 3.13. `vermin` unchanged from `development`.

One question. Should `update_config.py` rewrite `archive_services` for existing
installations? They work as they are. But they do not get the split until someone
edits the line by hand.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -25,7 +25,7 @@ Removed the APRS "messaging-capable" packet flag from `restx.py`.
 
 ### 5.5.0 6-Aug-2026
 
-Split `StdArchive` into `StdArchiveGenerator`, which makes archive records, and
+Split `StdArchive` into `StdArchiveCreator`, which makes archive records, and
 `StdArchiveStore`, which saves them, so that either can be replaced on its own.
 `StdArchive` still runs both, so an existing configuration is unaffected.
 [PR #1129](https://github.com/weewx/weewx/issues/1129).
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -22,13 +22,12 @@ could no longer edit it, and the passwords in it became world-readable. [PR
 Removed the APRS "messaging-capable" packet flag from `restx.py`.
 [PR #1108](https://github.com/weewx/weewx/pull/1108), by `W0CHP`.
 
-
-### 5.5.0 6-Aug-2026
-
 Split `StdArchive` into `StdArchiveCreator`, which makes archive records, and
 `StdArchiveStore`, which saves them, so that either can be replaced on its own.
 `StdArchive` still runs both, so an existing configuration is unaffected.
 [PR #1129](https://github.com/weewx/weewx/issues/1129).
+
+### 5.5.0 6-Aug-2026
 
 Added the ability for services to bind to a `SHUTDOWN` event. This allows
 services to be notified when `weewxd` is being shutdown. 
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -25,6 +25,11 @@ Removed the APRS "messaging-capable" packet flag from `restx.py`.
 
 ### 5.5.0 6-Aug-2026
 
+Split `StdArchive` into `StdArchiveGenerator`, which makes archive records, and
+`StdArchiveStore`, which saves them, so that either can be replaced on its own.
+`StdArchive` still runs both, so an existing configuration is unaffected.
+[PR #1129](https://github.com/weewx/weewx/issues/1129).
+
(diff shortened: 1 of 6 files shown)


bellrichm:
This is exciting. If I understand correctly, to modify the archive record and get it persisted I'd have to insert my service between the two WeeWX services. In my prototype of 901, I tried two approaches. A new event to persist the data. And alternatively a new service stanza, such as persist_services. 
As an extension writer, I liked the new service stanza. It made it clear that these services should only persist the data. I just wasn't sure how intrusive Tom would want the update....
Of course with all that said, it's too late for me. I already have a working workaround.... :(

hilman2:
Yes, between the two. `archive_services` runs in order:

```
archive_services = weewx.engine.StdArchiveGenerator, user.mine.MyService, weewx.engine.StdArchiveStore
```

The record is augmented before the event goes out, so your service sees what the store
will save. That way it needs no new service list.

What does your workaround do? If this would replace it, that is worth knowing.

bellrichm:
Thanks! I know enough to get an extension installed. How hard is it to insert a service in a specific location in the service list?

bellrichm:
> What does your workaround do? If this would replace it, that is worth knowing.

Unfortunately I don't think so. The base of my problem was that my service was updating loop packets and did not "know" that a new archive period had started. This resulted in some data from a prior archive period "bleeding into" the next archive period. Tom had suggested using the existing accumulator functionality, I just couldn't figure out how...  so I now do the processing on the new_archive_record event. It has a check that the data belongs to the current archive period. It does assume that loop packets are generated in sequential order, but that is whole different topic.

tkeffer:
To answer your question: no, the install process should not rewrite `archive_services`. We used to do that (hence all the `update_to_vXX` functions in `weecfg.update_config.py`), but it proved very difficult to maintain. You also couldn't count on it running if you had two processes that used the same code base, but different configuration files. So now we depend on runtime compatibility.

tkeffer:
I appreciate your decision to share `[StdArchive]` between the two new services. It certainly makes upwards compatiblity simpler.

Question: suppose an extension replaces one of the two services and it needs additional options not found in `[StdArchive]`? How would that work? Look in a new stanza, plus `[StdArchive]`. 

That's not so bad. Just wondering if you've thought it through.

tkeffer:
The name `StdArchiveGenerator` could be confused with report generators. How about `StdArchiveCreator`?

Also, the entry in the change log got messed up. It should be under V5.6.0.

hilman2:
`StdArchiveCreator` it is, and the change log entry moved to 5.6.0. Rebased on
`development` while I was in there.

Runtime compatibility is what this relies on. `StdArchive` still runs both halves,
so an existing `archive_services` needs no rewriting.

On the extension question: it needs nothing from the framework. A replacement
calls up to the constructor it replaces, which reads `[StdArchive]`, and then
reads its own stanza for the rest. Three tests assert it now, including one that
inherits nothing at all and is served anyway, since the event is the whole
contract.

The shared options stay in `[StdArchive]` on purpose, and `data_binding` is the
one that matters. The creator asks the database for the last record before it
catches up. If a replacement store wrote somewhere else, that check would be
against the wrong database.

One thing the reading turned up. `StdArchive` did not pass `shutDown()` on. The
engine calls it on the services it loaded, and the two it holds are not among
them. Neither has anything to release today, so nothing showed. It delegates now.

PR #1128, opened 2026-08-26
A JSON generator, and a skin that draws its charts in the browser

hilman2:
Closes #1111. To `development`, as this adds generators and a skin.

Three parts. The first two stand on their own.

**`weewx/jsongenerator.py`.** It reads the same `[ImageGenerator]` sections, fetches
the same series through `weewx.xtypes.get_series()`, applies the same units and
labels, and writes JSON instead of drawing a PNG. Plots people already defined come
out as data, with no new configuration. It also writes an archive of the whole
record, split by calendar year on a fixed grid, so a client can ask for last March.

**`weewx/summaryimage.py`.** WeeWX can draw a time series but not the numbers, and a
picture of the current readings is what people paste into a forum post or a
signature. One PNG, redrawn each cycle, labels and units from the skin.

**`skins/Horizon`.** A skin that draws its charts in the browser from that JSON, and
steps back through the record in whole calendar units. uPlot is bundled, MIT, with
its licence beside it. No `ImageGenerator` in its `generator_list`.

Everything is additive. `[[HorizonReport]]` ships with `enable = false` and its own
`HTML_ROOT`, so an upgrade keeps the reports it had. The image generator and every
existing skin are untouched.

Documentation in `docs_src/custom/json-generator.md` and
`docs_src/custom/horizon-skin.md`, with `nav` entries in `zensical.toml`.
`src/weewx/tests/test_jsongenerator.py` covers the generator, 91 tests. Full suite is
467 passed on 3.12 and 3.13.

PR #1130, opened 2026-08-27
A climate page for the Horizon skin

hilman2:
Submitting to `horizon-skin` rather than `development`, since it builds on the skin in #1128 and the diff is then the new work rather than the whole skin again.

A climate page for `Horizon`, one per year. The other pages answer what the weather is doing. None answers what kind of year it has been.

- **The counted days.** Summer, hot and desert days, tropical nights, frost, ice, storm and rain days, each with its threshold and a breakdown by month. WeeWX counts these already, through `max_ge`, `min_le` and `sum_ge`. The NOAA reports have used them for years, in a text file few people open.
- **The year's shape**, after Walter and Lieth: mean temperature as a line, rainfall as bars at twice the scale. Where a bar falls below the line, the month was dry. Plotted in Celsius and millimetres, since that is what the ratio is stated in, and labelled back into the reader's own unit.
- **The water balance.** Rain above the line, evapotranspiration below it, and what the ground kept between them. Only where the station records ET.
- **The whole record**, month by month, for temperature, rainfall and evapotranspiration.

The year in progress is written every cycle. Each year that has ended is written once, since `CheetahGenerator` skips a summary page whose period is over. Fourteen years of record cost one page per cycle, not fourteen. `show_climate` turns it off.

Two fixes came out of building it. The unit picker's table only knew the observation types that appear in a plot, so rain rate, UV and the statistics tables never moved. And on a phone the masthead was 60 pixels wider than the page: below 40rem it folds into a menu, and the tab strips scroll rather than putting tabs out of reach.

The first commit is the rain fix from #1128. It drops out of this diff once that branch is pulled.

Checked at 305, 375 and 685 pixels, in both unit systems, with every period selected. Suite and a render of every shipped skin are clean on 3.7 and 3.14.


src/weewx_data/skins/Horizon/climate.inc
```diff
@@ -0,0 +1,378 @@
...
+##         min_ge and min_le test the daily maximum or minimum, and sum_ge tests the
+##         daily total.
+##
+## The thresholds are metric whatever the station records in, because that is how the
+## definitions read: a summer day is 25 C, and 77 F is the same day under another
+## name. WeeWX converts.
+## The threshold is a full ValueTuple, (number, unit, group). Two of the three would
+## be enough to convert it, but weewx.xtypes reads the group as well.
+#set $climate_days = [
+  {'key': 'rain_day',     'obs': 'rain',     'agg': 'sum_ge',
```
tkeffer, line 27: Good choices ... if you're using metric units. For others, the cutoffs come out at random numbers. For example, "Summer day" is 77°.

tkeffer (commented):
There is also a merge conflict with `horizon.js`

src/weewx_data/skins/Horizon/climate.inc
hilman2: They are the standard cutoffs. The standard defines them in Celsius.

`SU`, `TR`, `FD` and `ID` are ETCCDI core indices: 25, 20, 0 and 0 degrees C. The DWD names go with those. It adds 30 for a hot day, 35 for a desert day.

77°F is what 25°C comes to. A round 75 or 80 counts a different set of days.

https://etccdi.pacificclimate.org/list_27_indices.shtml

hilman2:
Folded into #1128, which carries the climate page now. Closing this one.

PR #1133, opened 2026-08-29
Horizon: charts on ECharts, a forecast, the planets, and a second look

hilman2:
Charts drawn with ECharts, a forecast, the planets, and a second look for the skin. One fix in the JSON generator.

image

Submitting to `development`, as you asked in #1128. It builds on what went in there: the three tiers, the extend mechanism, and the time budget.

## The generator fix

A finished raw day held one slot more than its day. Nothing could ever fill it. The report that could is the one that finds the day over and skips the file.

```
tempdew-raw-2026-08-27.json  start=1787814000 interval=1800 count=49
                             start + count*interval = 1787902200
tempdew-raw-2026-08-28.json  start=1787900400
```

A line drawn from both files had a hole where they met. `stop` was rounded down to a whole slot and then a slot added. It is rounded up instead.

The slot a value goes in now comes from the interval's start, as `get_series()` reports it. `intervalgen()` clips its last interval to the end of the span. So an interval is not always a whole step long.

Two tests cover it. One walks five days of junctions. The other checks that no bar lands closer to its neighbour than the hour it is totalled over.

`test_extending_matches_a_full_rebuild_over_a_dst_boundary` was comparing two different states of the database. 13 March is 23 hours long, so the fixed step did not divide the span. Both ends now finish on the same slot boundary.

## ECharts

uPlot draws lines quickly and little else. Bars, wind arrows, night shading and a tooltip each needed code of their own. The wind vector plot had no drawing at all.

ECharts brings all of it. Night shading is a `markArea`, the wind arrows a `custom` series. A bar knows how wide the interval is that it stands for. 1 MB against uPlot's 50 kB, fetched once.

Three things about the axis:

- A step is round in the reader's unit, not only in the report's. A fifth of an inch of mercury is 6.77 hectopascals. An axis built on that reads 989.5, 996.3, 1003.0.
- The unit stands above the axis again.
- `dataZoom` is gone. Its `moveOnMouseWheel` defaults to true, which swallowed the wheel: the page could not be scrolled with the pointer over a chart. Measured with a real wheel event, `scrollY` 0 to 0 with it, 0 to 400 without.

## Cards that follow the span

Over a week the reading at this instant says nothing about the week. The card shows the mean instead, with the ends of the span under it.

| span | barometer | min | max |
|---|---|---|---|
| day | 29.293 inHg | 29.293 | 29.902 |
| week | 29.957 inHg | 29.000 | 31.000 |
| year | 30.002 inHg | 29.000 | 31.000 |

Rain and evapotranspiration are summed, not averaged. The wind shows the mean, and the direction it mostly came from. That is the vector mean, not where the vane points now. The figures come from the template, so nothing is fetched.

## New panels

Every reading carries a symbol and its range for the day. The symbols are IBM Carbon, Apache-2.0, listed in `icons/LICENSE-Carbon.txt`. They are drawn as a CSS mask, so they take the page's own ink.

Sun and moon can share a panel. `sun.inc` and `moon.inc` are each a frame around a body file. `sunmoon.inc` puts both bodies in one panel.

`planets.inc` is new: the naked-eye planets, their altitude, and when they rise and set. Any body pyephem knows can be named in `planets`.

The forecast reads `data/forecast.json` where something on the station writes one. Where nothing does, it asks Open-Meteo itself. `forecast-fetch.py` writes the format and documents it, so weewx-dwd or a script of your own can feed it instead. `browser_fetch = false` turns the direct fetch off. That matters on a station published to the world: every reader's browser otherwise talks to Open-Meteo. Codes are WMO 4677, translated in the language files.

## The second look

```
[DisplayOptions]
    custom_css = flavor-deck.css
    custom_js  = flavor-deck.js
```

`custom_js` is new and works like `custom_css`: any script, deferred. The stylesheet does the paint. The script moves the navigation into a column and puts the station's particulars in a footer. On a phone that column joins the menu the skin already has.

Without those two lines the skin looks as it did.

## What it does not touch

The `ImageGenerator` and the PNGs. The archive tiers and the extend mechanism, apart from the boundary above.

## Tests

567 tests, no failures, on 3.7 and 3.14. Ten skins rendered in two languages, no errors and no empty pages. The forecast, the span switching and the wheel were measured in a browser against a live station.


src/weewx/jsongenerator.py
```diff
@@ -428,15 +509,28 @@ def write_tier(spans, kind, grids, stamp_of, name_of, grid_of, tier_from,
...
+                to carry forward and what to record are the same for all of them.
+
+                'metered' tiers are held to the budget. The raw tier is not one:
+                it is cheap, and it is what the day view is drawn from, so a report
+                that deferred it would leave the page without today.
+
+                Newest span first, so that a run giving up partway leaves the far end
+                of the record unbuilt rather than this year.
+
+                Args:
```
tkeffer, line 518: An improvement, but still not enough for an IDE to infer the types. I would try:

```
                    Args:
                        spans (Iterable[weeutil.weeutil.TimeSpan]): The spans to write,
                            one file each.
                        kind (str): Which tier, as the index names it.
                        grids (str): The index key holding the grid each file was written on.
                        stamp_of (Callable[[weeutil.weeutil.TimeSpan], int | str]): Called
                            as ``stamp_of(span)``. Returns the index stamp for a span.
                        name_of (Callable[[int | str], str]): Called as ``name_of(stamp)``.
                            Returns the JSON file name for a stamp.
                        grid_of (Callable[[int | str, int | None], int]): Called as
                            ``grid_of(stamp, existing)``. Returns the grid, in seconds, for
                            a stamp. ``existing`` is the grid recorded for that stamp in the
                            previous index, or None if there is none.
                        tier_from (int): The oldest instant this tier reaches.
                        metered (bool): Whether the budget applies. The raw tier is not
                            metered.```

Note that you don't need to actually do `from typing import Callable`. PyCharm figures it out.

**Comment by bellrichm** [comment]
Question:
I was experimenting with creating current.json more frequently (in the hope of getting “live” updates) and noticed that the min/max are not updated. So it looks a bit strange. So I guess that the current.json needs to be “in lock step” with the rest of the data being updated?

I got an experimental pi-3 up and running. This is looking great! I can’t wait to replace my experimental skin with this!

**Comment by hilman2** [comment]
Open an issue for it. Live data is its own subject, and I would rather not push anything
onto this PR before Tom is through it.

The background is short. `$current` is the last archive record: `StdReport` caches it on
`NEW_ARCHIVE_RECORD` and hands it to the report thread, and LOOP packets never reach
Cheetah. So `current.json` holds the same numbers between records, however often it is
rendered. What it buys is that the page is not reloaded, and that it is a few hundred
bytes instead of the page and every PNG.

The order matters as well. It is written last because its timestamp is what tells the
page that the rest has been re-rendered. Written ahead of the rest, the page fetches
`index.html`, gets the one from the last cycle, and marks the timestamp as seen. It does
not ask again. The readings keep moving, since they come out of `current.json` itself.
The min/max sit in the panels, so they stay where they were.

The skin will be a live project for months, and real live data is what I want to look at
next. Good to hear it runs on a pi-3.

**Comment by tkeffer** [comment]
Miscellaneous comments

### Comments on `skin.conf`

Options `hide_plot_groups` and `hide_panels` are going to create support issues. I get why you created them, but someone can just as easily copy `plot_groups`, comment out one version, then edit the other. This keeps state in one place, which is a lot simpler than having to look in two places. It's also the way other skins do it.

The section `[Generators]` should go on the bottom. It's an implementation detail. That's also the way the other skins do it.

I don't think people are going to want to run both the Image and JSON generators at the same time. They serve the same function: display time series. You should just focus on running the `Horizon` skin using the JSON generator and make it simple. It should accept a stanza `[HorizonReport]`, that shares the syntax for specifying plots with `[ImageGenerator]`. No `source` option.

For option `sidebar` and `sidebar_responsive` explicitly list the possible choices. 

I love the modular panel concept, but it only covers a subset of the widgets. For example, it does not cover the plots ("History"), or the statistical summary ("Statistics"). Is it possible to include them? Then a user can use the `panels` option to lay out the page exactly as s/he wants.


### Comments on functionality

I used to have an extra temperature sensor on my VantagePro2, but it broke nearly 3 years ago (2023-10-17 to be precise), yet I'm seeing a plot panel for `Temperature1`. The scan for `has_data` must be way too wide.

The "Statistics" panel shows highs and lows, but only the high gets a time.

Using the "default" units, the forecast is always in metric. To me, default means what the user has specified in the `lang` setting, not the units that the information comes in.

### SummaryImage generator

As I mentioned earlier, the best PRs create infrastructure that can be used by other extension writers. Your JSON generator does this.

The SummaryImage generator does not. Yes, it's definitely nice to have, but it should be packaged as an extension. Then people can then install it only if they have a use for it.

**Comment by hilman2** [comment]
All nine points are in, in seven commits.

One of them departs from what you asked. `[HorizonReport]` holds until the skin
runs twice. Its own language switcher runs it as `[[HorizonReport]]` and
`[[HorizonReportEN]]` against one `skin.conf`. A section named for the report
then serves the first and leaves the second with no plots. The end-to-end render
caught it:

```
No plot definitions found, in [JSONGenerator] or [HorizonReport-de] or [ImageGenerator]
```

So the definitions sit in `[JSONGenerator]` itself, in the syntax the
`ImageGenerator` uses. `source` is gone. `[ImageGenerator]` is read only where
that section holds no plots, so a skin drawing both still defines each plot once.

The rest:

- `hide_plot_groups` and `hide_panels` are gone, and the documentation says to
  copy the line and comment one copy out.
- `[Generators]` is last, as in the other skins.
- `sidebar` and `sidebar_responsive` list their choices.
- Both columns are panel lists. `main_panels = history, hilo` beside
  `panels = current, forecast, …`, so the charts and the statistics move or drop
  out like anything else. The charts are `history.inc` now.
- `Temperature1`: a card was drawn for every group the archive index names, and
  it names a group for as long as any of its files survive. It now asks which
  files the span on screen would be drawn from, so a sensor that stopped in 2023
  leaves no card behind.
- The low carries its own time.
- `index.json` says which units the report rendered in, and "Default" follows
  those rather than what Open-Meteo happens to send.
- The summary image generator is out. It never shipped, so there is no change
  log entry for it.

563 tests, no failures, on 3.7 and 3.14. Ten skins rendered in two languages, no
errors and no empty pages.

tkeffer:
Thanks. Looks great.

The PR has been merged and pushed to `development`. No more force pushes, please!

hilman2:
Understood, no more force pushes.

One had already gone out: #1129 was rebased and force-pushed about half an hour
before your message, so that branch is new since you last saw it. It merges into
the current `development` without conflict.

PR #1135, opened 2026-08-30
Add option loop_timeout, a deadline on driver calls

### LINKED ISSUE #1134: A driver that blocks stalls weewxd silently
A driver that blocks in `genLoopPackets()` stops weewxd, silently. Would you take
an option that puts a deadline on driver calls? I have an implementation with
tests.

Here is `weewxd` with a station that goes quiet after three packets:

```
03:54:21 INFO weewx.engine: Starting main packet loop.
LOOP:   ... outTemp: 0.0
LOOP:   ... outTemp: 1.0
LOOP:   ... outTemp: 2.0
03:54:46 INFO __main__: Received signal TERM (15).
```

Nothing in those 25 seconds. No archive record, no report, no log entry. The
process is alive and its exit code is 0.

No service can catch it. `POST_LOOP` is dispatched only from the `BreakLoop`
handler in `StdEngine.run()`. `BreakLoop` comes from `StdArchive.check_loop`,
which the engine dispatches per packet. No packet, no event.

Calls outside the loop have the same problem. `_get_console_time()` runs before
every archive period and `genArchiveRecords()` at the end of one, both on the
engine's thread.

hilman2:
Submitting to `development`, as this is a feature. Fixes #1134.

With `loop_timeout` set, the engine wraps the driver. Every call runs on a worker
thread, and the engine waits for the answer only that long. After that it raises
`WeeWxIOError`, which `weewxd` already handles: it waits `retry_wait` seconds and
builds a new engine.

``` ini
[Station]
    station_type = Vantage
    loop_timeout = 300
```

Without the option nothing is wrapped. `station_type` stays the real driver, so
`weectl device` and the configurators are untouched.

The guard relays reads, writes, method calls and generators, one element at a
time. The driver runs no further ahead of the engine than it does now.

Two things it cannot do. A thread blocked in a read cannot be interrupted, so it
stays for the life of the process and keeps the device open. The engine comes
free, the thread does not. And the driver no longer runs on the main thread, so
it cannot register signal handlers.

Tested on 3.7 and 3.14. 20 new tests, plus a run of `weewxd` against the
interceptor driver, fed real ecowitt packets across two archive periods.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -1,6 +1,14 @@
 WeeWX change history
 --------------------
 
+### 5.6.0
+
+Added option `loop_timeout`. If set, WeeWX gives up on a station driver that has
+stopped answering, then restarts the engine. Without it, a driver that blocks and
+never returns takes the archive, the reports and the uploads with it.
+[PR #1135](https://github.com/weewx/weewx/pull/1135).
+
+
 ### 5.5.0 6-Aug-2026
 
 Added the ability for services to bind to a `SHUTDOWN` event. This allows
--- a/docs_src/reference/weewx-options/stations.md
+++ b/docs_src/reference/weewx-options/stations.md
@@ -85,6 +85,25 @@ Valid station types include:
 | **WS28xx**        | La Crosse 28xx stations.                                                |
 
 
+#### loop_timeout
+
+The time (in seconds) WeeWX will wait for the station driver to answer. If the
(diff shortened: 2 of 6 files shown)


tkeffer:
I'm not keen on the idea of introducing another thread between the engine and the driver. 

Any reason why this could not be packaged as a type of driver that sits between the engine and the "real" driver?

PR #1137, opened 2026-08-30
Add StdStationWatch, which notices that the data has stopped

### LINKED ISSUE #1136: Let a service know that the station has stopped reporting
WeeWX cannot tell a service that the data has stopped arriving. Would you take a
small watcher in the core that does? The notification side stays out of it.

Nothing in the core knows the age of its own data. `NEW_ARCHIVE_RECORD` is the
event that would carry it, and during an outage that is exactly the event that
does not arrive. A service can only find out by running a thread of its own and
querying the database, and every service that wants to know does it again.

What I would add:

- Two event types, `STATION_DOWN` and `STATION_UP`, carrying the timestamp of
  the newest record and the length of the gap.
- A service that watches the record age and dispatches them, and logs the
  change.
- One option for the deadline, off by default.

That is the part only the core can do. What to do about it stays outside: mail,
Telegram, a webhook, a program, all of that belongs in extensions that bind to
the events. The log entry alone already covers anyone watching their journal.

Related to #1134, which is about a driver that stalls rather than a station that
goes quiet. They catch different failures and neither one finds the other.

hilman2:
Submitting to `development`, as this is a feature. Fixes #1136.

`StdStationWatch` checks how old the newest archive record is. Past `max_age` it
logs a warning and issues `STATION_DOWN`, and when records arrive again it logs
and issues `STATION_UP` with the length of the gap.

``` ini
[StdStationWatch]
    max_age = 1800
```

Without `max_age` it binds to nothing and starts nothing.

Both events are issued from the service's own thread, which is the one thing
worth arguing about. They cannot come from the main loop: it only comes round
when a LOOP packet arrives, and that is what has stopped happening. So anything
bound to these two runs off the main thread and must not touch the console. It
is documented on the event classes and in the service docstring.

The state is held in memory, not written anywhere. An engine that restarts
during an outage therefore issues `STATION_DOWN` again. Consumers that mind can
keep their own record, and the alternative was writing to the daily summary
metadata through a private method.

What this does not do is decide what happens next. Mail, Telegram, a webhook, a
program: all of that belongs in an extension binding to the events. I have one,
and it is not part of this.

Tested on 3.7 and 3.14. Nine new tests, and five deliberate mutations of the
service, each caught by the suite.

### DIFF
--- a/docs_src/changes.md
+++ b/docs_src/changes.md
@@ -1,6 +1,13 @@
 WeeWX change history
 --------------------
 
+### 5.6.0
+
+Added service `StdStationWatch`, which issues `STATION_DOWN` and `STATION_UP` as
+the archive data stops and starts. Set `max_age` to switch it on.
+[PR #1137](https://github.com/weewx/weewx/pull/1137).
+
+
 ### 5.5.0 6-Aug-2026
 
 Added the ability for services to bind to a `SHUTDOWN` event. This allows
--- a/docs_src/reference/weewx-options/stdstationwatch.md
+++ b/docs_src/reference/weewx-options/stdstationwatch.md
@@ -1,0 +1,29 @@
+# [StdStationWatch]
+
+This service watches how old the newest archive record is. When it gets older
+than `max_age`, WeeWX logs a warning and issues a `STATION_DOWN` event. When
+records arrive again, it logs and issues `STATION_UP`.
+
+Both events are issued from the service's own thread, not from the main loop.
+Anything bound to them must not use the console.
+
+#### max_age
+
+Seconds the newest archive record may be, before the station counts as having
+stopped reporting. Make it a multiple of the archive interval, so that a single
+missing record does not trip it. No default: without this option WeeWX does not
+watch at all.
+
+``` ini
+[StdStationWatch]
+    max_age = 1800
+```
+
+#### data_binding
+
+The binding to watch. Default is `wx_binding`.
+
+``` ini
+[StdStationWatch]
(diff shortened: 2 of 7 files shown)


roe-dl:
Why not use existing [WeeWX Healthcheck extension](https://github.com/uajqq/weewx-healthchecks)?

hilman2:
Thanks, I did not know that one. It solves the same problem from the other side.

`weewx-healthchecks` binds to `NEW_ARCHIVE_RECORD` and pings a URL. The deciding
happens at healthchecks.io: no ping, an alert. That also catches a machine that
is switched off, which nothing running inside WeeWX can do. It needs an account
and a working internet connection.

This one decides locally. Nothing leaves the machine, and the answer is
available to WeeWX itself: a log entry, and an event that any service can bind
to. It says nothing when the machine is off.

They are complementary rather than alternatives. What is proposed here is not a
notifier: an extension like that one could bind to the events instead of running
a thread and a database query of its own.

roe-dl:
> I did not know that one.

You sent lots of PRs the last days. Very huge amount of PRs. For people that use and develop WeeWX in their spare time, it is not easy to follow. 

I guess, you should study all the existing extensions before you start coding. This prevents your PRs from conflicting with existing and well-known solutions.

tkeffer:
This is best done as a WeeWX extension.

The best PRs create leverage. They add new capabilities that extension writers can use. Your JSON generator is a great example of this.

This PR is a feature, not new functionality.

PR #1139, opened 2026-09-04
Horizon rework v2

tkeffer:
Ready for reviewing the JSON generator and Horizon skin.


hilman2:
The tree in `horizon-rework-v2` is byte-identical to `d5ec438a`, and to the `horizon-on-development` branch I sent you on 2 September. `development` is an ancestor, so the merge is a fast-forward. `summaryimage.py` stays gone. `history.inc` is back.

I rendered the Horizon skin from `f3f2a034` and from this branch against the synthetic database, in `en` and `de`, and compared the pages. Two runs of the same commit came out byte-identical, so there is no noise to subtract.

Six of 950 files differ: `index.html`, `climate.html` and `climate/climate-2010.html`, in both languages. All of it is whitespace from bd0b8a92 and d5ec438a. Nothing moves in a browser. Comparing every element for tag, class, position, size and text gives 0 differences out of 1139 on `index.html`, 891 on `climate.html` and 875 on `climate-2010.html`. Same at 375px.

Enjoy the long weekend.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # If you have a Google Analytics GA4 tag, uncomment and edit the next line.
+    #googleAnalyticsId = G-ABCDEFGHI
+
+
+###############################################################################
+
+[DisplayOptions]
+
+    # Show link to NOAA-style summary reports?
+    show_reports = True
```
tkeffer, line 24: It looks like either way the NOAA reports get generated. The option only controls whether the link is displayed. This does not look like a useful option.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+
+    # Show the climate page: the counted days, the year's shape, and the record
+    # year by year. It costs one query per figure per month, against the daily
+    # summaries, so a station with a long record may want it off.
+    show_climate = True
+
+    # Show a link to the RSS feed?
```
tkeffer, line 31: Again, not very useful. Either way, the feed gets generated.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+
+[DisplayOptions]
+
+    # Show link to NOAA-style summary reports?
+    show_reports = True
+
+    # Show the climate page: the counted days, the year's shape, and the record
+    # year by year. It costs one query per figure per month, against the daily
+    # summaries, so a station with a long record may want it off.
+    show_climate = True
```
tkeffer, line 29: But turning it off doesn't actually save any compute time. It just controls whether to show the link.  Again, not a useful option

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    show_rss = True
+
+    # Which types appear in the RSS feed.
+    observations_rss = outTemp, inTemp, barometer, windSpeed, rain, rainRate, windchill, heatindex, dewpoint, outHumidity, inHumidity
+
+    # Which plots appear on the telemetry page.
+    telemetry_plot_groups = rx, volt
+
+    # Link to the server-rendered PNG beside each chart. Left out, the links
+    # appear whenever the ImageGenerator is in [Generators].
```
tkeffer, line 41: To generate static PNG images you not only need to include `weewx.imagegenerator.ImageGenerator` in `generator_list`, you also need a `[ImageGenerator]` stanza. 

I see being able to generate PNG files as being of limited usefulness. I don't think many people are going to want to do this. It complicates your comments and will make support more difficult.

If someone really wants them, they can run a 2nd skin.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # show_image_links = True
+
+    # How often the page checks for a new record, in seconds. 0 disables live
+    # updating. There is no point going below your archive interval.
+    refresh_interval = 60
+
+    # Which types appear in "current conditions", and in what order.
+    observations_current = outTemp, heatindex, windchill, dewpoint, outHumidity, barometer, windSpeed, rain, rainRate, UV, radiation, lightning_strike_count, inTemp, inHumidity, extraTemp1, extraHumid1, extraTemp2, extraHumid2, pm1_0, pm2_5, pm10_0
+
+    # The card of current conditions: which reading is set large at the top, and
```
tkeffer, line 52: This next section is very confusing. Terminology shifts. Is it a "card", or "dashboard", or a "headline?"

Do you even need the dashboard settings? What do they do that observation headline doesn't do? 

Anytime you have overlapping functionality between options you are going to confuse users. Each option should have one, well-defined role.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    show_climate = True
+
+    # Show a link to the RSS feed?
+    show_rss = True
+
+    # Which types appear in the RSS feed.
+    observations_rss = outTemp, inTemp, barometer, windSpeed, rain, rainRate, windchill, heatindex, dewpoint, outHumidity, inHumidity
+
+    # Which plots appear on the telemetry page.
+    telemetry_plot_groups = rx, volt
```
tkeffer, line 38: This option, `telemetry_plot_groups` feels out of place. It controls which observation types to display and in what order. Put it with the other options in this category, such as `observations_current`, `observation_stats`, etc.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    sensor_connections = rxCheckPercent, signal1, signal2, signal3, signal4, signal5, signal6, signal7, signal8
+    sensor_batteries = outTempBatteryStatus, inTempBatteryStatus, rainBatteryStatus, hailBatteryStatus, snowBatteryStatus, windBatteryStatus, uvBatteryStatus, txBatteryStatus, batteryStatus1, batteryStatus2, batteryStatus3, batteryStatus4, batteryStatus5, batteryStatus6, batteryStatus7, batteryStatus8
+    sensor_voltages = consBatteryVoltage, heatingVoltage, supplyVoltage, referenceVoltage
+
+    # Which plots to show, and in what order. The names refer to plots defined
+    # in [JSONGenerator] below, without the time span prefix: 'wind' covers
+    # 'daywind', 'weekwind', 'monthwind' and 'yearwind'.
+    plot_groups = tempdew, tempfeel, hum, barometer, wind, winddir, windvec, rain, radiation, UV, ET, lightning, tempin, humin, tempext, humext, templeaf, wetleaf, tempsoil, moistsoil, pm
+
+    # The time spans offered by the period switcher. How long each one covers is
```
tkeffer, line 82: Assume no PNGs.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # NOT set here. It comes from 'time_length' in the matching [[*_images]]
+    # section below, so that the charts and the PNGs show the same span.
+    periods = day, week, month, year
+
+    # Which side the panels sit on when there is room for two columns.
```
tkeffer, line 87: I have a very wide screen, but I'm seeing no difference between these two options. Also, which "panels"?

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # Either 'right' or 'left'.
+    sidebar = right
+
+    # And where they go when there is not. Either 'top', above the charts, or
```
tkeffer, line 91: Again, what is "they"?

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # 'bottom', below them.
+    sidebar_responsive = top
+
+    # The page is two columns of panels. Each name is an .inc file in the skin
```
tkeffer, line 95: If this is what "penels" refers to, wouldn't it make more sense to put this explanation up at the top?

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # of these lists. No template editing needed.
+    #
+    # The wide column. 'history' is the charts and their span switcher, 'hilo' the
+    # statistics table. Only one 'history' belongs on a page.
+    main_panels = history, hilo
+
+    # And the narrower one beside it.
+    panels = current, forecast, sunmoon, planets, imagery, sensors, about
+    #
+    # "sunmoon" is the sun and the moon in one panel; name "sun" and "moon"
```
tkeffer, line 106: I don't think you need this explanation. In any case, it belongs above the option, not below.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+
+
+    # A stylesheet of your own, loaded after horizon.css, and a script, loaded
+    # after horizon.js. Put the file in the skin directory and add it to
+    # 'copy_once' in [CopyGenerator], or it will not be installed.
+    #
+    # custom_css = my-station.css
+    # custom_js  = my-station.js
+    #
+    # 'flavor-deck.css' and 'flavor-deck.js' ship with the skin and give it a
```
tkeffer, line 120: You make it sound like `horizon.css` is the default look and feel of the skin, but the fact is, as it ships, `flavor-deck` is the default look.

Rethink this explanation, and how it relates to the user overriding with his/her own.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # second look: flatter, denser, with the navigation in a column. Comment
+    # them out for the skin's own appearance.
+    custom_css = flavor-deck.css
+    custom_js  = flavor-deck.js
+
+    # The look of the skin. horizon.css defines it as custom properties, and any
```
tkeffer, line 126: What is the advantage of this approach over just modifying the CSS file, then using it in `custom_css` above? After all, both `skin.conf` and the CSS files get overwritten in an upgrade. 

This adds complexity.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    #         font    = system-ui, -apple-system, "Segoe UI", Roboto, sans-serif
+    #
+    #     [[[Dark]]]
+    #         bg      = "#0e161f"
+    #         surface = "#16222e"
+    #         accent  = "#6fa9d6"
+
+    # A language switcher, shown only if there is more than one entry below.
+    #
+    # WeeWX renders each page in a single language, so this cannot translate the page
```
tkeffer, line 151: What is "this?"

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # in the browser. It links to a copy of the page rendered in that language. To
+    # use it, run the report once per language in weewx.conf, each with its own lang
```
tkeffer, line 153: Don't think this explanation is really necessary. This is the standard way of supporting more than one language. It's documented in the Customization Guide. Or, if it's not, let's change the Guide as necessary.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    #         label = English
+    #         path = en
+
+
+###############################################################################
+
+    [[Forecast]]
+        # Where the forecast comes from.
+        #
+        # The page reads 'data/forecast.json' first. Anything can write it: the
```
tkeffer, line 184: I found these three paragraphs confusing.

It also seems there is overlap between the effects of `browser_fetch` and `file`.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+        browser_fetch = true
+
+        # Whether anything on this station writes 'data/forecast.json'. Left
+        # false the page does not ask for it, which keeps a 404 out of every
+        # reader's console; set it true once something does write the file, and
+        # it takes precedence over the fetch above.
+        file = false
+
+        # How many days, and how many of the hours ahead. The hours are shown
+        # every third one.
```
tkeffer, line 202: "...every third one." Third one of what?

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+[JSONGenerator]
+
+    # This section controls the JSON files the charts are drawn from, and holds
+    # the plot definitions themselves further down. See "The JSON generator" in
+    # the Customization Guide, and [JSONGenerator] in the Reference Guide.
+
+    # Subdirectory of HTML_ROOT for the JSON files.
+    json_dest_dir = data
+
+    # One file per plot per time span. Off: the archive below covers the same
```
tkeffer, line 269: Did not understand this comment. Also, the option is "false", not "off".

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    #                 yscale, y_nticks, x_interval, vector_rotate, rose_label,
+    #                 show_daynight, skip_if_empty, stale_age
+    # PNGs only:      image_width, image_height, image_background_color,
+    #                 chart_background_color, chart_gridline_color, anti_alias,
+    #                 every *_font_* option, daynight_*_color, line_type,
+    #                 marker_type, marker_size, line_gap_fraction, rose_width,
+    #                 rose_height, rose_diameter, rose_line_width,
+    #                 bottom_label_format
+    #
+    # The PNGs are not generated at all unless you put the ImageGenerator back
```
tkeffer, line 300: Confusing. Again, focus on the case where you do not generate the PNGs. I don't think many people are going to want to do that.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # and it still scales down cleanly wherever it is embedded at 500 wide.
+    image_width = 1000
+    image_height = 360
+    image_background_color = "#ffffff"   # the skin's --surface, light theme
+
+    chart_background_color = "#ffffff"
+    chart_gridline_color = "#dbe4ec"    # the skin's --border
+
+    # Render larger still, then downsample: this is what removes the jagged
+    # edges the old plots were known for.
```
tkeffer, line 315: Which old plots? It's sufficient to say what the Seasons skin says: "Setting to 2 or more might give a sharper image with fewer jagged edges"
"

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+            [[[[referenceVoltage]]]]
+
+        [[[yearET]]]
+            plot_type = bar
+            aggregate_type = sum
+            aggregate_interval = 1d
+            [[[[ET]]]]
+                label = Evapotranspiration (daily total)
+
+    # The whole record, cut into files a page fetches one at a time. This is what
```
tkeffer, line 863: In WeeWX, "record" means something specific: the archive record. Perhaps use "The whole archive".

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+    # lets the charts show any span, rather than only the last day, week, month
+    # and year.
+    [[Archive]]
+        enable = true
+
+        # How readings are combined into a slot.
+        aggregate_type = avg
+
+        # One file per calendar year, on this grid, for this many years back.
```
tkeffer, line 872: When I think of "grid", I think of a two-dimensional structure. Here, things are one-dimensional. Why not just use the word "resolution": it's what you're using for the option name.

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+        fine_resolution = 900
+
+        # One file per day at the station's own archive interval, for this many
+        # days back. This is what the day view is drawn from. Older files are
+        # removed, so this tier does not grow.
+        raw_days = 30
+        # 0 means the archive interval, whatever the hardware turned out to use.
+        raw_resolution = 0
+
+        # How many seconds a report may spend on the archive. A report that runs
```
tkeffer, line 894: "How many seconds a report may spend before it is terminated. Work done before termination is saved and the next report carries on from there. The day view..."

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+        # out writes what it has worked out, and the next one carries on from
+        # there. The day view is never deferred. 0 does the lot in one run.
+        budget = 30
+
+        # Types that also carry their lowest and highest reading in each slot.
+        # An average turns a gust into a breeze. Each name costs two more
+        # queries per slot.
+        extremes = windGust, windSpeed, rainRate, radiation, UV
```
tkeffer, line 902: What is a "slot"?

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+        # than carrying it forward. 0 never does. Set it to '1d' if readings in
+        # your database are ever edited in place.
+        rebuild = 0
+
+        # Which subsection of the plot definitions to take the groups from, and
+        # what to strip off the plot names ('daytempdew' -> 'tempdew').
+        source_group = day_images
+        strip_prefix = day
+
+        # How far back to go. 0 means the entire record.
```
tkeffer, line 914: record = archive database

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+        max_days = 0
+
+        dest_dir = data/archive
```
tkeffer, line 917: What do these two options do?

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+
+###############################################################################
+
+[CopyGenerator]
+
+    copy_once = horizon.css, horizon.js, climate.js, echarts.min.js, favicon.ico, LICENSE-ECharts.txt, flavor-deck.css, flavor-deck.js, forecast.js, icons/*, icons/forecast/*, moon/*
+
+
+[Units]
+    [[TimeFormats]]
```
tkeffer, line 932: Why are you overriding the defaults? If you feel that the defaults are wrong, let's change them!

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,955 @@
...
+        day        = %H:%M
+        week       = "%H:%M, %A"
+        month      = "%d %b, %H:%M"
+        year       = "%d %b %Y, %H:%M"
+
+
+###############################################################################
+
+[Generators]
+    # The list of generators that are to be run. Horizon draws its charts in the
```
tkeffer, line 950: Abbreviate this to the same as Seasons: "The list of generators that are to be run."

tkeffer (commented):
I found the explanations in `skin.conf` to be very confusing because the terminology is not always well defined, and the effects of the options are not always clear.

I would suggest instead of trying to describe the effects of the options in the text file `skin.conf`, put them in the documentation. Then you can offer graphical examples of the effects of the options.

Overall, there are too many options and too many ways of doing the same thing. If I could give one bit of advice it would be to run this program, then follow the wisdom:

    python -c "import this"

hilman2:
The `skin.conf` comments are taken up in #1140.

src/weewx_data/skins/Horizon/horizon.js
```diff
@@ -0,0 +1,2603 @@
...
+
+  function remember(key, value) {
+    try { localStorage.setItem(STORE + key, value); } catch (e) { /* private mode */ }
+  }
+
+  function recall(key, fallback) {
+    try {
+      var v = localStorage.getItem(STORE + key);
+      return v === null ? fallback : v;
+    } catch (e) { return fallback; }
```
bellrichm, line 22: I’m not a front-end developer, but catching all exceptions is usually not a good idea. If necessary, a comment on why will help future developers…

hilman2: Fair point as a general rule, and it deserves an explanation in the code rather than in my head, so I have put one there (see the suggestion a few lines up; the branch lives in weewx/weewx, so I cannot push to it myself).

For the record, this is the one browser API where an unconditional catch is the idiomatic pattern rather than a shortcut. The Web Storage spec reports failure by throwing a `DOMException`, not by returning an error, and every case is a normal runtime condition rather than a bug:

- `SecurityError`: the user blocks cookies / site data, or the page has no origin (`file://`, sandboxed iframe, some embedded web views). Merely *evaluating* `window.localStorage` throws here, so `getItem()` is affected exactly like `setItem()`.
- `QuotaExceededError`: the origin's quota is exhausted, or the browser sets it to zero on purpose (Safari private browsing did that for years, some privacy modes still do).
- `TypeError`: the embedding context removed the API and `localStorage` is `undefined`.

JavaScript has no typed catch clause. The closest approximation is `catch (e) { if (!(e instanceof DOMException)) throw e; }`, which would rethrow precisely the `TypeError` case and turn a missing convenience (remembered theme, units, period) into a blank page for the users whose browsers are the most locked down. The correct response is identical for all three: do not persist, let the caller's fallback win.

That is also why the swallow is confined to these two helpers. They are the only place in the skin that touches storage, and nothing else in `horizon.js` catches broadly.

src/weewx_data/skins/Horizon/horizon.js
```diff
@@ -0,0 +1,2603 @@
...
+
+  var CFG = window.HORIZON || {};
+  var DATA_DIR = CFG.dataDir || 'data';
+  var STORE = 'weewx.horizon.';
+
+  /* ------------------------------------------------------------- storage */
+
+  function remember(key, value) {
+    try { localStorage.setItem(STORE + key, value); } catch (e) { /* private mode */ }
+  }
```
hilman2, line 16: Explanation for the unconditional catch, as requested in the thread below:

```suggestion
  /* ------------------------------------------------------------- storage */

  /* localStorage only remembers user preferences (theme, units, period);
   * nothing the page needs to render lives there.  The Web Storage API
   * signals failure by throwing rather than by returning an error, and each
   * throw below is an expected runtime condition, not a bug:
   *
   *   SecurityError       cookies / site data blocked by the user, or a page
   *                       without an origin (file://, sandboxed iframe, some
   *                       embedded web views).  Merely touching the
   *                       `localStorage` property throws, so getItem() fails
   *                       as well as setItem().
   *   QuotaExceededError  the origin's quota is exhausted, or the browser
   *                       sets it to zero on purpose (Safari private browsing
   *                       did this for years; some privacy modes still do).
   *   TypeError           the embedding context removed the API entirely and
   *                       `localStorage` is undefined.
   *
   * Which of these fired changes nothing about the right response: the
   * preference is simply not persisted and the caller's fallback wins.
   * JavaScript has no typed catch, and rethrowing anything that is not a
   * DOMException would turn a missing convenience into a blank page for
   * exactly the users whose browsers are the most locked down, so the catch
   * is deliberately unconditional.  These two helpers are the only place in
   * the skin that talks to storage. */

  function remember(key, value) {
    try { localStorage.setItem(STORE + key, value); } catch (e) { /* see above */ }
  }
```

src/weewx_data/skins/Horizon/horizon.js
bellrichm: Thanks

src/weewx_data/skins/Horizon/skin.conf
```diff
@@ -0,0 +1,772 @@
...
+
+[JSONGenerator]
+
+    # Where the JSON files will be put, relative to HTML_ROOT
+    json_dest_dir = data
+
+    # Whether to write one file for each plot and time span, true or false. The
+    # skin sets false, because the files of [[Archive]] below cover the same time
+    # spans and go further back.
+    periods = false
```
tkeffer: How useful is this option? When do you see someone using the period files, instead of `archive`? After all, you do not use them.

The idea that describing the JSON files in the same way as the image files is a powerful one, but perhaps it has outlived its usefulness?

In any case the name `periods` is not a good one. It conflicts with `periods` under `[DisplayOptions]`.

PR #1140, opened 2026-09-10
Horizon skin.conf: fewer options, plainer words

hilman2:
Submitting to `horizon-rework-v2`, so that #1139 stays the place for it.

Six commits, one per theme, for the 28 comments on `skin.conf`.

Two questions first.

`sidebar` does nothing with `flavor-deck`, the look that ships. The comment says so now. Should the option go instead?

`[Units] [[TimeFormats]]`: yes, let's change the defaults. That is a PR against `development`, and this section goes with it. Shall I open it?

Gone: `show_climate`, `show_reports`, `show_rss`, `dashboard_lead`, `dashboard_readings`, `observations_headline`, the forecast option `file`, `[[Theme]]`, `show_image_links`, every `[JSONGenerator]` option the generator does not read, and `font/`, which only the PNGs used. `custom_css` and `custom_js` take a list now, so a stylesheet of your own can follow `flavor-deck`.

The rest is wording, in the words the comments asked for. The language switcher and the two looks are explained in the guide.

Behaviour that changes beyond the options:

- The first of `observations_current` leads the current conditions card and `current.json` alike. Before, the card led with the headline and `current.json` with `dashboard_lead`: set the headline to `windSpeed` and the live update put the day's range of `outTemp` beside a wind speed.
- Which plot options the generator reads is in the reference now, taken from the code. The list `skin.conf` had put `line_gap_fraction` on the wrong side.

Rendered against the synthetic database before and after, every JSON file is byte for byte the same. Green on 3.7 and 3.14.


tkeffer:
I'm still finding the Horizon / flavor-deck dichotomy terribly confusing for two reasons.

1. By making the `flavor-deck` look the default, you have to explain how a supposedly optional look has been, in fact, activated in the shipped version. If you want to make it the default, why not flip the Javascript and CSS files with their Horizon counterparts? Then `custom_css` and `custom_js` would be commented out in the shipped version. 
2. But there's an additional problem. The explanation about the page being two columns, one wide, one narrow, applies only to the `horizon` look. The `flavor-deck` renders completely differently, so that explanation just confuses the user.

Personally, I would stick with just one look. 

But if you really want to ship with two different looks then, at least, segregate out the options needed for the alternative look in one place and make it totally clear that the user doesn't have to mess with these options unless they want the different look.

tkeffer:
Manuel?

hilman2:
Pushed: one look now.

The Deck look replaced the Horizon files, as you suggested. `flavor-deck.css` and `flavor-deck.js` are gone. `custom_css` and `custom_js` are commented out in the shipped `skin.conf`. `sidebar` is gone too, and the two-column text with it. That settles my first question.

The diff is big: 58 files, +3175 / -4862. Most of it is `horizon.css` and `horizon.js` taking over the content of the deleted flavor files. 9575af8c is the change, 9eb59d03 the guide.

`[[TimeFormats]]` stays in the skin for now. If you change the defaults, it can go.

44b1ef2f adds the comment bellrichm asked for on the `catch` in `horizon.js`.

Green on 3.7 and 3.14.

PR #1145, opened 2026-09-16
Fix StdWunderground startup crash with rtfreq, and rapidfire endpoint routing

rnavarro:
I was moving my Wunderground uploads from an Ecowitt gateway to WeeWX, and while working out how to keep the post interval at the 16 seconds the gateway had been using, I found that setting `rtfreq` in `[[Wunderground]]` stops weewxd from starting.

`StdWunderground` pops `rapidfire` and `archive_post` out of the shared options dictionary, then passes what is left of that dictionary on to `AmbientThread`, which has no `rtfreq` parameter. Service construction therefore raises `TypeError: AmbientThread.__init__() got an unexpected keyword argument 'rtfreq'`. Since the failure happens while the services are being constructed, it takes the other RESTful services down with it, CWOP among them.

Because `archive_post` defaults to the negation of `rapidfire`, setting `rtfreq` at all is enough to reach this, unless `rapidfire` is enabled and `archive_post` is left off. The documented advice permits both options to be set to true, and that combination is one of the ones that crashes.

There is a second problem in the same block. With both modes enabled and no explicit `rtfreq`, the archive branch calls `setdefault('server_url', ...)` on the shared dictionary, so the rapidfire thread built afterwards picks up the archive endpoint and posts to `weatherstation.wunderground.com` with `realtime=1` instead of to `rtupdate.wunderground.com`.

This change pops `server_url` and `rtfreq` out of the shared dictionary before either thread is built, and gives each thread its own endpoint, passing `rtfreq` only to the rapidfire thread, which is the only one that accepts it. A `server_url` that the user has configured still overrides both defaults. `rtfreq` is not exposed in the shipped `weewx.conf` template, so it is only reachable if a user adds it by hand.

`src/weewx/tests/test_restx.py` gains a `TestWunderground` class which constructs the service with the thread `start()` methods and the manager lookup mocked, covering archive-only with `rtfreq`, both modes with `rtfreq`, both modes without `rtfreq`, and rapidfire-only with `rtfreq`. Four of the five tests fail against master and pass with this change.


tkeffer:
Thanks Robert!

I will note that option `rtfreq` is not actually used by Wunderground. See [this comment](https://github.com/weewx/weewx/issues/128#issuecomment-225394020). It is a relic from Adobe Flash days (remember that?!)

rnavarro:
Wow that's a blast from the past! Good to know!

Thanks for being a great steward of this software!

